PR #2, opened 2019-03-12
Feature/skeleton

agend:


proto/bob.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
```
ikopylov: This file was added in the previous PR. Why is it here again?

src/bin/bobd.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    let _server = server.build().expect("server");
+
+    println!(
+        "bobd started on port {}",
+        port
+    );
+
+    loop {
+        thread::park();
```
ikopylov: Is this a test server implementation? If it is not, then proper termination should be implemented.

PR #3, opened 2019-03-13
burn this stuff

agend:


src/bin/bobp.rs
```diff
@@ -1,3 +1,136 @@
...
+    }
+}
+
+fn help() {
+    println!("usage: bobp <target ip> <target port> <payload size> <worker count> <record count>");
+}
+
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
```
pyakushin: I think, it is more convenient to use clap.rs for args parsing

src/bin/bobp.rs
```diff
@@ -1,3 +1,136 @@
...
+            high_idx: task_conf.low_idx + task_size * (i+1),
+            payload_size: task_conf.payload_size
+        };
+        workers.push(thread::spawn(move || {    
+            bench_worker(nc, tc, stat_inner);
+        }));
+    }
+
+    for worker in workers {
+        let _ = worker.join();
```
pyakushin: I think, it's better to use `.join().unwrap()`, or match result

src/bin/bobp.rs
```diff
@@ -1,3 +1,136 @@
...
+            low_idx: task_conf.low_idx + task_size * i,
+            high_idx: task_conf.low_idx + task_size * (i+1),
+            payload_size: task_conf.payload_size
+        };
+        workers.push(thread::spawn(move || {    
+            bench_worker(nc, tc, stat_inner);
+        }));
+    }
+
+    for worker in workers {
```
pyakushin: Another way: workers.drain(..).for_each(|w| w.join().unwrap());

ikopylov:
LGTM

PR #6, opened 2019-04-01
logging and better result propagation

agend:


src/bin/bobd.rs
```diff
@@ -32,26 +38,52 @@ impl server::BobApi for BobSrv {
...
-                    let mut opts: BobOptions = Default::default();
-                    if let Some(vopts) = param.options.as_ref() { 
-                        if vopts.force_node {
-                            opts |= BobOptions::FORCE_NODE;
+            let sw = Stopwatch::start_new();
+            let param = req.into_inner();
+
+            if !self.put_validate(&param) {
+                warn!("PUT[-] invalid arguments - key and data is mandatory");
+                Box::new(future::err(tower_grpc::Status::new(tower_grpc::Code::InvalidArgument, "Key and data is mandatory")))
```
ikopylov: I would prefer to see the `return` expression here. It is easy to make a mistake with this functional style by  simply adding a code after the `if` statement.

src/bin/bobd.rs
```diff
@@ -32,26 +38,52 @@ impl server::BobApi for BobSrv {
...
-                    opts
-                }
-            ).then(|_r| ok(Response::new(OpStatus {
-                error: None
-            }))))
+                ).then(move |r| {
+                    let elapsed = sw.elapsed_ms();
+                    match r {
+                        Ok(r_ok) => {
+                            info!("PUT[{}]-OK local:{} ok dt: {}ms", key, !r_ok.is_clustered, elapsed);
```
ikopylov: This message will be written for every successful request. It is better to write it at the 'debug' level.

src/bin/bobd.rs
```diff
@@ -32,26 +38,52 @@ impl server::BobApi for BobSrv {
...
+                Box::new(future::err(tower_grpc::Status::new(tower_grpc::Code::InvalidArgument, "Key and data is mandatory")))
+            } else {
+                let key = BobKey{
+                            key: param.key.unwrap().key
+                        };
+                let data = BobData{
+                            data: param.data.unwrap().data
+                        };
+
+                debug!("PUT[{}] data size: {}", key, data.data.len());
```
ikopylov: It is better to use 'trace' level here

src/core/bob_client.rs
```diff
@@ -67,11 +67,14 @@ impl BobClient {
...
-            println!("ERR = {:?}", e);
+            debug!("BobClient: ERR = {:?}", e);
         })
     }
 
-    pub fn put(&mut self, key: BobKey, data: &BobData) -> impl Future<Item=BobPutResult, Error=BobError> {
+    pub fn put(&mut self, key: BobKey, data: &BobData)
+            -> impl Future<Item=ClusterResult<BobPutResult>, Error=ClusterResult<BobError>> {
+        let n1 = self.node.clone();
+        let n2 = self.node.clone();
```
ikopylov: Why do you clone it twice in the beginning of method?

src/bin/bobd.rs
agend: This is rust style) and clippy (rust linter) will generate warning of needless return

src/bin/bobd.rs
agend: this is intentional now

src/bin/bobd.rs
agend: make sense

src/core/bob_client.rs
agend: as below we need this variable in two closures. In rust they prefer explicit cloning

PR #8, opened 2019-04-10
Feature/config

nmenshov:


tests/config_tests.rs
```diff
@@ -0,0 +1,418 @@
+#[cfg(test)]
```
agend: @pyakushin where proper place for test in this case?

src/core/config.rs
```diff
@@ -0,0 +1,257 @@
+extern crate itertools;
```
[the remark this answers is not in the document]
nmenshov: place configs to src/<folder> or src/core/<folder>?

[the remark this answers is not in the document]
agend: src/core/config/ or src/config/ 
@pyakushin what do you think?

tests/config_tests.rs
pyakushin: In the `config` mod, tests in project root are integration for crate.
[Rust Book Tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest)

src/core/configs/cluster.rs
```diff
@@ -0,0 +1,397 @@
...
+            .as_ref()?
+            .iter()
+            .map(|elem| elem.validate())
+            .filter(|f| f.is_some())
+            .collect::<Vec<Option<String>>>();
+        if options.len() > 0 {
+            return Some(
+                options
+                    .iter()
+                    .fold("".to_string(), |acc, x| acc + &x.as_ref().unwrap()),
```
pyakushin: You will get all errors in one line without delimiters

src/core/configs/cluster.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+}
+
+use crate::core::data::Node as DataNode;
+use crate::core::data::NodeDisk as DataNodeDisk;
+use crate::core::data::VDisk as DataVDisk;
+
+use std::collections::HashMap;
+
+pub trait BobConfig {
```
pyakushin: Why trait and empty struct?
Isn't it better to implement same methods for Cluster?
With usage:
```rust 
let cluster_config = Cluster::from_file(filename).unwrap();
let vdisk_data: Vec<_> = cluster_config.to_data().unwrap();
```

src/core/configs/cluster.rs
```diff
@@ -0,0 +1,397 @@
...
+                    },
+                };
+                disk.replicas.push(node_disk);
+            }
+            result.push(disk);
+        }
+        Some(result)
+    }
+    fn get_cluster_config(&self, filename: &String) -> Result<Vec<DataVDisk>, String> {
+        let file = self.read_config(filename);
```
pyakushin: Shorter, without `match`
```rust
let file = self.read_config(filename)?;
```

src/core/configs/cluster_tests.rs
```diff
@@ -0,0 +1,454 @@
...
+        - name:
+          path: /tmp/d1
+vdisks:
+    - id: 0
+      replicas:
+        - node: n1
+          disk: disk1
+";
+        let d: Cluster = YamlConfig {}.parse_config(&s.to_string()).unwrap();
+        assert_eq!(false, d.validate().is_none());
```
pyakushin: Or 
```rust
assert!(d.validate().is_some());
```

src/core/configs/cluster.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    fn read_config(&self, filename: &String) -> Result<Cluster, String>;
+    fn parse_config(&self, config: &String) -> Result<Cluster, String>;
+    fn convert_to_data(&self, cluster: &Cluster) -> Option<Vec<DataVDisk>>;
+}
+
+pub struct YamlConfig {}
+
+impl BobConfig for YamlConfig {
+    fn read_config(&self, filename: &String) -> Result<Cluster, String> {
```
pyakushin: More convenient to use `&str`, because &String means, that you have to pass reference to object String. But if you want to use `static str`:
```rust
let s = "test static str";
```
 as filename, you will have to create new owned `String` first: 
```rust
config.read_config(&s.to_string());
```
. But in case of `&str` it would be only: 
```rust
config.read_config(s);
```

src/core/configs/cluster_tests.rs
```diff
@@ -0,0 +1,454 @@
...
+        - name: disk2
+          path: /tmp/d2
+vdisks:
+    - id: 0
+      replicas:
+        - node: n1
+          disk: disk1
+";
+        let d: Cluster = YamlConfig {}.parse_config(&s.to_string()).unwrap();
+        assert_eq!(true, d.validate().is_none());
```
pyakushin: Validation will pass with `address: 9999,0:-100000`, and next assertion failes with:
```
'assertion failed: `(left == right)`  left: `111`, right: `31072`
```

src/core/configs/cluster.rs
```diff
@@ -0,0 +1,397 @@
...
+    pub host: RefCell<String>,
+    #[serde(skip)]
+    pub port: Cell<i32>,
+}
+
+impl Node {
+    fn prepare(&self) -> Option<String> {
+        let ip: Vec<&str> = self.address.as_ref()?.split(":").collect::<Vec<&str>>();
+        self.host.replace(ip[0].to_string());
+        let port = ip[1].parse::<i32>();
```
pyakushin: If `address == "0.0.0.0" thread will panic with
`'index out of bounds: the len is 1 but the index is 1'`

PR #9, opened 2019-04-11
added read via any success

agend:


src/bin/bobc.rs
```diff
@@ -50,7 +50,7 @@ fn main() {
     wait_for_input();
     let get_req = client
         .get(Request::new(GetRequest {
-            key: None,
+            key: Some(BlobKey { key: 0 }),
```
pyakushin: You can use enum or tuple struct
```rust
struct BlobKey(usize);
// or
enum BlobKey(usize);
```

src/bin/bobd.rs
```diff
@@ -30,21 +30,25 @@ impl BobSrv {
         self.grinder.get_periodic_tasks(ex)
     }
 
-    fn put_validate(&self, req: &PutRequest) -> bool {
+    fn put_is_valid(req: &PutRequest) -> bool {
         !(req.key == None || req.data == None)
```
pyakushin: Another way:
```rust
req.key.is_some() && req.data.is_some()
```

src/bin/bobd.rs
```diff
@@ -30,21 +30,25 @@ impl BobSrv {
...
     }
+
+    fn get_is_valid(req: &GetRequest) -> bool {
+        req.key != None
```
pyakushin:
```rust
req.key.is_some()
```

src/bin/bobd.rs
```diff
@@ -70,45 +74,106 @@ impl server::BobApi for BobSrv {
...
+                    }),
+            )
+        }
+    }
+
+    fn get(&mut self, req: Request<GetRequest>) -> Self::GetFuture {
+        let sw = Stopwatch::start_new();
+        let param = req.into_inner();
+        if !Self::get_is_valid(&param) {
+            warn!("GET[-] invalid arguments - key and data is mandatory");
```
ikopylov: What do you mean by "data" in this case? Is this a copy-paste mistake?

src/bin/bobd.rs
```diff
@@ -70,45 +74,106 @@ impl server::BobApi for BobSrv {
...
-                                    r_err.to_string()
+                                    r_err
                                 );
-                                ok(Response::new(OpStatus {
-                                    error: Some(BobError {
-                                        code: 1,
-                                        desc: format!("Failed to write {}", r_err.to_string()),
-                                    }),
-                                }))
+                                let err = match r_err {
```
pyakushin: Looks like you just convert errors, you can define your own `enum Error` and use combinators instead of `match`

src/bin/bobp.rs
```diff
@@ -84,13 +85,19 @@ impl Service<()> for Dst {
...
 
     while !stop_token.load(Ordering::Relaxed) {
         thread::sleep(pause);
         let cur_put_count = stat.put_total.load(Ordering::Relaxed);
         let put_count_spd = (cur_put_count - last_put_count) / (period_ms / 1000);
         last_put_count = cur_put_count;
-        println!("put: {:5} rps", put_count_spd);
+
+        let cur_get_count = stat.get_total.load(Ordering::Relaxed);
+        let get_count_spd = (cur_get_count - last_get_count) / (period_ms / 1000);
```
ikopylov: When the `period_ms < 1000` there will be a division by zero.

src/core/sprinkler.rs
```diff
@@ -187,4 +161,80 @@ impl Sprinkler {
...
+        );
+        let reqs = Self::call_nodes(&mut self.get_connections(&target_nodes), |conn| {
+            Box::new(conn.get(key))
+        });
+
+        Box::new(
+            select_ok(reqs) // any result will enought
+                .map(|(r, _)| r)
+                .map_err(|r| {
+                    println!("FUCK {:?}", r);
```
ikopylov: Too expressive

src/bin/bobc.rs
agend: this particular code comes from generated part by tower-grpc. but other place can be fixed - in future PR

PR #15, opened 2019-06-06
Feature/pearl integration

nmenshov:


src/core/backend/pearl_backend.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    pub fn new_read(key: BobKey) -> Self {
+        PearlKey {
+            key: key.key.clone().to_be_bytes().to_vec(),
+            timestamp: 0, // TODO
+        }
+    }
+}
+impl Key for PearlKey {
+    const LEN: u16 = 16;
```
agend: we use u64 as key. so should be 8 bytes

src/core/backend/pearl_backend.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        PearlBackend {
+            config: pearl_config,
+            pool,
+            vdisks: Arc::new(vec![]),
+            alien_dir: Arc::new(None),
+        }
+    }
+
+    pub fn init(&mut self, mapper: &VDiskMapper) -> Result<(), String> {
```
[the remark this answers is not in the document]
agend: I see it will be done in background

src/core/backend/pearl_backend.rs
```diff
@@ -0,0 +1,313 @@
...
+        builder = match config.max_blob_size {
+            Some(max_blob_size) => builder.max_blob_size(max_blob_size),
+            _ => panic!("'max_blob_size' is not set in pearl config"),
+        };
+
+        builder.build().unwrap()
+    }
+
+    fn run_storage(&mut self, storage: &mut PearlStorage) {
+        self.pool.run(storage.init(self.pool.clone())).unwrap();
```
agend: this could fail and we don't need to fail all service

src/core/backend/pearl_backend.rs
```diff
@@ -0,0 +1,313 @@
...
+        builder = match config.max_blob_size {
+            Some(max_blob_size) => builder.max_blob_size(max_blob_size),
+            _ => panic!("'max_blob_size' is not set in pearl config"),
+        };
+
+        builder.build().unwrap() //TODO handle fail and try restart
+    }
+
+    fn run_storage(&mut self, storage: &mut PearlStorage) {
+        self.pool.run(storage.init(self.pool.clone())).unwrap(); //TODO handle fail and try restart
```
agend: Before we can handle put/get operations - we need to wait for init to complete.

PR #19, opened 2019-06-18
Feature/sprinkler refactoring

nmenshov:


src/core/cluster.rs
```diff
@@ -91,17 +91,12 @@ impl Cluster for QuorumCluster {
                 );
                 // TODO: send actuall list of vdisk it has been written on
                 if ok_count >= l_quorum as usize {
-                    ok(SprinklerResult {
-                        total_ops: total_ops as u16,
-                        ok_ops: ok_count as u16,
-                        quorum: l_quorum,
-                    })
+                    ok(BackendPutResult {})
```
pyakushin: I think `map` is simpler

src/core/configs/node.rs
```diff
@@ -185,6 +189,13 @@ impl NodeConfig {
 }
 impl Validatable for NodeConfig {
     fn validate(&self) -> Result<(), String> {
+        match self.ping_threads_count {
```
pyakushin:
```rust
if let None = self.ping_threads_count { . . }
```

src/core/grinder.rs
```diff
@@ -1,70 +1,86 @@
...
+    NotFound,
+    Other,
 }
 
-impl<CT, BT> ServeTypeOk<CT, BT> {
-    pub fn is_cluster(&self) -> bool {
-        match *self {
-            ServeTypeOk::Cluster(_) => true,
-            ServeTypeOk::Local(_) => false,
+impl BobError {
```
pyakushin: What is the purpose of this transformation?

src/core/grinder.rs
```diff
@@ -1,70 +1,86 @@
...
+            _ => BobError::Other,
+        }
+    }
+    fn is_cluster(&self) -> bool {
         match *self {
-            ServeTypeError::Cluster(_) => true,
-            ServeTypeError::Local(_) => false,
+            BobError::Cluster(_) => true,
+            BobError::Local(_) => false,
+            _ => panic!("dont use anything except Local and Cluster types"),
```
pyakushin: I think, panic is too much

src/core/link_manager.rs
```diff
@@ -52,17 +53,18 @@ impl NodeLinkHolder {
...
 
-    async fn check(&self, client_fatory: BobClientFactory) -> Result<(), ()> {
+    async fn check(self, client_fatory: BobClientFactory) -> Result<(), ()> {
         match self.get_connection().conn {
             Some(mut conn) => {
                 let nlh = self.clone();
-                match conn.ping().await {
-                    Ok(_) => debug!("All good with pinging node {:?}", nlh.node),
-                    Err(_) => {
+                let _ = conn
```
pyakushin: It is better to call `?` on result, so you wouldn't miss the error

src/core/server.rs
```diff
@@ -14,20 +13,26 @@ use stopwatch::Stopwatch;
...
-    ) -> Box<impl Future<Item = (), Error = ()> + Send> {
-        let grinder = self.grinder.clone();
-        let q = async move { grinder.get_periodic_tasks(ex).await };
-        Box::new(q.boxed().compat())
+        spawner: S,
+    ) -> Result<(), ()>
+    where
+        S: Spawn + Clone + Send + 'static + Unpin + Sync,
+    {
+        // let grinder = self.grinder.clone();
```
pyakushin: This code may still need?

PR #21, opened 2019-07-16
Feature/pools and timeouts

nmenshov:


src/core/bob_client.rs
```diff
@@ -194,31 +195,20 @@ impl BobClient {
...
-                                ))
-                            }
-                        },
-                    }),
+                    .map_err(|e| Error::StorageError(format!("ping operation error: {}", e))),
                 Err(_) => panic!("Timeout failed in core - can't continue"), //TODO
             })
             .compat()
             .boxed()
+            .timeout(to)
```
agend: Does this one chaining works for checking timeout after making `map` and `map_err`?

PR #24, opened 2019-09-16
Feature/cluster logic

nmenshov:


proto/bob.proto
```diff
@@ -46,8 +46,9 @@ message OpStatus {
 
 // Put operation options
 message PutOptions {
-   bool force_node = 1; // Force operation to be served by node to which it comes
-   bool overwrite = 2; // Overwrite data in case of id existens
+    repeated string remote_nodes = 1;
```
agend: what is format for this field?

src/core/backend/core.rs
```diff
@@ -16,7 +17,8 @@ use std::{pin::Pin, sync::Arc};
 pub struct BackendOperation {
     vdisk_id: VDiskId,
     disk_path: Option<DiskPath>,
-    alien: bool, // flag marks data belonging for different node
+    remote_node_name: Option<String>, // save data to alien/<remote_node_name>
```
agend: if we have `remote_node_name` do we need `alien` flag?

src/core/mapper.rs
```diff
@@ -0,0 +1,114 @@
...
+            local_node_name: config.name.as_ref().unwrap().to_string(),
+            disks: config
+                .disks()
+                .iter()
+                .map(|d| DiskPath::new(&d.name.clone(), &d.path.clone()))
+                .collect(),
+            nodes,
+        }
+    }
+    pub fn new2(
```
agend: what is `new2`??

src/core/mapper.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl VDiskMapper {
+    pub fn new(vdisks: Vec<VDisk>, config: &NodeConfig, cluster: &ClusterConfig) -> VDiskMapper {
+        let (nodes, vdisks) = Self::prepare_nodes(vdisks, &cluster);
+
+        VDiskMapper {
+            vdisks,
+            local_node_name: config.name.as_ref().unwrap().to_string(),
+            disks: config
```
agend: Could we call `new2` here with `new2(..., config.disks...)`?

src/core/backend/core.rs
nmenshov: in next branch i want to refact this logic

proto/bob.proto
nmenshov: it's array of node names

src/core/cluster/quorum.rs
```diff
@@ -0,0 +1,1007 @@
...
+                        err
+                    )))
+                }
+            }
+        };
+        Put(p.boxed())
+    }
+
+    //todo check no data (no error)
+    fn get_clustered(&self, key: BobKey) -> Get {
```
agend: Add async suffix for function to mark as it's async

src/core/mapper.rs
nmenshov: rename) it used when you run cluster from script and must run node with target name (not from node.yaml config)

PR #28, opened 2019-09-23
Feature/pearl backend folders

nmenshov:


src/core/backend/core.rs
```diff
@@ -67,6 +70,12 @@ pub struct BackendGetResult {
     pub data: BobData,
 }
 
+impl std::fmt::Display for BackendGetResult {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.data)
```
agend: Print out data? Is it a good idea?

src/core/backend/pearl/group.rs
```diff
@@ -0,0 +1,277 @@
...
+            vdisk_id,
+            node_name,
+            directory_path,
+            config,
+            spawner,
+            disk_name,
+        }
+    }
+
+    pub fn equal(&self, operation: &BackendOperation) -> bool {
```
agend: It's more that simple `equals`. Consider rename this method

src/core/backend/pearl/group.rs
```diff
@@ -0,0 +1,277 @@
...
+            }
+            exit = true;
+        }
+
+        exit = false;
+        while !exit {
+            if let Err(err) = self.run_pearls().await {
+                error!("can't start pearls: {:?}", err);
+                let _ = sleep(delay).compat().boxed().await;
+                continue;
```
agend: avoid using continue

src/core/mapper.rs
```diff
@@ -107,4 +111,10 @@ impl VDiskMapper {
         }
         (vdisk_id, Some(DiskPath::from(disk.unwrap())))
     }
+
+    pub fn is_node_holds_vdisk(&self, node_name: &str, id: VDiskId) -> bool {
```
agend: `is_node_holds_vdisk` -> `does_node_hold_vdisk`

src/core/backend/core.rs
nmenshov: I rewrite datas Display trait

PR #47, opened 2019-12-24
Informative not found error

pyakushin:
Contains changes from #46 PR

### DIFF
--- a/src/core_inner/backend/pearl/group.rs
+++ b/src/core_inner/backend/pearl/group.rs
@@ -242,22 +242,18 @@ impl PearlGroup {
                     trace!("get data: {} from: {}", data, holder);
                     results.push(data);
                 }
-                Err(err) if err != BackendError::KeyNotFound => {
+                Err(err) if err != BackendError::KeyNotFound(key) => {
                     has_error = true;
                     debug!("get error: {}, from : {}", err, holder);
                 }
                 _ => debug!("key not found from: {}", holder),
             }
         }
-        if results.is_empty() {
-            if has_error {
-                debug!("cannot read from some pearls");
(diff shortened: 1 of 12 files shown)


src/core_inner/backend/pearl/group.rs
```diff
@@ -242,22 +242,18 @@ impl PearlGroup {
...
             }
         }
-        if results.is_empty() {
-            if has_error {
-                debug!("cannot read from some pearls");
-                Err(Error::Failed("cannot read from some pearls".to_string()))
-            } else {
-                Err(Error::KeyNotFound)
-            }
+        if results.is_empty() && has_error {
```
agend: Why error and empty results branches put together?

src/core_inner/cluster/simple.rs
```diff
@@ -97,26 +97,29 @@ impl Cluster for Quorum {
...
-                }
-            })
-            .boxed();
+        let task = async move {
+            let results = reqs.collect::<Vec<_>>().await;
+            let err_results = results
+                .iter()
+                .filter_map(|r| r.as_ref().err())
+                .collect::<Vec<_>>();
+            let errors = err_results
```
agend: errors only logs, should we do something useful with them?

src/core_inner/backend/pearl/group.rs
pyakushin: I preserve the same logic. Settings::choose data returns result value with only KeyNotFound error. I changed return type to option and did equal logic transormation

src/core_inner/cluster/simple.rs
pyakushin: I've thought, that it is the task for the next refactoring PR. There are a lot of that kind of code

PR #50, opened 2020-01-15
Cluster config generator

pyakushin:


src/bin/ccg.rs
```diff
@@ -0,0 +1,393 @@
...
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Write};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+const ORD: Ordering = Ordering::Relaxed;
+
+#[tokio::main]
+async fn main() {
+    init_logger();
+    if let Some(output) = read_from_file().and_then(generate_config) {
```
agend: it would be nice to have option to output to stdout or have only option to output to stdout. then it could be used as `./ccg --inut ttt > output.cfg`

src/bin/ccg.rs
```diff
@@ -0,0 +1,393 @@
...
+use log::{Level, LevelFilter};
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Write};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+const ORD: Ordering = Ordering::Relaxed;
+
+#[tokio::main]
+async fn main() {
+    init_logger();
```
agend: --help with example how to execute would be nice

src/bin/ccg.rs
pyakushin: Done.

src/bin/ccg.rs
pyakushin: clap provides help message with descrition of the each option

PR #52, opened 2020-01-25
Bobp improvement

idruzhitskiy:


agend:
LGTM. @pyakushin  take a look

src/bin/bobp.rs
```diff
@@ -1,150 +1,494 @@
...
+            put_error,
+            get_error,
+            put_count_spd as f64 * k,
+            get_count_spd as f64 * k,
+            finite_or_default(cur_st_put_time / cur_st_put_count / 1e9),
+            finite_or_default(cur_st_get_time / cur_st_get_count / 1e9)
         );
     }
+    let elapsed = start.elapsed();
+    let elapsed_secs = elapsed.as_secs() as f64;
```
pyakushin: better to use inner float representation of the duration [as_secs_f64()](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64)

src/bin/bobp.rs
```diff
@@ -1,150 +1,494 @@
...
         stat.put_total.fetch_add(1, Ordering::SeqCst);
     }
 }
 
-#[tokio::main]
-async fn main() {
-    let matches = App::new("Bob benchmark tool")
+async fn test_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statistics>) {
+    put_worker(net_conf.clone(), task_conf.clone(), stat.clone()).await;
+    get_worker(net_conf.clone(), task_conf.clone(), stat.clone()).await;
```
pyakushin: redundant clone()

src/bin/bobp.rs
```diff
@@ -1,150 +1,494 @@
...
+}
+
+async fn ping_pong_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statistics>) {
+    let mut client = net_conf.build_client().await;
+
+    let get_options = task_conf.find_get_options();
+    let put_options = task_conf.find_put_options();
+    let measure_time = task_conf.is_time_measurement_thread();
+    let upper_idx = task_conf.low_idx + task_conf.count;
```
pyakushin: what if low_idx set to 0 in config?

idruzhitskiy: We'll iterate from 0 to count

pyakushin: I made a mistake, opposite - what if low_idx > 0 in config. There will be no time measurement threads?

idruzhitskiy: Yep, currently there will be no time measuring... I'll fix this

PR #53, opened 2020-01-29
measure speed as average of values

idruzhitskiy:


src/bin/bobp.rs
```diff
@@ -276,22 +278,24 @@ fn stat_worker(
 
         let put_error = stat.put_error.load(Ordering::Relaxed);
         let get_error = stat.get_error.load(Ordering::Relaxed);
-
+        let put_spd = put_count_spd as f64 * k;
+        let get_spd = get_count_spd as f64 * k;
+        put_speed_values.push(put_spd);
```
agend: how large could this `vec` get?

idruzhitskiy: len = execution time in seconds

src/bin/bobp.rs
```diff
@@ -318,6 +320,10 @@ fn stat_worker(
     );
 }
 
+fn average(values: &[f64]) -> f64 {
+    finite_or_default(values.iter().sum::<f64>() / values.len() as f64)
```
agend: better to have average in a window. @pyakushin

pyakushin: It can be both. Average on all values for results and moving average for current statistics

pyakushin: It is final stats, so I think it is fine

PR #54, opened 2020-02-03
add full conversion between error and status

idruzhitskiy:


src/core_inner/backend/error.rs
```diff
@@ -83,15 +83,50 @@ impl Into<Status> for Error {
...
+            Self::Internal => Status::internal("Internal"),
+            Self::PearlChangeState(msg) => Status::internal(format!("PearlChangeState {}", msg)),
         }
     }
 }
 
 impl From<Status> for Error {
     fn from(status: Status) -> Self {
-        Self::Failed(format!("{:?}", status))
+        let mut words = status.message().split_whitespace();
```
agend: Why do we need this?

idruzhitskiy: Format of messages: Type <optional args>. Can be easily separated by whitespaces

PR #60, opened 2020-03-04
Delete partition request

idruzhitskiy:


openapi.yaml
```diff
@@ -109,6 +109,25 @@ paths:
                 type: string
         default:
           $ref: '#/components/responses/DefaultError'
+  /vdisks/{vdiskId}/partitions/{partitionId}:
```
pyakushin: {VDiskId}, you can check at editor.swagger.io

openapi.yaml
```diff
@@ -109,6 +109,25 @@ paths:
...
+    delete:
```
pyakushin: API already has that path, add just `delete` method to it

src/api/http.rs
```diff
@@ -234,7 +235,7 @@ fn change_partition_state(
     let task = async move {
         match action {
             Action::Attach => group.attach(partition_id).await,
-            Action::Detach => group.detach(partition_id).await,
+            Action::Detach => group.detach(partition_id).await.map(|_| ()),
```
pyakushin: Why do you need to map to zero type?

pyakushin: I get it, maybe just:
```
{
    group.detach(partition_id).await;
}
```
?

src/core_inner/backend/pearl/group.rs
```diff
@@ -308,8 +308,9 @@ impl PearlGroup {
                         warn!("pearl closed: {:?}", e);
                     }
                 }
+                let pearl = pearl.clone();
```
pyakushin: Is it possible to find pearl and remove it, to prevent clone?

idruzhitskiy: I'll check

src/api/http.rs
idruzhitskiy: Oh, sure, missed that opportunity :)

idruzhitskiy: Oh, wait, actually we need error from that result, and map applies to Ok only

src/core_inner/backend/pearl/group.rs
idruzhitskiy: Done

pyakushin:
My tests
- [ ] deny to delete active partition
- [ ] inactive partition delete
- [ ] inactive partition delete under the load

PR #69, opened 2020-03-19
Add errors detalization to bobp

idruzhitskiy:


src/bin/bobp.rs
```diff
@@ -128,6 +134,9 @@ struct Statistics {
     get_count_single_thread: AtomicU64,
 
     unverified_puts: AtomicU64,
+
+    get_errors: Mutex<HashMap<CodeRepresentation, u64>>,
```
pyakushin: why do you need Mutex without Arc?

PR #85, opened 2020-05-30
Docker cluster runner

idruzhitskiy:


agend:
@idruzhitskiy please fix conflicts

src/bin/dcr.rs
```diff
@@ -0,0 +1,56 @@
...
+        config_dir.to_string(),
+        "/tmp".to_string(),
+    );
+    let configuration = get_configuration()?;
+    configuration.save_cluster_configuration(config_dir)?;
+    let compose = configuration.create_docker_compose(fs_configuration, "bobnet".to_string())?;
+    let arc = Arc::new(Mutex::new(Box::new(compose)));
+    let ctrlc_arc = arc.clone();
+    ctrlc::set_handler(move || {
+        let compose = ctrlc_arc.lock().unwrap();
```
pyakushin, line 30: Is it possible thread lock here?

src/bin/dcr.rs
```diff
@@ -0,0 +1,56 @@
...
+        compose
+            .down(config_dir)
+            .expect("failed to run \"down\"")
+            .wait_with_output()
+            .expect("failed to wait for compose down otuput");
+    })?;
+    arc.lock().unwrap().up(config_dir)?.wait_with_output()?;
```
pyakushin, line 37: Because we won't release mutex until the end of main() in the same thread (presumably)

src/bin/dcr.rs
```diff
@@ -0,0 +1,56 @@
...
+    Ok(())
+}
+
+fn get_configuration() -> Result<TestClusterConfiguration, Box<dyn Error>> {
+    let matches = App::new(env!("CARGO_PKG_NAME"))
+        .arg(
+            Arg::with_name("config")
```
pyakushin, line 44: If we have default config in git repo, it makes sense to add it as default arg value here

src/bin/dcr.rs
```diff
@@ -0,0 +1,56 @@
...
+                .help("config for docker cluster")
+                .takes_value(true)
+                .short("c")
+                .required(true)
+                .long("config"),
+        )
+        .get_matches();
+    let config_filename = matches.value_of("config").expect("required");
+    let file_content = fs::read_to_string(config_filename)?;
+    let configuration: TestClusterConfiguration = serde_yaml::from_str(&file_content)?;
```
pyakushin, line 54: You can deserialize directly from file, with `serde_yaml::from_reader`

build.rs
```diff
@@ -4,6 +4,7 @@ fn main() {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
+        .format(false)
```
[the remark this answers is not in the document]
idruzhitskiy: We have external format in build scripts, and sometimes rustfmt isn't available for latest nightly, so dockers won't build

cluster_test/Dockerfile
```diff
@@ -0,0 +1,56 @@
+# build image
+FROM rustlang/rust:nightly as cargo-build
+
+RUN apt-get update
+
+RUN apt-get install musl-tools -y
+
+RUN rustup target add x86_64-unknown-linux-musl
```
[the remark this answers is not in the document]
idruzhitskiy: Yep, for usage in alpine image

src/bin/dcr.rs
idruzhitskiy: Actually no, I'll introduce new var so it will be clear

src/bin/dcr.rs
idruzhitskiy: `    let file = File::open(path)?;
    let reader = BufReader::new(file);
    let obj = serde_yaml::from_reader(reader)?;`

Something like this? Reading as string looks more concise...

src/bin/dcr.rs
idruzhitskiy: Added default value

src/bin/dcr.rs
idruzhitskiy: Added new variable

Cargo.toml
```diff
@@ -36,6 +36,8 @@ rocket = "0.4.4"
 rocket_contrib = "0.4.4"
 async-trait = "0.1.31"
 termion = "1.5"
+derive-new = "0.5.8"
+ctrlc = "*"
```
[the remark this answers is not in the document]
idruzhitskiy: Fixed 3.1.4 version

agend:
@idruzhitskiy could you add some documentation on how to use it? that would be useful

PR #86, opened 2020-06-09
add hash to blob and partition names

idruzhitskiy:


src/core/backend/pearl/group.rs
```diff
@@ -286,15 +290,13 @@ impl Group {
...
-        Holder::new(
-            start_timestamp,
-            end_timestamp,
-            self.vdisk_id,
-            path,
-            self.settings.config().clone(),
-        )
+        let hash = self.get_node_hash();
+        let partition_name = PartitionName::new(start_timestamp, &hash);
+        path.push(format!("{}/", partition_name.to_string()));
```
pyakushin, line 295: If I remember correctly, you don't need `/` at path end. Anyway, it's better to push twice `partition_name` and `/`, instead of `format!()`.
(to avoid memory allocation)

src/core/backend/pearl/group.rs
```diff
@@ -319,16 +321,30 @@ impl Group {
...
                 .map_err(|_| warn!("cannot parse file name: {:?}", entry))
             {
-                let start_timestamp = file_name
-                    .parse()
-                    .map_err(|_| warn!("cannot parse file name: {:?} as timestamp", entry))
-                    .expect("parse file name");
-                let pearl_holder = self.create_pearl_holder(start_timestamp);
-                holders.push(pearl_holder);
+                let partition_name = PartitionName::try_from_string(&file_name);
+                if let Some(partition_name) = partition_name {
```
pyakushin, line 325: Add here log msg, if partition name parsing failed

src/core/backend/pearl/group.rs
```diff
@@ -342,3 +358,40 @@ impl Display for Group {
...
+impl PartitionName {
+    fn new(timestamp: u64, hash: &str) -> Self {
+        Self {
+            timestamp,
+            hash: hash.to_string(),
+        }
+    }
+
+    fn try_from_string(s: &str) -> Option<Self> {
+        let (timestamp_string, hash_string) = if let Some(index) = s.find('_') {
```
pyakushin, line 376: You can use [str.split](https://doc.rust-lang.org/std/primitive.str.html#method.split)

src/core/backend/pearl/group.rs
```diff
@@ -342,3 +358,40 @@ impl Display for Group {
...
+        } else {
+            (s, "")
+        };
+        if let Ok(timestamp) = timestamp_string.parse() {
+            Some(Self {
+                timestamp,
+                hash: hash_string.to_string(),
+            })
+        } else {
+            warn!("Failed to parse timestamp from {}", timestamp_string);
```
pyakushin, line 387: I think log message needed on higher level. Plus, you can write shorter: `timestamp_string.parse().ok()`.
I don't insist, on your consideration (:

src/core/backend/pearl/group.rs
```diff
@@ -319,16 +321,32 @@ impl Group {
...
     }
+
+    fn get_node_hash(&self) -> String {
+        let hash = digest(&SHA256, self.node_name.as_bytes());
+        let hex: Vec<u8> = hash
+            .as_ref()
+            .iter()
+            .flat_map(|&byte| {
+                vec![
+                    ((byte >> 4) & 0b01111u8 | 0b01000000) + 1,
```
agend, line 343: some comments what is happening here would be nice

idruzhitskiy: Added comments

agend:
LGTM

PR #88, opened 2020-06-24
Update search logic on get

pyakushin:


src/core/cluster/quorum.rs
```diff
@@ -250,37 +274,36 @@ impl Cluster for Quorum {
...
-        let second_attempt = Self::get_all(key, &sup_nodes, GetOptions::new_alien()).await;
-        debug!("GET[{}] cluster ans sup: {:?}", key, second_attempt);
+        let local_node = self.mapper.local_node_name();
+        let target_nodes = self
+            .mapper
+            .nodes()
+            .iter()
+            .filter(|node| node.name() != local_node);
+        // @TODO don't wait for all answers, return on first success
+        let results = Self::get_all(key, target_nodes, GetOptions::new_alien()).await;
```
agend, line 290: we already sent request for ALL before. do we need to query alien directly?

src/core/cluster/quorum.rs
```diff
@@ -231,14 +231,38 @@ impl Cluster for Quorum {
...
+            Err(e) => error!("local node backend returned error: {}", e),
+        };
+        trace!("GET[{}] ~~~LOOKUP REMOTE NODES~~~", key);
+        let local_node = self.mapper.local_node_name();
+        let target_nodes = self
+            .get_target_nodes(key)
+            .iter()
+            .filter(|node| node.name() != local_node);
+        // @TODO don't wait for all answers, return on first success
+        let results = Self::get_all(key, target_nodes, GetOptions::new_all()).await;
```
agend, line 265: Should it have force option as well as we set particular nodes list?

src/core/cluster/quorum.rs
```diff
@@ -113,7 +113,7 @@ impl Quorum {
```
agend, line 113: We should also rename quorum to smething else - like simple_cluster_operations

src/core/cluster/quorum.rs
pyakushin: [outdated] `force_node` flag is in `Operation` and is true in all backend operations after cluster

src/core/cluster/quorum.rs
pyakushin: fixed

Issue #89, opened 2020-07-02
On cluster put. Return result as soon as [quorum] successful answers receied


pyakushin:
Currently, after `put` we will wait for answers from all nodes. But we can return successful result as soon as [quorum] number of answers will be returned with `Ok`.

PR #90, opened 2020-07-09
introduce local part of the put operation

pyakushin:
Issue #89

### DIFF
--- a/src/core/cluster/operations.rs
+++ b/src/core/cluster/operations.rs
@@ -14,6 +14,16 @@ pub(crate) async fn get_any(
         .await
 }
 
+pub(crate) async fn put_at_least(
+    key: BobKey,
+    data: BobData,
+    target_nodes: impl Iterator<Item = &Node>,
+    at_least: usize,
+    options: PutOptions,
+) -> PutResult {
+    unimplemented!()
+}
+
 pub(crate) fn get_support_nodes<'a>(
     mapper: &'a Virtual,
     mut target_indexes: impl Iterator<Item = u16>,
@@ -184,5 +194,21 @@ pub(crate) async fn put_sup_nodes(
     } else {
         let msg = ret.iter().map(|x| format!("{:?}\n", x)).collect();
         Err((requests.len() - ret.len(), msg))
+    }
+}
+
+pub(crate) async fn put_local_node(
+    backend: &Backend,
+    key: BobKey,
+    data: BobData,
+    vdisk_id: VDiskId,
+    disk_path: Option<DiskPath>,
+) -> PutResult {
+    if let Some(path) = disk_path {
+        debug!("local node has vdisk replica, put local");
+        let op = Operation::new_local(vdisk_id, path);
+        backend.put_local(key, data, op).await
+    } else {
+        Ok(())
     }
 }
--- a/src/core/cluster/quorum.rs
+++ b/src/core/cluster/quorum.rs
@@ -14,12 +14,44 @@ impl Quorum {
             quorum,
         }
     }
+
+    pub(crate) async fn put_remote_nodes(&self, key: BobKey, data: BobData) -> PutResult {
+        let local_node = self.mapper.local_node_name();
+        let target_nodes = get_target_nodes(&self.mapper, key)
+            .iter()
+            .filter(|node| node.name() != local_node);
+        put_at_least(
+            key,
+            data,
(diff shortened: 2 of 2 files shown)


src/core/cluster/operations.rs
```diff
@@ -14,6 +15,60 @@ pub(crate) async fn get_any(
...
+) {
+    let mut handles: FuturesUnordered<_> = target_nodes
+        .map(|node| {
+            debug!("PUT[{}] put to {}", key, node.name());
+            let node = node.clone();
+            let data = data.clone();
+            let options = options.clone();
+            tokio::spawn(async move {
+                LinkManager::call_node(&node, |conn| {
+                    conn.put(key, data.clone(), options.clone()).boxed()
```
agend, line 36: second clone for data?

src/core/cluster/quorum.rs
```diff
@@ -14,99 +16,187 @@ impl Quorum {
...
-        let errors = results
-            .into_iter()
-            .filter_map(Result::err)
+    async fn put_at_least(&self, key: BobKey, data: BobData) -> Result<(), Error> {
+        debug!("PUT[{}] ~~~PUT LOCAL NODE FIRST~~~", key);
+        let mut local_put_ok = 0_usize;
+        let mut remote_ok_count = 0_usize;
+        let mut at_least = self.quorum;
+        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
+        if let Some(path) = disk_path {
```
agend, line 26: may be put local put logic to function?

src/core/cluster/quorum.rs
```diff
@@ -14,99 +16,187 @@ impl Quorum {
...
+            if let Err(err) = self.put_aliens(failed_nodes, key, data).await {
+                if errors.is_empty() {
+                    error!("PUT[{}] smth wrong with cluster/node configuration", key);
+                    error!(
+                        "PUT[{}] local_put_ok: {}, remote_ok_count: {}, quorum: {},no errors",
+                        key, local_put_ok, remote_ok_count, self.quorum
+                    );
+                    Err(err)
+                } else {
+                    let err = errors.remove(errors.len() - 1).into_inner();
```
agend, line 78: why delete last one?

src/core/cluster/quorum.rs
```diff
@@ -14,99 +16,187 @@ impl Quorum {
...
-                let nodes = errors
+    pub(crate) async fn put_aliens(
+        &self,
+        mut failed_nodes: Vec<String>,
+        key: BobKey,
+        data: BobData,
+    ) -> Result<(), Error> {
+        let vdisk_id = self.mapper.id_from_key(key);
+        let operation = Operation::new_alien(vdisk_id);
+        let local_put = put_local_all(
```
agend, line 142: do we put at least one copy locally?

src/core/cluster/quorum.rs
```diff
@@ -14,99 +16,187 @@ impl Quorum {
...
+            &self.backend,
+            failed_nodes.clone(),
+            key,
+            data.clone(),
+            operation,
+        )
+        .await;
+        if local_put.is_err() {
```
agend, line 150: some annotation would be nice )

src/core/cluster/quorum.rs
pyakushin: In that case we return last occured error.

src/core/cluster/quorum.rs
pyakushin: Added log

src/core/cluster/quorum.rs
pyakushin: It surprised me, but yes. Maybe this place needs a rework.

src/core/cluster/quorum.rs
pyakushin: It won't be nicer) Logic was already put to `put_local_all`

src/core/cluster/operations.rs
pyakushin: Fixed

src/core/cluster/quorum.rs
agend: Can we use `errors.last().into_inner()` ?

src/core/cluster/quorum.rs
agend: #95

PR #97, opened 2020-08-09
fix pearl create multiple threads

idruzhitskiy:


src/core/backend/pearl/group.rs
```diff
@@ -1,16 +1,17 @@
 use super::prelude::*;
 use ring::digest::{digest, SHA256};
+use std::collections::HashMap;
```
agend, line 3: Does it worth to move to prelude ?

src/core/backend/pearl/group.rs
```diff
@@ -1,16 +1,17 @@
...
 
 #[derive(Clone, Debug)]
 pub(crate) struct Group {
     holders: Arc<RwLock<Vec<Holder>>>,
-    pearl_sync: Arc<SyncState>,
```
agend, line 7: Show we delete `SyncState` trait then?

src/core/backend/pearl/group.rs
```diff
@@ -121,32 +123,40 @@ impl Group {
...
-        self.settings
-            .config()
-            .try_multiple_times_async(
-                || self.try_create_write_pearl(timestamp),
-                "pearl init failed",
-                self.settings.config().settings().create_pearl_wait_delay(),
-            )
+    async fn create_write_pearl(&self, ts: u64) -> BackendResult<Holder> {
+        if !self.created_holder_indexes.read().await.contains_key(&ts) {
+            let mut indexes = self.created_holder_indexes.write().await;
```
agend, line 128: Could it be here a Deadlock if you take read lock right before?

src/core/backend/pearl/group.rs
```diff
@@ -121,32 +123,40 @@ impl Group {
...
-    async fn create_write_pearl(&self, timestamp: u64) -> BackendResult<Holder> {
-        self.settings
-            .config()
-            .try_multiple_times_async(
-                || self.try_create_write_pearl(timestamp),
-                "pearl init failed",
-                self.settings.config().settings().create_pearl_wait_delay(),
-            )
+    async fn create_write_pearl(&self, ts: u64) -> BackendResult<Holder> {
+        if !self.created_holder_indexes.read().await.contains_key(&ts) {
```
agend, line 127: aren't here a race condition?

src/core/backend/pearl/group.rs
idruzhitskiy: Readlock gets dropped, isn't it?

agend: @pyakushin what do you think.
@idruzhitskiy what about race condtion?

PR #104, opened 2020-09-24
Allow partial completion of PUT

pyakushin:
Put data on all available support nodes. 
Return error if there weren't enough writes.
This ensures that the data is written, unless there are no available nodes at all.

### DIFF
--- a/src/core/cluster/operations.rs
+++ b/src/core/cluster/operations.rs
@@ -111,11 +111,6 @@ pub(crate) fn get_support_nodes(
     )
 }
 
-#[inline]
-pub(crate) fn get_target_nodes(mapper: &Virtual, key: BobKey) -> &[Node] {
-    mapper.get_vdisk_for_key(key).nodes()
-}
-
 pub(crate) fn group_keys_by_nodes(
     mapper: &Virtual,
     keys: &[BobKey],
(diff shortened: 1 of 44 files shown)


src/core/cluster/operations.rs
```diff
@@ -71,23 +70,39 @@ pub(crate) fn get_support_nodes<'a>(
...
-) -> Result<Vec<&'a Node>, Error> {
-    let (len, _) = target_indexes.size_hint();
+) -> (Vec<&'a Node>, Result<(), Error>) {
+    let len = target_indexes.size_hint().0;
     debug!("iterator size lower bound: {}", len);
     trace!("nodes available: {}", mapper.nodes().len());
-    if mapper.nodes().len() < len + count {
-        let msg = "cannot find enough support nodes".to_owned();
+    let res = if mapper.nodes().len() < len + count {
+        let msg = format!(
```
agend, line 78: If client trigger re-write on error - that could be lead to huge retries from clients.

pyakushin: But we can't return Ok, because we did'nt reach quorum. Maybe it's better for client to handle this kind of errors? In that case I will add different enum value, eg `Error::QuorumCannotBeReached`

agend: Error is error. It should return **Ok** if client request is fulfilled and further actions is required. Cluster administrator should see metrics, logs and act on them.

ikopylov: If the quorum is not reached, then this is definitely an error. The best behavior would be to not write the data at all in this case, but this is quite hard to implement correctly and efficiently. 

In the current state of things, if Bob returns an error, there will be a discrepancy, because the client can read the record from Bob successfully. If Bob will return `Ok`, then we will lose the quorum guarantee (the number of copies will be unpredictable).

My suggestion would be to write several `alien` copies on the local node in order to satisfy the quorum condition. With this solution, after restoring the failed cluster nodes, the number of copies will become correct (recovery tool will copy that aliens to the correct nodes). If we do this, then it will be ok to return `Ok` status.

PR #109, opened 2020-10-01
add method to api to receive local dir structure of vdisk replicas

idruzhitskiy:


src/api/http.rs
```diff
@@ -307,6 +315,64 @@ fn alien(_bob: State<BobServer>) -> &'static str {
...
+            false,
+            format!("Failed to get dirs for replica {:?}", path),
+        ))?;
+        result.push(dir);
+    }
+    Ok(Json(result))
+}
+
+fn create_directory(root_path: &Path) -> Option<Dir> {
+    if let (Some(name), Some(root_path_str)) = (
```
pyakushin, line 347: I think it's better to use `?` syntax to process Options.

src/api/http.rs
```diff
@@ -307,6 +315,64 @@ fn alien(_bob: State<BobServer>) -> &'static str {
...
+        root_path.file_name().and_then(|n| n.to_str()),
+        root_path.to_str(),
+    ) {
+        let children = std::fs::read_dir(root_path)
+            .ok()
```
pyakushin, line 352: Same, if you don't process errors, just unwrap it with `?`

src/api/http.rs
```diff
@@ -307,6 +315,64 @@ fn alien(_bob: State<BobServer>) -> &'static str {
...
+                            if child.path().is_dir() {
+                                create_directory(&child.path())
+                            } else {
+                                None
+                            }
+                        })
+                    })
+                    .collect::<_>()
+            })
+            .unwrap_or(vec![]);
```
pyakushin, line 366: `unwrap_or_default()`

PR #115, opened 2020-10-15
Add dockerfile

idruzhitskiy:


Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+  && chmod 600 -R ~/.ssh \
+  && echo "#!/bin/bash\n\
+    cp /local_ssh/* ~/.ssh\n\
+    chown -R root ~/.ssh\n\
+    eval $(ssh-agent)\n\
+    ssh-add ~/.ssh/id_rsa\n\
+    /usr/sbin/sshd -D &\n\
+    su -c \"./bobd -c /configs/\$1 -n /configs/\$2\" bobd" >> run.sh \
+  && chmod +x run.sh
+
```
ikopylov, line 63: I suggest to add `EXPOSE` instruction

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+COPY build.rs .
+RUN cargo build --release --target=x86_64-unknown-linux-gnu \
+  && rm -f target/x86_64-unknown-linux-gnu/release/deps/bob*
+
+# final build
+COPY . .
+RUN cargo build --release --target=x86_64-unknown-linux-gnu
+
+# bobd image
+FROM ubuntu:latest
```
ikopylov, line 36: Can you add a separate Dockerfile for Alpine as we discussed

compose_examples/two_nodes/configs/node_0.yaml
```diff
@@ -0,0 +1,25 @@
+---
+log_config: /configs/logger.yaml
+name: node_0
+quorum: 2
+operation_timeout: 3sec
+check_interval: 5000ms
+cluster_policy: quorum
+backend_type: pearl
+pearl:
+  max_blob_size: 1000000
```
ikopylov, line 10: I suggest to increse `max_blob_size` up to 10Gb

compose_examples/two_nodes/configs/cluster.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
ikopylov, line 1: It would be good to add an example with single node cluster (without replicas)

Makefile
```diff
@@ -0,0 +1,10 @@
+default: test
```
ikopylov, line 1: `build` target is commonly used as default one

compose_examples/one_node/configs/node.yaml
```diff
@@ -0,0 +1,25 @@
...
+log_config: /configs/logger.yaml
+name: node
+quorum: 1
+operation_timeout: 3sec
+check_interval: 5000ms
+cluster_policy: quorum
+backend_type: pearl
+pearl:
+  max_blob_size: 10000000000
+  max_data_in_blob: 10000
```
ikopylov, line 11: Set to `max_data_in_blob` 100000. Small blob size is bad for performance

dockerfiles/alpine/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    cp /local_ssh/* ~/.ssh\n\
+    chown -R root ~/.ssh\n\
+    eval $(ssh-agent)\n\
+    ssh-add ~/.ssh/id_rsa \n\
+    /usr/sbin/sshd -D & \n\
+    su -c \"./bobd -c /configs/\$1 -n /configs/\$2\" bobd" >> run.sh \
+  && chmod +x run.sh
+
+EXPOSE 80
+EXPOSE 22
```
ikopylov, line 71: Port 20000 must also be exposed as this is the entry port for Bob's clients

dockerfiles/alpine/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends musl-tools \
+  && rustup install $RUST_TC_VER \
+  && rustup default $RUST_TC_VER \
+  && rustup target add x86_64-unknown-linux-musl
+
+WORKDIR /usr/src/bob
+
+# crates downloading and initial build
+COPY Cargo.toml Cargo.toml
```
ikopylov, line 16: Now that the `Dockerfile` is in `dockerfile/alpine/` directory, should the paths for the `COPY` command be updated?

idruzhitskiy: Dockerfile is supposed to be run with context set to root dir (as in compose examples)

PR #120, opened 2020-10-19
Additional performance counters

### LINKED ISSUE #98: Additional performance counters
Add several useful performance counters:
1. Counters to track `exists` method performance: total request count, failed request count, request per second, elapsed time
2. Additional counters for `get` and `put`: failed request count
3. Counter to track total number of blobs
4. Counter to track number of blobs in `alien` directory
5. Counter to track `Bob` status (this will allow to identify stuck during startup): 0 - starting, 1 - fully started
6. Counters to track side nodes to which the current node connected (should help to detect network issues): one counter per node with the number of connections to that node
7. Counter to track the number of lost/inaccessible blobs (should help to identify disk fault)

Justarone:
closes #98


Justarone:
parts 3-7 of issue have been being disscussed now, so they are in the process.
**THIS PULL REQUEST IS NOT READY TO BE MERGED NOW.**

ikopylov:
Files without changes look very suspicious. Can you exclude them from PR?

src/core/backend/pearl/core.rs
```diff
@@ -73,6 +75,40 @@ impl Pearl {
...
     }
+
+    async fn count_blobs(groups: Arc<Vec<Group>>, alien_groups_guard: Arc<RwLock<Vec<Group>>>,
+        count_interval: Duration) {
+        let mut interval = interval(count_interval);
+        loop {
+            interval.tick().await;
+
+            let mut cnt = 0;
+            for group in groups.iter() {
```
ikopylov, line 86: `groups` is not protected by any synchronization primitive, so I think it is unsafe to iterate it in this manner

src/core/backend/pearl/core.rs
```diff
@@ -73,6 +75,40 @@ impl Pearl {
                 Error::failed(format!("cannot find actual alien folder. {:?}", operation))
             })
     }
+
+    async fn count_blobs(groups: Arc<Vec<Group>>, alien_groups_guard: Arc<RwLock<Vec<Group>>>,
```
ikopylov, line 79: Please, recheck that the references (`group` and `alien_groups_guard`) are never changed during runtime. I.e. some other code does not perform collection reallocation.

src/core/backend/pearl/core.rs
pyakushin: why? for iterates over immutable reference.
Btw, @Justarone you can write:
```
for group in &groups {}
```

src/core/metrics.rs
```diff
@@ -26,13 +33,34 @@ metrics! {
...
+        pub CLIENT_EXIST_ERROR_COUNTER: Counter = "exist_error_count";
+        /// Measures processing time of the EXIST request
+        pub CLIENT_EXIST_TIMER: Timer = "exist_timer";
+    }
+}
+
+metrics! {
+    LINK_MANAGER: Proxy = "link_manager" => {
+        /// Observes number of connected nodes
+        pub NODES_NUMBER: Gauge = "nodes_number";
```
pyakushin, line 56: I think it's better to name it `AVAILABLE_NODES_COUNT`

src/core/backend/pearl/core.rs
Justarone: Yes, I've rechecked that and discussed with @pyakushin . There is no problem because groups doesn't change during runtime of the program (once they are built; you can check it: no method in public interface of Pearl (BackendStorage) that changes groups vector) and alien groups are under RWLock, so they can't be changed while I iterate on them.

src/core/metrics.rs
```diff
@@ -90,6 +124,22 @@ impl BobClient {
     pub(crate) fn get_error_count(&self) {
         self.get_error_count.count(1);
     }
+
+    pub(crate) fn exist_count(&self) {
```
pyakushin, line 128: Maybe better to use name `exist_counter_inc()` or `exist_inc()`

pyakushin (approved):

src/core/metrics.rs
Justarone: I've tried to save the name policy of other methods. So should I change only this one or all of them?

src/core/backend/pearl/core.rs
Justarone: Yes, there is no problem, because inside I call `group.holders()`, which returns cloned `Arc<RwLock<Vec<Holder>>>>` and then work with it (it's protected by sync primitivies)

Justarone:
Pull request of fork isn't usual practice for this project, so I decided to move all this code to inner branch. You can find it in PR #131

PR #126, opened 2020-10-29
Finalize old blobs

idruzhitskiy:


src/api/http.rs
```diff
@@ -163,6 +164,17 @@ fn vdisks(bob: State<BobServer>) -> Json<Vec<VDisk>> {
     Json(vdisks)
 }
 
+#[delete("/blobs/outdated")]
+fn finalize_outdated_blobs(bob: State<BobServer>) -> Result<StatusExt, StatusExt> {
+    let backend = bob.grinder().backend();
+    runtime().block_on(backend.cleanup_outdated(1));
```
agend, line 170: that could take a while - so don't block

openapi.yaml
```diff
@@ -142,6 +142,15 @@ paths:
                 type: string
         default:
           $ref: '#/components/responses/DefaultError'
+  /blobs/outdated:
+    delete:
+      summary: Finalize old blobs
```
agend, line 147: I would extend: Close active blobs not recently used

src/core/backend/pearl/holder.rs
```diff
@@ -58,6 +64,20 @@ impl Holder {
         self.start_timestamp <= timestamp && timestamp < self.end_timestamp
     }
 
+    pub(crate) fn is_outdated(&self) -> bool {
```
agend, line 67: I think it's better to close not recently used blobs. It should be easy to track last time blob was written.

src/core/backend/core.rs
```diff
@@ -296,4 +296,24 @@ impl Backend {
             None
         }
     }
+
+    pub(crate) async fn cleanup_outdated(&self, max: usize) {
```
agend, line 300: cleanup or close?

src/core/backend/pearl/holder.rs
agend: As well we can set not to close too small blobs

src/core/backend/core.rs
idruzhitskiy: close, typo

src/api/http.rs
pyakushin: it's different runtime, it won't block main RT

PR #127, opened 2020-10-30
add nodes to api

idruzhitskiy:


openapi.yaml
```diff
@@ -17,6 +17,21 @@ paths:
                 $ref: '#/components/schemas/Node'
         default:
           $ref: '#/components/responses/DefaultError'
+  /nodes:
+    get:
+      summary: Returns a list of known nodes
+      operationId: getNodes
+      responses:
+        '200':
+          description: "A JSON array of nodes info"
```
agend, line 26: no vdisks?

src/api/http.rs
```diff
@@ -160,6 +162,33 @@ fn status(bob: State<BobServer>) -> Json<Node> {
...
 
+#[get("/nodes")]
+fn nodes(bob: State<BobServer>) -> Json<Vec<Node>> {
+    let mapper = bob.grinder().backend().mapper();
+    let mut nodes = vec![];
+    let vdisks = collect_disks_info(&bob);
+    for node in mapper.nodes().values() {
+        let vdisks: Vec<VDisk> = vdisks
+            .iter()
+            .cloned()
```
pyakushin, line 173: may be it's better to use filter_map and then clone filtered vdisks to reduce clones.

src/api/http.rs
```diff
@@ -160,6 +162,33 @@ fn status(bob: State<BobServer>) -> Json<Node> {
...
+            .map(|mut vd| {
+                vd.replicas.drain_filter(|r| r.node != node.name());
+                vd
+            })
+            .filter(|vd| !vd.replicas.is_empty())
```
pyakushin, line 178: And include this condition to the filter_map above

PR #128, opened 2020-11-03
New metrics drain and counters initialization

Justarone:
* new metrics drain: now there is only average time, counters sum and gauge max value (no more unnecessary metrics); 
* counters initialization (need it to have errors counters right after startup, not after first error).


ikopylov:
LGTM

PR #131, opened 2020-11-10
Add additional counters

### LINKED ISSUE #98: Additional performance counters

Justarone:
closes #98 

New pull request of old code. This code has been reviewed in another PR (#120), but that pull request was created from pretty old master and it was from forked repository, but in this project this isn't a usual practice, so the new pull request was created.


Justarone:
The only difference from reviewed version is that in backend/core.rs I call `self.inner.run()` to activate metrics.
I also want all to pay attention that there is a warning on implemented trait method.

src/core/backend/core.rs
```diff
@@ -82,6 +85,15 @@ pub(crate) trait BackendStorage: Debug {
...
     async fn exist_alien(&self, op: Operation, keys: &[BobKey]) -> Result<Vec<bool>, Error>;
 
+    async fn run(&self) -> Result<()> {
+        BACKEND_STATE.value(BACKEND_STARTING);
+        let result = self.run_backend().await;
+        BACKEND_STATE.value(BACKEND_STARTED);
+        result
+    }
+
+    fn spawn_counter(&self) {}
```
pyakushin, line 95: Why do you think this func belongs in BackendStorage trait?

Justarone: First of all I thought that it would be good to have background workers in one place. To spawn blobs (or not blobs) counter I need to do next steps:
1. Call counter from `grinder`  (from method `run_periodic_tasks`) using `backend` method
2. Then backend somehow need to spawn counter, BUT it has storage field of `dyn BackendStorage` type, so it can only use trait methods to spawn counter of storage.

So I found only 2 solutions:
1. Make trait method with empty default implementation, to call it from `Backend` (and from `Grinder`; so every storage may spawn it's counters)
2. Make this method on `Pearl` level, but in this case it seems like we should explicitly spawn counter from binary (that's not a problem, but makes us to change the code in binary either)

src/core/backend/pearl/core.rs
```diff
@@ -215,8 +203,8 @@ impl BackendStorage for Pearl {
         }
     }
 
-    fn spawn_counter(&self) {
-        self.spawn_blobs_counter();
+    async fn blobs_count(&self) -> (usize, usize) {
+        Self::count_blobs(self.vdisks_groups.clone(), self.alien_vdisks_groups.clone()).await
```
pyakushin, line 207: It seems like you can inline count_blobs here since it is the only place where it used

PR #132, opened 2020-11-10
Fix of "Config validation bug that misses the case where the quorum value exceeds the number of replicas"

### LINKED ISSUE #130: Config validation bug that misses the case where the quorum value exceeds the number of replicas

Justarone:
closes #130

### DIFF
--- a/src/core/configs/cluster.rs
+++ b/src/core/configs/cluster.rs
@@ -287,8 +287,8 @@ impl Cluster {
                 debug!("cannot find node: {} in cluster config", node.name());
                 format!("cannot find node: {} in cluster config", node.name())
             })?;
-        if self.vdisks().iter().any(|vdisk| vdisk.replicas().len() < node.quorum()) {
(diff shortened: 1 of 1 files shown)


src/core/configs/cluster.rs
```diff
@@ -287,6 +287,9 @@ impl Cluster {
                 debug!("cannot find node: {} in cluster config", node.name());
                 format!("cannot find node: {} in cluster config", node.name())
             })?;
+        if self.vdisks().iter().any(|vdisk| vdisk.replicas().len() < node.quorum()) {
+            return Err(String::from("Quorum is more than amount of replicas for at least one VDisk"));
```
ikopylov, line 291: `vdisk` name should be specified inside the error description. This will make it easier to find and fix the problem

Justarone: Done in new commit!

ikopylov:
LGTM

PR #133, opened 2020-11-11
Allow different keys

idruzhitskiy:


ikopylov:
This request contains a breaking change in network protocol, so it should be delayed until C# BobClient is updated

PR #137, opened 2020-11-25
Change dipstick crate for metrics crate

### LINKED ISSUE #129: Consider using 'metrics' crate for performance counters
This crate doesn't support Graphite out of the box, so it is required to implement metrics delivery by hand

Justarone:
closes #129 

This solution solves our problem:
* I do aggregation by myself, so now counters don't reset on flush;
* I memorize last values of metrics and send them regularly, even they are not updated.


Justarone:
ATTENTION!

This PR is not ideal and is not ready to be merged at all. I need someone to instruct me "what is good and what is bad".

Cargo.toml
```diff
@@ -40,6 +40,7 @@ derive-new = "0.5.8"
 ctrlc = "3.1"
 lazy_static = "1.4.0"
 thiserror = "1.0"
+metrics = { version = "0.12.1", features = ["std"] }
```
pyakushin, line 43: [`std`](https://github.com/metrics-rs/metrics/blob/main/metrics/Cargo.toml#L45) is a default feature, you don't need to specify it

src/core/backend/pearl/holder.rs
```diff
@@ -148,17 +148,17 @@ impl Holder {
...
-            PEARL_GET_COUNTER.count(1);
-            let timer = PEARL_GET_TIMER.start();
+            counter!(PEARL_GET_COUNTER, 1);
+            let timer = Instant::now();
             storage
                 .read(Key::from(key))
                 .await
                 .map(|r| {
-                    PEARL_GET_TIMER.stop(timer);
+                    counter!(PEARL_GET_TIMER, timer.elapsed().as_nanos() as u64);
```
pyakushin, line 157: We need to count all get requests. Here you count only Ok results

Cargo.toml
Justarone: I don't know why, but it doesn't find `set_boxed_recorder` method without explicit `std` feature. We have to use boxed recorder because we use tcp connection to send metrics, so we build our recorder in runtime (it's possible to use static recorder for simple LogRecorder for example, that's not our case). In metrics lib exporters examples tcp exporter also implements boxed recorder instead of static one.

src/core/backend/pearl/holder.rs
Justarone: I didn't touch this part of code, but it seems that it's bug, I fixed that.

Cargo.toml
pyakushin: Ok, you are right

src/core/metrics/exporter/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+        self
+    }
+
+    pub(crate) fn install(self) -> Result<(), Error> {
+        let recorder = self.build()?;
+        metrics::set_boxed_recorder(Box::new(recorder))?;
+        Ok(())
+    }
+
+    pub(crate) fn build(self) -> Result<GraphiteRecorder, Error> {
```
pyakushin, line 92: this function's return value is unnecessarily wrapped by `Result`

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,103 @@
...
+        let addresses = addresses.to_socket_addrs()?.collect();
+        let mut socket = RetrySocket {
+            retries: 0,
+            next_try: Instant::now() - Duration::from_millis(MIN_RECONNECT_DELAY_MS),
+            addresses,
+            socket: None,
+        };
+
+        // try early connect
+        let _ = socket.flush().ok();
```
pyakushin, line 40: I think it's better to print error msg with `info` or `warn` level`. Or maybe it is unnecessary, please check.

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,103 @@
+//! A TCP Socket wrapper that reconnects automatically.
+
+use log::{info, warn};
+use std::fmt;
+use std::io;
+use std::io::Write;
+use std::net::TcpStream;
```
pyakushin, line 7: It is better to use tokio async TcpStream because bob is built as an async app

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,103 @@
+//! A TCP Socket wrapper that reconnects automatically.
+
+use log::{info, warn};
```
pyakushin, line 3: We have global macro import, it is redundant here

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,103 @@
...
+                socket.set_nonblocking(true)?;
+                self.retries = 0;
+                info!("Connected to {:?}", addresses);
+                self.socket = Some(socket);
+            }
+        }
+        Ok(())
+    }
+
+    fn backoff(&mut self, e: io::Error) -> io::Error {
```
pyakushin, line 61: We have import `use std::io::Error as IOError`

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        let opres = if let Some(ref mut socket) = self.socket {
+            operation(socket)
+        } else {
+            // still none, quiescent
+            return Err(io::Error::from(io::ErrorKind::NotConnected));
+        };
+
+        match opres {
```
pyakushin, line 88:
```
opres.map_err(|e| self.backoff(e))
```

src/core/metrics/exporter/retry_socket.rs
pyakushin: You can skip part with implementation of AsyncWrite/AsyncRead for RetrySocket, if it seems too complicated

src/core/metrics/exporter/send.rs
```diff
@@ -0,0 +1,152 @@
+use log::{debug, trace};
+use std::collections::HashMap;
+use std::io::Write;
+use std::sync::mpsc::Receiver;
```
pyakushin, line 4: Switch to async mpsc channel

pyakushin (changes requested):

src/core/metrics/exporter/mod.rs
Justarone: You're right, changed that

src/core/metrics/exporter/retry_socket.rs
Justarone: Removed!

src/core/metrics/exporter/retry_socket.rs
Justarone: Changed it

src/core/metrics/exporter/retry_socket.rs
Justarone: Did it!

src/core/metrics/exporter/send.rs
Justarone: I have no idea, how to do it:( 

metrics crate has sync trait with sync functions for recorder (which has functions for sending every single metric) and I need to call from sync method another method of object, which is not sync (async `push_metrics` from sync `increment_counter`). I don't understand how to send do it (can't do method of object async)

src/core/metrics/exporter/retry_socket.rs
Justarone: I am not sure that we do need this now because the code of this `RetrySocket` I found in dipstick - crate, which we used for metrics before. Maybe we should leave this task in backlog for now?

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,118 @@
...
+            socket: State::Task(receiver),
+        }
+    }
+
+    fn spawn_task(address: SocketAddr) -> Receiver<TcpStream> {
+        let (tx, rx) = channel();
+        tokio::spawn(async move {
+            let mut delay = MIN_RECONNECT_DELAY_MS;
+            loop {
+                if let Ok(stream) = TcpStream::connect(address).await {
```
pyakushin, line 45: I think it's better to wrap `connect` into `tokio::time::timeout`. Because now it will just block current task. Or it works differently?

src/core/metrics/exporter/retry_socket.rs
```diff
@@ -0,0 +1,118 @@
...
+                    if let Err(e) = tx.send(stream) {
+                        warn!("Failed to send stream from task: {:?}", e);
+                    }
+                    return;
+                }
+
+                delay = MAX_RECONNECT_DELAY_MS.min(delay << 1);
+                let mut conn_interval = interval(Duration::from_millis(delay));
```
pyakushin, line 53: And for the future: you should create interval before loop and call tick on it inside.

pyakushin (approved):

src/core/metrics/exporter/retry_socket.rs
Justarone: Hm, on tiny example:
```rust
use tokio::net::TcpStream;

#[tokio::main]
async fn main() {
    let res = TcpStream::connect("127.0.0.1:2003").await;
    println!("{:?}", res);
}
```
This error occurs immediately:
```
Err(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
```
So I think, there is no reason to wrap this in `timeout`

src/core/metrics/exporter/retry_socket.rs
Justarone: This interval is dynamic: it grows from min time (for example its value is 5) to max time (for example: 1000).
On first try (to connect) interval would be 5, on second - 10, on third - 20 and etc. On the ninth try this interval will become stable and equal to max time (1000). So I can't move it outside of the loop with this condition (But I would able to do this, if I make this interval constant)

PR #140, opened 2020-11-26
Parallel initialization of disks

### LINKED ISSUE #138: Run storage initialization for different physical disks in parallel
Currently, initialization runs sequentially, which leads to slow startup. But the data is stored on different physical disks, so we can easily parallelize the startup across these disks. In addition, I suggest to limit the degree of parallelism to some value (specified in the configuration), because we still have a shared resources: data bus and CPU.

idruzhitskiy:
Closes #138

### DIFF
--- a/dcr_config.yaml
+++ b/dcr_config.yaml
@@ -21,3 +21,5 @@ cleanup_interval: "15sec"
 open_blobs_soft_limit: 2
 # Hard limit for open blobs
 open_blobs_hard_limit: 2
+# Degree of parallelism in disks initialization
+init_par_degree: 4
--- a/src/core/backend/pearl/core.rs
+++ b/src/core/backend/pearl/core.rs
@@ -10,6 +10,7 @@ pub(crate) struct Pearl {
     alien_vdisks_groups: Arc<RwLock<Vec<Group>>>,
     pearl_sync: Arc<SyncState>, // holds state when we create new alien pearl dir
     node_name: String,
+    init_par_degree: usize,
 }
 
 impl Pearl {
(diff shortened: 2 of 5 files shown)


dcr_config.yaml
```diff
@@ -21,3 +21,5 @@ cleanup_interval: "15sec"
 open_blobs_soft_limit: 2
 # Hard limit for open blobs
 open_blobs_hard_limit: 2
+# Degree of parallelism in disks initialization
+init_par_degree: 4
```
pyakushin, line 25: I think it's better to name it like `init_threads` or smth like that

src/core/backend/pearl/core.rs
```diff
@@ -79,14 +81,48 @@ impl Pearl {
 #[async_trait]
 impl BackendStorage for Pearl {
     async fn run_backend(&self) -> Result<()> {
+        use futures::future::join_all;
```
pyakushin, line 84: I think it's better to use tokio::spawn and wait on JoinHandlers to check result. I'm not sure that join_all provides true threaded parallelism. You can create generic init func, eg `init_groups(group: Vec<Group>, ...)`, and spawn it for each bunch

src/core/configs/node.rs
```diff
@@ -349,6 +349,7 @@ pub struct Node {
     cleanup_interval: String,
     open_blobs_soft_limit: Option<usize>,
     open_blobs_hard_limit: Option<usize>,
+    init_par_degree: Option<usize>,
```
pyakushin, line 352: It's better to use attr `#[serde(default = "default_fn")]` on struct field. And set the default value to 1

pyakushin (changes requested):

dcr_config.yaml
ikopylov: `degree of parallelism`  is the standard naming for such parameters. `threads` is less obvious

PR #158, opened 2021-01-20
Add alien directory new configuration

### LINKED ISSUE #146: Allow to configure the location of alien directory
It would be nice to be able to place `alien` directory at different disk

Justarone:
Closes #146 

I need somebody to review it, cause I am not sure, that I didn't forget anything...

### DIFF
--- a/src/core/backend/pearl/settings.rs
+++ b/src/core/backend/pearl/settings.rs
@@ -12,7 +12,12 @@ pub(crate) struct Settings {
 impl Settings {
     pub(crate) fn new(config: &NodeConfig, mapper: Arc<Virtual>) -> Self {
         let config = config.pearl().clone();
-        let alien_folder = config.alien_folder().to_owned().into();
+        let disk_path = mapper
+            .get_disk(&config.alien_disk())
+            .expect("cannot find alien disk in config")
+            .path();
+        let alien_folder =
+            format!("{}/{}/", disk_path, config.settings().alien_root_dir_name()).into();
 
         Self {
             bob_prefix_path: config.settings().root_dir_name().to_owned(),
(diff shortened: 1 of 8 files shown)


src/core/backend/pearl/settings.rs
```diff
@@ -65,7 +57,8 @@ impl Settings {
                 for vdisk_id in vdisks {
                     if let Ok((entry, vdisk_id)) = self.try_parse_vdisk_id(vdisk_id) {
                         if self.mapper.is_vdisk_on_node(&node_name, vdisk_id) {
-                            let disk_name = config.pearl().alien_disk().to_owned();
+                            // name of the disk is not necessary for alien group
+                            let disk_name = String::from("alien_disk");
```
pyakushin, line 61: I think it's better to leave an option to choose disk for alien folder. If disk is set: alien_full_path = disk_path + alien_path
if not set: alien_full_path = alien_path.

pyakushin (approved):

src/core/backend/pearl/settings.rs
```diff
@@ -12,12 +12,16 @@ pub(crate) struct Settings {
...
     pub(crate) fn new(config: &NodeConfig, mapper: Arc<Virtual>) -> Self {
         let config = config.pearl().clone();
-        let disk_path = mapper
-            .get_disk(&config.alien_disk())
-            .expect("cannot find alien disk in config")
-            .path();
-        let alien_folder =
-            format!("{}/{}/", disk_path, config.settings().alien_root_dir_name()).into();
+        let alien_disk = config.alien_disk();
+        let alien_folder = if alien_disk.is_empty() {
```
pyakushin, line 16: If I remember it right, alien_disk won't be empty, if will be `~`. And lets wrap into Option<String>, because it is an optional parameter now.

Justarone: I set method for default alien_disk, which returns empty string, so it will be empty, but, you are right, that solution with option is more elegant, I'll fix that

PR #167, opened 2021-02-09
fix dump sems not accounting for disk names

idruzhitskiy:


src/core/mapper.rs
```diff
@@ -14,6 +16,8 @@ pub struct Virtual {
     disks: Vec<DiskPath>,
     vdisks: VDisksMap,
     nodes: NodesMap,
+    disk_access_sems: Arc<std::sync::RwLock<SemMap>>,
```
pyakushin, line 19: Why do you use std RwLock instead of async?

idruzhitskiy: Used this one for tests. Changed to async rwlock

PR #169, opened 2021-02-17
add exact flag

### LINKED ISSUE #163: Add the ability to set the exact number of vdisks in the cluster config generation tool

idruzhitskiy:
closes #163

### DIFF
--- a/src/bin/ccg.rs
+++ b/src/bin/ccg.rs
@@ -279,10 +279,12 @@ fn lcm(a: usize, b: usize) -> usize {
 fn simple_gen(
     mut config: ClusterConfig,
     replicas_count: usize,
-    vdisks_count: usize,
+    mut vdisks_count: usize,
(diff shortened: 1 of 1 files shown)


src/bin/ccg.rs
```diff
@@ -358,6 +360,10 @@ fn write_to_file(mut output: String, name: String) {
     debug!("write to file: OK");
 }
 
+fn exact_vdisks_count() -> bool {
+    get_matches().is_present("exact_vdisks_count")
```
pyakushin, line 364: I think it redundant to have one line fn with call from the only place

pyakushin:
LGTM

src/bin/ccg.rs
idruzhitskiy: It's here to avoid string constants in large functions, and in case of future reuse

PR #170, opened 2021-02-17
Fix metrics prefix

Justarone:
Now we are able to set root dir for metrics optionally + all metrics from one node are in the dir `<node_name>/<ip>/`


Justarone:
Maybe we should also include pattern string to set graphite paths in more flexible way?

pyakushin (approved):

PR #176, opened 2021-02-24
Add cluster expansion to ccg

### LINKED ISSUE #172: The cluster config generation tool must be able to update the configuration for cluster expansion
The tool should receive as input the existing cluster config and the new hardware structure for the expanded cluster. As a result, it should produce the updated cluster configuration. This configuration should be such that the number of VDisks to move between servers is minimal, and at the same time, the requirement of placing VDisk replicas on different servers must be met.

vovac12:
Closes #172


src/bin/ccg.rs
```diff
@@ -1,24 +1,36 @@
...
+use config_cluster_generator::{
+    center::{get_new_disks, get_pairs_count, get_structure},
+    utils::{init_logger, lcm, read_config_from_file, write_to_file},
+};
 
 #[tokio::main]
 async fn main() {
     init_logger();
-    if let Some(output) = read_from_file().and_then(generate_config) {
+    match dbg!(get_matches().subcommand()) {
```
pyakushin, line 16: It is better to log with env_logger

src/bin/ccg.rs
```diff
@@ -28,261 +40,77 @@ async fn main() {
...
 
 fn simple_gen(
     mut config: ClusterConfig,
     replicas_count: usize,
     mut vdisks_count: usize,
+    exact_vdisks_count: bool
 ) -> ClusterConfig {
     let center = get_structure(&config);
-    if !exact_vdisks_count() {
+    if !exact_vdisks_count {
```
pyakushin, line 113: I suggest `vdisks_counts_match` or smth like that

src/bin/config_cluster_generator/center.rs
```diff
@@ -0,0 +1,309 @@
...
+        vdisk
+    }
+}
+
+#[derive(Debug)]
+pub struct Rack {
+    name: String,
+    used_count: AtomicUsize,
+    nodes: Vec<Node>,
+    old: bool,
```
pyakushin, line 120: I think `is_old` is more clear

pyakushin (approved):

pyakushin (commented):
CCG should return error if input hardware configuration is less than in input config.

Issue #179, opened 2021-02-24
Add performance counters to track disk status


ikopylov:
Add a counter for every physical disk that shows its status in the form: 0 (offline) or 1 (online).

Add a counter with the number of working disks on the node.

Justarone:
27.02 - 2h (metrics + pearl level error processing)

Justarone:
28.02 - 2h (add third state for disk metric and change state management logic)

Issue #177, opened 2021-02-24
Bring disk online by API request (not automatically)


ikopylov:
Automatic procedure for bringing disk online is dangerous, because there is no control over the disk content. It could be a disk from nearby server with some irrelevant bob data, that should be removed before being used

Justarone:
27.02 - 5h (api method and code restructuring)

PR #182, opened 2021-02-27
Add api method to start disk

Justarone:


Justarone:
Closes #179 and #177

src/api/http.rs
```diff
@@ -139,19 +148,31 @@ fn not_acceptable_backend() -> Status {
...
-    let groups = backend.vdisks_groups().ok_or_else(not_acceptable_backend)?;
+    let (dcs, _) = backend
+        .disk_controllers()
+        .ok_or_else(not_acceptable_backend)?;
     debug!("get vdisks groups: OK");
-    groups
+    let needed_dc = dcs
         .iter()
-        .find(|group| group.vdisk_id() == vdisk_id)
+        .find(|dc| dc.vdisks().iter().find(|&&vd| vd == vdisk_id).is_some())
```
pyakushin, line 161: I think it's this way to find needed dc is more clear:
```
.flat_map(|dc| dc.vdisks())
.find(|&&vd| vd == vdisk_id)
```

src/api/http.rs
```diff
@@ -198,10 +219,85 @@ fn nodes(bob: State<BobServer>) -> Json<Vec<Node>> {
...
+            disks.push(DiskState {
+                name: disk_path.name().to_owned(),
+                path: disk_path.path().to_owned(),
+                is_active: dc.is_ready().await,
+            });
+        }
+        disks
+    };
+
+    let rt = Runtime::new().expect("create runtime");
```
pyakushin, line 242: There is `runtime()` helper

src/api/http.rs
```diff
@@ -198,10 +219,85 @@ fn nodes(bob: State<BobServer>) -> Json<Vec<Node>> {
...
+        .chain(std::iter::once(&adc))
+        .filter(|dc| dc.disk_name() == disk_name)
+        .map(|dc| dc.run())
+        .collect::<FuturesUnordered<_>>();
+    if target_dcs.len() == 0 {
+        let err = format!("Disk Controller with name '{}' not found", disk_name);
+        warn!("{}", err);
+        return Err(StatusExt::new(Status::NotFound, false, err));
+    }
+    let rt = Runtime::new().expect("create runtime");
```
pyakushin, line 274: `runtime()`

src/api/http.rs
```diff
@@ -198,10 +219,85 @@ fn nodes(bob: State<BobServer>) -> Json<Vec<Node>> {
...
+    match rt.block_on(task) {
+        Ok(()) => {
+            let msg = format!(
+                "all disk controllers for disk '{}' successfully started",
+                disk_name
+            );
+            info!("{}", msg);
+            Ok(StatusExt::new(Status::Ok, true, msg))
+        }
+        Err(e) => Err(StatusExt::new(Status::Ok, false, e.to_string())),
```
pyakushin, line 299: Why do you return Status::Ok in case of an error?

src/core/backend/pearl/core.rs
```diff
@@ -128,32 +156,32 @@ impl BackendStorage for Pearl {
...
+    async fn close_unneeded_active_blobs(&self, soft: usize, hard: usize) {
         for dc in self.disk_controllers.iter() {
-            cnt += dc.index_memory().await
+            dc.close_unneeded_active_blobs(soft, hard).await;
         }
-        cnt += self.alien_disk_controller.index_memory().await;
-        cnt
+    }
+
+    async fn collect_metrics(&self) -> BackendMetrics {
```
pyakushin, line 181: Are you sure it is the good way to collect all metrics at once? Because if we need one metric, we still will block at least three times more.

src/core/backend/pearl/disk_controller/logger.rs
```diff
@@ -0,0 +1,43 @@
+use super::*;
+use chrono::Local;
+use std::fs::{File, OpenOptions};
```
pyakushin, line 3: I suggest to use tokio FIle. Or there is a reason to use sync version?

src/core/backend/pearl/group.rs
```diff
@@ -397,6 +391,59 @@ impl Group {
...
+        let mut is_small = vec![];
+        for h in &close {
+            is_small.push(h.active_blob_is_small().await);
+        }
+
+        let mut close: Vec<_> = close.into_iter().enumerate().collect();
+        Self::sort_by_priority(&mut close, &is_small);
+
+        while close.len() > hard {
+            let (_, holder) = close.pop().unwrap();
```
pyakushin, line 427: I think it's better to use `expect()` with panic description in not dev code

src/api/http.rs
Justarone: That's not what I actually want. When you do `flat_map` you start to iterate through vdisks ids and `find` method is working with iterator through this ids (so it'll return id), but I need `DiskController`

src/api/http.rs
Justarone: There may be 2 kinds of errors:
1) The requested disk_name doesn't exist
2) The requested disk can't start
I decided to return Status::NotFound for the first kind of errors, and false on `ok` flag in `StatusExt` struct, when disk can't start, but I'll change it on Status::InternalServerError

src/core/backend/pearl/core.rs
Justarone: I've tried to reduce number of functions in `BackendStorage` trait, because it becomes too bloated. But you're right, that the decision to place all metrics in one function call is weird, so I decided to separate metrics functions with another trait, and require it in `Backend` struct in `inner` field (I don't know about dependencies: should `BackendStorage` trait depend on `MetricsProducer` or not or vice versa, so I decided to make the independent)

src/core/backend/pearl/disk_controller/logger.rs
Justarone: The only thing, why I did this file sync is that we don't need to change interface of other function. If file is tokio, then we need `async new` (to create tokio File) -> `Pearl::new` becomes `async` -> `Backend::new` becomes `async` -> `Grinder::new` becomes `async`.
There is no problem in that, but disks events are so rare (I hope), that we won't see any difference between std and tokio `File`. But to keep `async` model of this lib I'll replace std `File` on tokio one

Justarone: One more possible solution to spawn runtime in one of this `new` methods (in `DisksEventsLogger`, for example) and make it sync, but that seems weird (it won't hit performance, because the creation of Logger is done once during application startup)

pyakushin (approved):
LGTM

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -216,28 +303,28 @@ impl DiskController {
...
-                    *self.state.write().await = GroupsState::NotReady;
-                    *self.groups.write().await = Vec::new();
+    async fn groups_run(&self) -> AnyResult<()> {
+        let state = self.state.read().await.clone();
+        match state {
+            GroupsState::Initialized => {
+                let res = {
+                    let _permit = self.run_sem.acquire().await.expect("Semaphore is closed");
+                    Self::run_groups(self.groups.clone()).await
+                };
```
idruzhitskiy, line 313: map and map_err. And probably separate function to keep "match" clean

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -271,20 +356,21 @@ impl DiskController {
...
+                    .cloned()
+            };
             if let Some(group) = vdisk_group {
-                let res = group.put(key, data).await;
-                // TODO: check if pearl WorkDirUnavailable error occured and change state
-                res.map_err(|e| {
-                    debug!("PUT[{}], error: {:?}", key, e);
-                    Error::failed(format!("{:#?}", e))
-                })
+                match group.put(key, data).await {
```
idruzhitskiy, line 367: map_err?

bob-backend/src/pearl/disk_controller/logger.rs
```diff
@@ -0,0 +1,55 @@
...
+use tokio::io::AsyncWriteExt;
+
+#[derive(Clone, Debug)]
+pub(crate) struct DisksEventsLogger {
+    fd: Arc<RwLock<File>>,
+}
+
+impl DisksEventsLogger {
+    pub(crate) async fn new(filename: impl AsRef<Path>) -> Self {
+        let fd = if filename.as_ref().exists() {
```
idruzhitskiy, line 13: Probably new should return Result (or Maybe), so we avoid panics here (and can start server even if events can't be logged, simply printing errors)

bob-common/src/configs/node.rs
```diff
@@ -310,6 +312,14 @@ impl Pearl {
         true
     }
 
+    pub fn disks_events_logfile(&self) -> &str {
+        &self.disks_events_logfile
+    }
+
+    fn default_disks_events_logfile() -> String {
+        "/tmp/bob_events.csv".to_owned()
```
idruzhitskiy, line 320: IMO we should try to get OS tmp dir here, and fallback to /tmp in case of no env set

bob/src/api/http.rs
```diff
@@ -155,19 +165,31 @@ fn not_acceptable_backend() -> Status {
...
+            StatusExt::new(Status::NotFound, false, err)
+        })?;
+    let task = async move {
+        needed_dc.vdisk_group(vdisk_id).await.map_err(|_| {
+            let err = format!("Disk Controller with vdisk #{} not found", vdisk_id);
             warn!("{}", err);
             StatusExt::new(Status::NotFound, false, err)
         })
+    };
+    let rt = Runtime::new().expect("create runtime");
```
idruzhitskiy, line 191: runtime()

bob-backend/src/pearl/disk_controller.rs
Justarone: I agree with you in part with separation, but I don't understand, how I can use `map` and `map_err` (because they get closuse, which is not `async`, but I need to perform async operations)

bob-backend/src/pearl/disk_controller.rs
Justarone: Problem with async operations again

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -81,52 +106,90 @@ impl DiskController {
...
         }
     }
 
-    pub(crate) async fn init(&self) -> BackendResult<()> {
-        if *self.state.read().await == GroupsState::NotReady {
+    async fn change_state(&self, new_state: GroupsState) {
+        let mut wlock = self.state.write().await;
+        let old_state = wlock.clone();
+        match (old_state, &new_state) {
+            (old_state, GroupsState::NotReady) if old_state != GroupsState::NotReady => {
```
idruzhitskiy, line 134: You can check equality of states before match to avoid clone and ifs here

pyakushin (approved):

PR #186, opened 2021-03-04
Use racks in ccg

### LINKED ISSUE #173: The cluster configuration tool can take server racks into account during generation
The network inside the rack is faster, but at the same time, the entire rack is a common point of failure in case of power problems. A good first guess would be to place at least one replica of vDisk on another server rack if possible

vovac12:
Closes #173


bob-tools/bin/config_cluster_generator/center.rs
```diff
@@ -0,0 +1,501 @@
...
+        let (node, disk) = first_rack.next_disk(&[]).expect("no disks in setup");
+        vdisk.push_replica(Replica::new(node.name.clone(), disk.name.clone()));
+
+        while vdisk.replicas().len() < replicas_count {
+            let banned_nodes = get_used_nodes_names(vdisk.replicas());
+            let (node, disk) = if vdisk.replicas().len() == 2 {
+                first_rack.next_disk(&banned_nodes).map_or_else(
+                    || {
+                        self.next_rack()
+                            .expect("no racks in setup")
```
idruzhitskiy, line 109: You should make "new" method accept racks, and check there it's emptyness. This way you can omit "Option" in "next_" methods, as you'll be sure it's not empty

idruzhitskiy: And "new" should  return Result<..Center>, so you can log error

idruzhitskiy (changes requested):
1. There should be Result+Error Message instead of None+Logs
2. There are hiddent mutability in center, rack etc methods (.inc()). It should be avoided, either by moving this counts to center (couple of Map<.., usize>), or marking those methods mutable. IMO first option is better and more maintainable

bob-tools/bin/config_cluster_generator/center.rs
```diff
@@ -0,0 +1,597 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    ops::AddAssign,
+};
+
+const REPLICA_IN_FIRST_RACK: usize = 2;
+
+#[derive(Debug, Default)]
+struct Counter {
+    racks_used_count: HashMap<String, usize>,
```
idruzhitskiy, line 15: IMO it's better to use type aliases here (like `type RackName = String`), it'll be more readable

idruzhitskiy (commented):
Functions probably too large and should be split, otherwise LGTM

agend:
It just hit me that term rack is a bit weird,especially for cloud environments. Would it make sense to rename to availability group or something similar?

pyakushin (changes requested):
Can you provide an example of cluster config with racks?

vovac12:
> Can you provide an example of cluster config with racks?

Done

pyakushin (approved):
LGTM

PR #189, opened 2021-03-12
HTTP/GRPC API users configuration

pyakushin:
- [ ] secure storage of passwords

### DIFF
--- a/bob/src/api/mod.rs
+++ b/bob/src/api/mod.rs
@@ -3,9 +3,75 @@ pub mod http;
 
 use crate::server::Server as BobServer;
 
-use bob_common::configs::User;
+use bob_common::configs::{Access as AccessConfig, User as UserConfig};
+use rocket::{
+    http::Status,
+    request::{FromRequest, Outcome},
+    Request,
+};
 
+const BOB_API_USER_HEADER: &str = "x-bob-user-key";
+const BOB_API_PASSWORD_HEADER: &str = "x-bob-user-key";
+
+#[derive(Debug)]
 pub struct Api {
     bob: BobServer,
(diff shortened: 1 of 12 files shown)


config-examples/users.yaml
```diff
@@ -0,0 +1,4 @@
+users:
+  - name: default
+    password: default
```
agend, line 3: never ever store password in plain string format.

config-examples/users.yaml
```diff
@@ -0,0 +1,4 @@
...
+    perms: Read
```
agend, line 4: It is not a good approach to creat internal user managment. Could create a lots of security issue. It's better to authorize via SAML, OpenID Connect.

bob/src/api/mod.rs
```diff
@@ -1,2 +1,77 @@
...
+
+use bob_common::configs::{Access as AccessConfig, User as UserConfig};
+use rocket::{
+    http::Status,
+    request::{FromRequest, Outcome},
+    Request,
+};
+
+const BOB_API_USER_HEADER: &str = "x-bob-user-key";
+const BOB_API_PASSWORD_HEADER: &str = "x-bob-user-key";
```
agend, line 14: same constant

config-examples/users.yaml
agend: bob is meant to be distributed. managing user in same way is not a nice.

agend:
I think it's better to create gateway service

config-examples/users.yaml
pyakushin: Of course (: It's the first iteration to figure out the whole feature organization.

PR #193, opened 2021-03-30
Amazon S3 support: GetObject, PutObject

### LINKED ISSUE #185: Investigate the possibility of implementing AWS S3 API
https://docs.aws.amazon.com/AmazonS3/latest/API/Welcome.html

vovac12:
Closes #185


bob-common/src/error.rs
```diff
@@ -6,7 +6,7 @@ use crate::data::{BobKey, VDiskId};
 
 #[derive(Debug, Clone, ErrorTrait)]
 pub struct Error {
-    ctx: Kind,
+    pub ctx: Kind,
```
pyakushin, line 9: I think you shouldn't change visibility of this field

bob/src/api/http.rs
```diff
@@ -485,6 +494,34 @@ async fn read_directory_children(mut read_dir: ReadDir, name: String, path: Stri
...
+    data.open()
+        .read_to_end(&mut data_buf)
+        .map_err(|err| -> StatusExt { err.into() })?;
+    let data = BobData::new(
+        data_buf,
+        BobMeta::new(chrono::Local::now().timestamp() as u64),
+    );
+
+    let opts = BobOptions::new_put(None);
+    runtime()
```
pyakushin, line 518: we now have runtime handler in `BobServer` so you should run async tasks with `bob.block_on(async { . . })`
Here and in other places below

bob/src/api/http.rs
```diff
@@ -485,6 +494,34 @@ async fn read_directory_children(mut read_dir: ReadDir, name: String, path: Stri
     }
 }
 
+#[get("/data/<key>")]
+fn get_data(bob: State<BobServer>, key: BobKey) -> Result<Content<Vec<u8>>, StatusExt> {
+    let opts = BobOptions::new_get(None);
+    let result = runtime()
+        .block_on(async { bob.grinder().get(key, &opts).await })
+        .map_err(|err| -> StatusExt { err.into() })?;
```
pyakushin, line 502: Are you sure that `?` doesn't perform conversion automatically? 
Because you have `impl From<Error> for StatusExt`
Here and in other places below

bob/src/api/http.rs
```diff
@@ -523,3 +560,41 @@ impl From<Status> for StatusExt {
         }
     }
 }
+
+impl From<std::io::Error> for StatusExt {
```
pyakushin, line 564: I think it's better to import `use std::io::Error as IoError`

bob/src/api/http.rs
```diff
@@ -523,3 +560,41 @@ impl From<Status> for StatusExt {
...
+                ErrorKind::NotFound => Status::NotFound,
+                _ => Status::BadRequest,
+            }, // TODO: Complete match
+            ok: false,
+            msg: err.to_string(),
+        }
+    }
+}
+
+impl From<bob_common::error::Error> for StatusExt {
```
pyakushin, line 577: I think it's better to import `use bob_common::error::Error as BobError`

bob/src/api/http/s3.rs
```diff
@@ -0,0 +1,210 @@
...
+pub(crate) fn get_object(
+    bob: State<BobServer>,
+    key: BobKey,
+    headers: GetObjectHeaders,
+) -> Result<GetObjectOutput, StatusS3> {
+    let opts = BobOptions::new_get(None);
+    let data = super::runtime()
+        .block_on(async { bob.grinder().get(key, &opts).await })
+        .map_err(|err| -> StatusExt { err.into() })?;
+    let content_type = match headers.content_type {
```
pyakushin, line 105: I think it's better to use `.unwrap_or_else(|| { . . })`

bob/src/api/http/s3.rs
```diff
@@ -0,0 +1,210 @@
...
+        Some(t) => t,
+        None => infer_data_type(&data),
+    };
+    let last_modified = data.meta().timestamp();
+    match headers.if_modified_since {
```
pyakushin, line 110: I suggest to use if let statement here and below

pyakushin (changes requested):

bob/src/api/http.rs
vovac12: Fixed in `http.rs`. Conversion doesn't perform in some places in `s3.rs`

PR #197, opened 2021-04-09
Fix backend storage trait object safety issue

pyakushin:


bob-backend/src/core.rs
```diff
@@ -80,7 +80,7 @@ impl Operation {
 
 #[async_trait]
 pub trait BackendStorage: Debug {
```
idruzhitskiy, line 82: Wouldn't it be easier to make BackendStorage: Debug + Send? We can left default implementations then

pyakushin: :facepalm: I tried to add Send, but it didn't work. But adding Sync worked

PR #217, opened 2021-06-12
Access management

pyakushin:
Main goals: 
- [x] add service to control incoming GRPC requests
- [x] automatically recognize GRPC requests from cluster nodes
[X] access control for HTTP API requests (moved to issue #319)


bob-access/src/authenticator.rs
```diff
@@ -0,0 +1,20 @@
+use crate::{credentials::Credentials, error::Error};
+
+pub trait Authenticator: Clone {
+    fn check_credentials(&self, credentials: Credentials) -> Result<(), Error>;
```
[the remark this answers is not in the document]
pyakushin: It can be easily updated, so for now we don't need async api

bob-access/src/settings.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(Debug)]
+pub struct Settings {}
+
+pub trait AccessStorage {
+    fn load(&self) -> Settings;
```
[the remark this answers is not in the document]
pyakushin: We can impl this trait for file storage/DB/remote auth server

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Debug, Default, Clone)]
+pub struct StubExtractor {}
+
+impl StubExtractor {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl<Request> Extractor<Request> for StubExtractor {
```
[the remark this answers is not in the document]
pyakushin: Outdated, for now we can't use services for rocket based rest api

bob-access/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::task::{Context, Poll};
+
+use authenticator::Authenticator;
+use extractor::Extractor;
+use tonic::transport::NamedService;
+use tower::{Layer, Service};
+
+#[derive(Debug, Default)]
+pub struct AccessControlLayer<A, E> {
+    authenticator: Option<A>,
```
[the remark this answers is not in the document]
pyakushin: Well, if one of it doesn't set, bob will start with warning that authentication doesn't work

bob-access/src/settings.rs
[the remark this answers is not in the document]
idruzhitskiy: But there is neither implementation of these trait nor generic constraints with these trait, right?

bob-apps/bin/bobd.rs
```diff
@@ -30,9 +38,9 @@ async fn main() {
     println!("Cluster config: {:?}", cluster_config);
     let cluster = ClusterConfig::try_get(cluster_config).await.unwrap();
 
-    let node_config = matches.value_of("node").unwrap();
-    println!("Node config: {:?}", node_config);
-    let node = cluster.get(node_config).await.unwrap();
+    let node_config_file = matches.value_of("node").unwrap();
+    println!("Node config: {:?}", node_config_file);
```
idruzhitskiy, line 42: log?

bob-apps/bin/bobd.rs
```diff
@@ -100,14 +108,78 @@ async fn main() {
...
-    Server::builder()
-        .tcp_nodelay(true)
-        .add_service(new_service)
-        .serve(addr)
-        .await
-        .unwrap();
+    let authentication_type = matches.value_of("authentication_type").unwrap();
+    match authentication_type {
+        "stub" => {
+            let bob_service = BobApiServer::new(bob);
```
idruzhitskiy, line 114: Both branches with different authenticators have plenty in common, I think it should be extracted into the function

bob-access/src/authenticator/users_storage/config.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::error::Error;
+
+use super::{Perms, User};
+
+#[derive(Debug, Clone, Deserialize)]
+pub(super) struct ConfigUser {
+    pub(super) username: String,
+    pub(super) password: String,
+    pub(super) role: String,
```
idruzhitskiy, line 11: Don't you think user should be allowed to have multiple roles? We can remove `Admin` role in example then, and just leave `reader` and `writer`

bob-access/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "bob-access"
+version = "2.0.0-alpha.3"
```
ikopylov, line 3: Version should be updated

bob-access/src/authenticator/basic.rs
```diff
@@ -0,0 +1,72 @@
+use std::{collections::HashMap, net::IpAddr};
+
+use crate::{credentials::Credentials, error::Error};
+
+use super::{users_storage::UsersStorage, Authenticator};
+
+#[derive(Debug, Default, Clone)]
+pub struct Basic<Storage: UsersStorage> {
+    users_storage: Storage,
+    nodes: HashMap<IpAddr, Credentials>,
```
ikopylov, line 10: Node address can be specified not only as an Ip-address, but also as a hostname. Does the current logic handle hostnames properly?

bob-access/src/authenticator/users_storage/config.rs
ikopylov: We implement a claim-based policy. Role in this case is just a group of claims, so we don't need to support multiple roles. What we need is the ability to specify claims directly for each user. In the first version we can keep the current logic, and I will create a new issue to add this ability in the future

ikopylov: https://github.com/qoollo/bob/issues/408

bob-access/src/authenticator/basic.rs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            self.nodes = nodes;
+            Ok(())
+        } else {
+            let message = "nodes credentials missing ip or username";
+            Err(Error::credentials_not_provided(message))
+        }
+    }
+
+    fn is_node(&self, other: &Credentials) -> Option<bool> {
```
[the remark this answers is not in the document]
pyakushin: Done

bob-access/src/authenticator/basic.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    pub fn set_nodes_credentials(
+        &mut self,
+        nodes: HashMap<IpAddr, Credentials>,
+    ) -> Result<(), Error> {
+        if nodes
+            .values()
+            .all(|cred| cred.ip().is_some() && cred.username().is_some())
```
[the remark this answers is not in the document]
pyakushin: Can you add some example, what do you mean?)

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,137 @@
...
+            .ok_or_else(|| Error::credentials_not_provided("token"))?;
+        Ok(Credentials::builder()
+            .with_token(token)
+            .with_address(req.remote())
+            .build())
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct MultiExtractor {
```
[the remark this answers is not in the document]
pyakushin: You have a point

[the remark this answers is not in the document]
pyakushin: But it will look like 
```rust
pub struct MultiExtractor<R>(Box<dyn Extractor<R>>);
```
And it's pain to make Extractor object-safe

bob-apps/bin/bobd.rs
pyakushin: Logger is not initialized yet

bob-access/src/authenticator/basic.rs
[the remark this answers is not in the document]
idruzhitskiy: I mean we have `CredentialsKind::Basic` and `CredentialsKind::Token`. Instead of using `username()` for all credentials, wouldn't it be better to work with `CredentialsKind` everywhere we need either username or password? In this case we can pattern match that `cred` is `Basic`.
This way if we add more `CredentialsKind` variants in the future, all places which relies on particular credential types will be visible with compile errors

bob-access/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+        debug!("request received");
+        let credentials = self.extractor.extract(&req).unwrap();
+        debug!("credentials: {:#?}", credentials);
+        match self.authenticator.check_credentials(credentials) {
+            Ok(permissions) => {
+                let required_permissions = req.get_permissions();
+                if permissions.contains(required_permissions) {
+                    Box::pin(self.service.call(req).map(|r| Ok(r.unwrap())))
+                } else {
+                    todo!("not enough permissions")
```
ikopylov, line 119: A special error should be created for this case

bob-access/src/permissions.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+pub trait GetPermissions {
+    fn get_permissions(&self) -> Permissions;
+}
+
+impl<T> GetPermissions for Request<T> {
+    fn get_permissions(&self) -> Permissions {
+        warn!("@TODO differentiate GRPC and REST requests");
+        match *self.method() {
```
ikopylov, line 42: There should be a way to specify the required permissions for each method

ikopylov: Here is an example how to do it: https://github.com/tokio-rs/axum/blob/main/examples/jwt/src/main.rs#L76
All methods should have special additional parameter of type, that implements `FromRequest<>`. That type by itself can have a required permissions as a type parameters.

I think it should look something like this:
```
fn test_method(..., authData: AuthData<Permissions::READ_REST>) -> ...
```

bob-access/src/extractor.rs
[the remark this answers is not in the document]
idruzhitskiy: Yeah, we'll have to use boxed impls... I think this can be skipped for now

bob-access/src/permissions.rs
```diff
@@ -32,16 +34,43 @@ impl From<&User> for Permissions {
...
-        warn!("@TODO differentiate GRPC and REST requests");
+        trace!("request: {:#?}", self);
+        if let Some(content_type) = self.headers().get(http::header::CONTENT_TYPE) {
+            let grpc_content_type = HeaderValue::from_static("application/grpc");
+            if content_type == grpc_content_type {
+                let perms = match *self.method() {
+                    Method::GET => Permissions::READ,
+                    Method::PUT | Method::DELETE | Method::POST => Permissions::WRITE,
+                    _ => Permissions::FORBIDDEN,
+                };
```
ikopylov, line 67: Have you checked, that this conversion works for the gRPC API?

bob-access/src/permissions.rs
```diff
@@ -32,16 +34,43 @@ impl From<&User> for Permissions {
...
+                return perms;
+            }
+        }
         match *self.method() {
-            Method::GET => Permissions::READ & Permissions::READ_REST,
-            Method::PUT => Permissions::WRITE & Permissions::WRITE_REST,
+            Method::GET => Permissions::READ_REST,
+            Method::PUT | Method::DELETE | Method::POST => Permissions::WRITE_REST,
```
ikopylov, line 73: As I wrote before, every method in REST API should be explicitly marked with required permissions. Some methods should work without any permission at all (e.g. `status`)

bob-access/src/error.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn description(&self) -> (StatusCode, &'static str) {
+        use Error::*;
+        match self {
+            _Unknown => (StatusCode::INTERNAL_SERVER_ERROR, "Unknown error"),
+            InvalidToken(_) => (StatusCode::BAD_REQUEST, "Invalid token"),
+            Validation(_) => (StatusCode::BAD_REQUEST, "Validation error"),
+            Os(_) => (StatusCode::INTERNAL_SERVER_ERROR, "Os error"),
+            UserNotFound => (StatusCode::BAD_REQUEST, "User not found"),
+            ConversionError(_) => (StatusCode::BAD_REQUEST, "Conversion error"),
+            CredentialsNotProvided(_) => (StatusCode::BAD_REQUEST, "Credentials not provided"),
```
ikopylov, line 35: This should be a `FORBIDDEN` status code

bob-access/src/error.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Error {
+    pub fn description(&self) -> (StatusCode, &'static str) {
+        use Error::*;
+        match self {
+            _Unknown => (StatusCode::INTERNAL_SERVER_ERROR, "Unknown error"),
+            InvalidToken(_) => (StatusCode::BAD_REQUEST, "Invalid token"),
+            Validation(_) => (StatusCode::BAD_REQUEST, "Validation error"),
+            Os(_) => (StatusCode::INTERNAL_SERVER_ERROR, "Os error"),
+            UserNotFound => (StatusCode::BAD_REQUEST, "User not found"),
```
ikopylov, line 33: This should also be a `FORBIDDEN` status code

bob-access/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    S: NamedService,
+{
+    const NAME: &'static str = S::NAME;
+}
+
+pub async fn handle_auth_error(err: BoxError) -> (StatusCode, String) {
+    error!("{}", err);
+    if let Ok(err) = err.downcast::<Error>() {
+        match err.as_ref() {
+            Error::InvalidToken(_) => (StatusCode::FORBIDDEN, "Invalid token.".into()),
```
ikopylov, line 140: Why there are a two places where `Error` is converted to `StatusCode` (another in [bob-access/src/error.rs](https://github.com/qoollo/bob/pull/217/files#diff-27f4a7db5d5eb59894fa6bb070daf5134db6529825c3f0edf9cd90e78a6cd6a6))?

bob-access/src/credentials.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[async_trait]
+impl<B> FromRequest<B> for Credentials
+where
+    B: Send,
+{
+    type Rejection = Error;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        req.extract()
```
ikopylov, line 103: If it is possible to parametrize `Credentials` with `Authenticator` and with `Perms` as generic parameters, then all the checks can be done right here

bob/src/api/mod.rs
```diff
@@ -1,2 +1,1075 @@
...
+
+async fn metrics<A>(
+    Extension(bob): Extension<&BobServer>,
+    Extension(auth): Extension<A>,
+    creds: Credentials,
+) -> Result<Json<MetricsSnapshotModel>, AuthError>
+where
+    A: Authenticator,
+{
+    if !auth.check_credentials(creds)?.has_rest_read() {
```
ikopylov, line 302: I think, that metrics should be available without authorization

bob/src/api/s3.rs
```diff
@@ -0,0 +1,239 @@
...
+        };
+        Response::builder()
+            .status(status)
+            .body(boxed(Empty::new()))
+            .expect("failed to set empty body for response")
+    }
+}
+
+pub(crate) fn routes() -> Vec<(&'static str, MethodRouter)> {
+    vec![
```
ikopylov, line 51: S3 API should also be protected

bob-access/src/error.rs
pyakushin: @ikopylov You previously suggested to use UNAUTHORIZED

bob/src/api/mod.rs
pyakushin: Removed

bob/src/api/s3.rs
pyakushin: Done

bob-access/src/authenticator/basic.rs
pyakushin: Should work now

bob-access/src/error.rs
ikopylov: Sorry. 
According to [this](https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses) the best status code is `UNAUTHORIZED` (which actually means `not authenticated`). `FORBIDDEN` is used when the user does not have permission to call the method.

bob-access/src/error.rs
ikopylov: Sorry one more time. Here the best status code is `UNAUTHORIZED`

ikopylov:
Now that you're using the approach with passing a parameter of type that implements `FromRequest`, the question is: is it possible to do the same with Rocket? 
I found this article: https://mmhaskell.com/blog/2020/8/17/authentication-in-rocket
It seems that authorization can be done in Rocket in the same way, so migration to Axum is not required.
Thus, maybe, it is better to keep Rocket as the REST API framework. What do you think?

bob/src/api/mod.rs
```diff
@@ -1,2 +1,1069 @@
...
+    let vdisks = collect_disks_info(bob);
+    let node = Node {
+        name,
+        address,
+        vdisks,
+    };
+    Json(node)
+}
+
+async fn metrics(Extension(bob): Extension<&BobServer>) -> Json<MetricsSnapshotModel> {
```
ikopylov, line 298: @Boneyan reported, that this method returns an error: `Missing request extension: Extension of type '&bob::server::Server' was not found. Perhaps you forget to add it?`

bob-access/src/permissions.rs
ikopylov: @Boneyan reported, that this approach does not work

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,117 @@
...
+            .extensions()
+            .and_then(|ext| ext.get::<TcpConnectInfo>()?.remote_addr());
+        let basic_credentials = self.extract_basic(header_map, addr)?;
+        let token_credentials = self.extract_token(header_map, addr)?;
+        match (basic_credentials, token_credentials) {
+            (Some(_), Some(_)) => Err(Error::MultipleCredentialsTypes),
+            (Some(basic_credentials), None) => Ok(basic_credentials),
+            (None, Some(token_credentials)) => Ok(token_credentials),
+            _ => Ok(Credentials::builder()
+                // .with_address(req.remote_addr())
```
idruzhitskiy, line 48: Why not leave `with_address` here and remove from specific extraction functions?

idruzhitskiy: `Builder` can be passed into `extract_*` functions, so this one will contain common setup for builder and specific functions will add either username+password or token

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,117 @@
...
+                .build()),
+        }
+    }
+
+    fn extract_basic(
+        &self,
+        header_map: &HeaderMap,
+        addr: Option<SocketAddr>,
+    ) -> Result<Option<Credentials>, Error> {
+        let username = if let Some(username) = header_map.get("username") {
```
idruzhitskiy, line 58: Getting `header_map` content with parsing seems to be common, so it should be extracted into a separate function, and extract functions will be left to `if let Some(...)` blocks and altering builder

bob-apps/bin/bobd.rs
```diff
@@ -81,33 +86,101 @@ async fn main() {
...
-        .tcp_nodelay(true)
-        .add_service(new_service)
-        .serve(addr)
-        .await
-        .unwrap();
+    let grinder = Grinder::new(mapper.clone(), &node).await;
+    let authentication_type = matches.value_of("authentication_type").unwrap();
+    match authentication_type {
+        "stub" => {
+            let users_storage =
```
idruzhitskiy, line 105: Match branches should be extracted into separate functions

idruzhitskiy (commented):
Also I still think that matching on `CredentialsType` is better than checking if `username` or `token` is `None`

ikopylov:
@Boneyan please, merge master into this request

bob-access/src/authenticator/basic.rs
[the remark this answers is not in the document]
ikopylov: @Boneyan

bob-access/src/error.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    pub fn status_code(&self) -> StatusCode {
+        use Error::*;
+        match self {
+            _Unknown => StatusCode::INTERNAL_SERVER_ERROR,
+            InvalidToken(_) => StatusCode::BAD_REQUEST,
+            Validation(_) => StatusCode::BAD_REQUEST,
+            Os(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            UserNotFound => StatusCode::FORBIDDEN,
```
ikopylov, line 48: This should be mapped to `StatusCode::UNAUTHORIZED`

bob-access/src/authenticator/users_storage/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+            read_rest,
+            write_rest,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct User {
+    _username: String,
+    password: String,
```
ikopylov, line 34: Future task: https://github.com/qoollo/bob/issues/514

bob-access/src/extractor.rs
ikopylov: This branch is for `(None, None)` case. As I understand, it is triggered with inter-node communication and also when no authorization is enabled. For the first case `remote_addr` is required. For the second case we should not try to extract any credentials at all

bob-access/src/permissions.rs
```diff
@@ -0,0 +1,90 @@
...
+            self.contains(Self::WRITE_REST)
+        )
+    }
+}
+
+pub trait GetGrpcPermissions {
+    fn get_grpc_permissions(&self) -> Option<Permissions>;
+}
+
+impl<T> GetGrpcPermissions for Request<T>
```
ikopylov, line 71: It doesn't seem to be used anywhere

ikopylov:
@Boneyan when you will be checking that all REST API methods are migrated, please also compare methods to those declared in https://github.com/qoollo/bob/blob/master/config-examples/openapi.yaml. If something is inconsistent in `openapi.yaml` then update the file

bob/src/api/mod.rs
```diff
@@ -1,2 +1,1066 @@
...
+
+
+async fn metrics<A: Authenticator>(
+    bob: Extension<BobServer<A>>,
+) -> Json<MetricsSnapshotModel> {
+    let snapshot = bob.metrics().read().await.clone();
+    Json(snapshot.into())
+}
+
+async fn version() -> Json<VersionInfo> {
```
ikopylov, line 309: Please add a comment to all REST API methods with the URL they are bound to

idruzhitskiy:
@Boneyan I've updated dockerfiles to include bob-access project, but there are build errors

ikopylov:
@idruzhitskiy thanks! It seems that `COPY` command did not replace `lib.rs`, because I see this in the output: `fn main() {println!("if you see this, the build broke")}`. Did it build successfully locally?

CHANGELOG.md
```diff
@@ -4,9 +4,13 @@ Bob versions changelog
 
 ## [Unreleased]
 #### Added
+- Access management
+  - Nodes authentication. (#318)
+  - Http API authentication.
 
 
 #### Changed
+- Rocket replaced with axum.
```
ikopylov, line 13: Please, add link to the current request for all new changelog lines

bob-access/src/permissions.rs
```diff
@@ -0,0 +1,69 @@
...
+            "GRPC: [read: {}, write: {}], REST: [read: {}, write: {}]",
+            self.contains(Self::READ),
+            self.contains(Self::WRITE),
+            self.contains(Self::READ_REST),
+            self.contains(Self::WRITE_REST)
+        )
+    }
+}
+
+pub trait GetGrpcPermissions {
```
ikopylov, line 66: It seems that this trait is also not used anywhere

bob/src/api/s3.rs
```diff
@@ -0,0 +1,264 @@
...
+        Response::builder()
+            .status(StatusCode::OK)
+            .header("Content-Type", self.content_type)
+            .header("Last-Modified", self.data.meta().timestamp().to_string())
+            .body(boxed(Full::new(self.data.into_inner().into())))
+            .expect("failed to set body")
+    }
+}
+
+async fn get_object<A>(
```
ikopylov, line 129: Please, add comments with URL to all REST API methods in this file too

bob-access/src/credentials.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+impl<T> From<&Request<T>> for Credentials {
+    fn from(req: &Request<T>) -> Self {
+        let md = req.metadata();
```
ikopylov, line 145: This function is incomplete. It does not have `token` extraction. Obviously, it should somehow use `extract` function from `ExtractorExt` implementation

bob-access/src/credentials.rs
```diff
@@ -0,0 +1,159 @@
...
+{
+    type Rejection = Error;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        req.extract()
+    }
+}
+
+impl<T> From<&Request<T>> for Credentials {
+    fn from(req: &Request<T>) -> Self {
```
ikopylov, line 144: This function is incomplete. It does not have `token` extraction. Obviously, it should somehow use `extract` function from `ExtractorExt` implementation

bob-access/src/credentials.rs
```diff
@@ -0,0 +1,159 @@
...
+        *self == CredentialsType::Basic
+    }
+
+    pub fn is_stub(&self) -> bool {
+        *self == CredentialsType::Stub
+    }
+}
+
+fn _credentials_type(tp: Option<CredentialsType>) -> CredentialsType {
+    static mut CREDENTIALS_TYPE: CredentialsType = CredentialsType::Stub;
```
ikopylov, line 41: This is not a very good solution. I suggest to try to use generic parametrization to pass the enum with data required for auth (None, UserPassword, Token). This can be achieved by declaring `CredentialsHolder<A>` struct, that directly wraps `Credentials` (the only difference is the generic parameter `A` which designate the authenticator). On `Authenticator` trait we should add the const enum field, that indicates the data that needs to be extracted from request for this particular authenticator. Additionally `FromRequest<B>` should be implemented not for `Credentials` but for `CredentialsHolder<A>` struct, and `CredentialsHolder<A>` struct should be used as a parameter in all REST API methods. In this way extractor can get information about fields from `Authenticator` type and pass them to `ExtractorExt::extract`

ikopylov:
@Boneyan `users.yaml` should be added to docker images (it should be placed into `dockerfiles/default-configs` directory)

bob-apps/bin/bobd.rs
```diff
@@ -82,38 +87,97 @@ async fn main() {
...
+        .unwrap_or_else(|| node.http_api_port());
     let http_api_address = matches
         .value_of("http_api_address")
         .and_then(|v| v.parse().ok())
-        .unwrap_or(node.http_api_address());
-    bob.run_api_server(http_api_address, http_api_port);
+        .unwrap_or_else(|| node.http_api_address());
 
-    create_signal_handlers(&bob).unwrap();
+    let authentication_type = matches.value_of("authentication_type").unwrap();
```
ikopylov, line 100: As I see, the authentication type read from command line parameters. This is not the best way to configure authentication. We need to configure it from `node.yaml`

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,119 @@
...
+                return Ok(self.extract_basic(header_map, builder)?);
+            },
+            CredentialsType::Token => {
+                let (header_map, builder) = prepare_builder(self)?;
+                return Ok(self.extract_token(header_map, builder)?);
+            },
+        }
+    }
+
+    fn extract_basic(
```
ikopylov, line 59: I just noticed that `extract_basic` and `extract_token` are part of `ExtractorExt` trait and are actually available from outside. In this situation, these functions should be self-contained: they should not take `header_map` and `builder` as parameters, but perform all data extraction inside their body. Basically, that means that `prepare_builder` should be called within these functions body

dockerfiles/default-configs/users.yaml
```diff
@@ -0,0 +1,24 @@
+---
+roles:
```
ikopylov, line 2: I think, the default user config is not the best one. I suggest to add `read` access to `writer`. 
`users` should be updated to have common user names: `admin`, `readonly`, `user`. 
For default config it is better to unify all passwords and set them to `password`.

Please, update `users.yaml` in the same way in `config_examples` too

bob/src/api/mod.rs
```diff
@@ -1,2 +1,1106 @@
...
+// GET /data/:key
+async fn get_data<A>(
+    bob: Extension<BobServer<A>>,
+    AxumPath(key): AxumPath<String>,
+    creds: CredentialsHolder<A>,
+) -> Result<impl IntoResponse, StatusExt>
+where
+    A: Authenticator,
+{
+    if !bob.auth().check_credentials(creds.into_credentials())?.has_rest_read() {
```
ikopylov, line 908: For data access in REST API we should use `has_read`, `has_write`. Please, update `get_data`, `put_data`, `delete_data` here and in S3 API

bob-access/src/authenticator/basic.rs
```diff
@@ -40,7 +40,14 @@ impl<Storage: UsersStorage> Basic<Storage> {
         }
         self.nodes
             .get(other.ip().as_ref()?)
-            .map(|cred| cred.username() == other.username())
+            .map(|cred| 
+                match (cred.kind(), other.kind()) {
+                    (Some(CredentialsKind::Basic { username: c_username, password: _ }),
+                    Some(CredentialsKind::Basic { username: o_username, password: _ })) => c_username == o_username,
+                    (Some(CredentialsKind::Token(c_token)), 
+                    Some(CredentialsKind::Token(o_token))) => c_token == o_token,
```
ikopylov, line 48: This is incorrect. Tokens can't be compared directly. Moreover, this is a function to perform inter-node authentication: it does not use either Token or Basic authentication. Nodes use a separate authentication scheme by IP and node name. Previously it worked, because the node name was passed as username, but this was a bad approach. Now things have gotten even worse. I suggest to add a separate `CredentialsKind` (`InterNode`), that stores `node_name` inside and used here for authentication. With this approach node name should be passed in separate header named `node_name`

bob-access/src/credentials.rs
```diff
@@ -1,19 +1,31 @@
...
     address: Option<SocketAddr>,
     kind: Option<CredentialsKind>,
 }
 
+pub struct CredentialsHolder<A: Authenticator> {
+    credentials: Credentials,
+    pd: PhantomData<A>,
+}
+
+impl<A: Authenticator> CredentialsHolder<A> {
```
ikopylov, line 23: It is better to implement standard `From<>` trait for `Credentials`

bob/src/server.rs
```diff
@@ -104,8 +104,8 @@ where
     A: Authenticator,
 {
     async fn put(&self, req: Request<PutRequest>) -> ApiResult<OpStatus> {
-        let creds = Credentials::from(&req);
-        if !self.auth.check_credentials(creds)?.has_write() {
+        let creds: CredentialsHolder<A> = CredentialsHolder::from(&req);
```
ikopylov, line 107: `req.into()` will be simpler

bob-access/src/authenticator/basic.rs
ikopylov: Additionally, it would be good to perform inter-node authentication only for gRPC API and ignore it for REST API. This can be achieved by creating separate `check_credentials` functions for gRPC and for REST API

bob/src/api/s3.rs
```diff
@@ -0,0 +1,266 @@
...
+            source_key,
+        })
+    }
+}
+
+async fn copy_object<A: Authenticator>(
+    bob: &BobServer<A>,
+    key: BobKey,
+    headers: CopyObjectHeaders,
+) -> Result<StatusS3, StatusS3> {
```
ikopylov, line 243: Authorization is missing

Boneyan: This function is called in PUT after authorization is done

bob-access/src/authenticator/basic.rs
```diff
@@ -0,0 +1,127 @@
...
+                );
+        
+                let user = self.users_storage.get_user(&username)?;
+                if user.password() == password {
+                    Ok(user.into())
+                } else {
+                    Err(Error::UnauthorizedRequest)
+                }
+            },
+            Some(CredentialsKind::Token(_token)) => {
```
ikopylov, line 111: Same here

bob-access/src/authenticator/basic.rs
```diff
@@ -0,0 +1,127 @@
...
+                    Err(Error::UnauthorizedRequest)
+                }
+            },
+            None => {
+                Err(Error::CredentialsNotProvided("missing credentials".to_string()))
+            }
+        }
+    }
+
+    fn check_credentials_rest(&self, credentials: Credentials) -> Result<Permissions, Error> {
```
ikopylov, line 94: `check_credentials_grpc` and `check_credentials_rest` have common parts and I suggest to extract that common parts into separate `check_credentials_common` function and call it from these two

bob-access/src/credentials.rs
```diff
@@ -20,23 +20,25 @@ pub struct CredentialsHolder<A: Authenticator> {
...
     Token(String),
+    InterNode(String),
 }
 
 #[derive(Clone, Copy, PartialEq, Debug, Serialize, Deserialize)]
 pub enum CredentialsType {
     None,
     Basic,
     Token,
+    InterNode,
```
ikopylov, line 41: `InterNode` is not needed in this enum. This is basically an `AuthenticationType`, so I suggest to rename it and move closer to `Authenticator`

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,131 @@
...
+pub trait Extractor {
+    fn get_header(&self, header: &str) -> Result<Option<&str>, Error>;
+    fn get_extension<T: Send + Sync + 'static>(&self) -> Option<&T>;
+}
+
+pub trait ExtractorExt {
+    fn extract(&self, cred_type: CredentialsType) -> Result<Credentials, Error>;
+    fn extract_basic(&self) -> Result<Credentials, Error>;
+    fn extract_token(&self) -> Result<Credentials, Error>;
+    fn extract_internode(&self) -> Result<Credentials, Error>;
```
ikopylov, line 16: This function is not needed as a part of trait

bob-access/src/extractor.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl<T: Extractor> ExtractorExt for T {
+    fn extract(&self, cred_type: CredentialsType) -> Result<Credentials, Error> {
+        match cred_type {
+            CredentialsType::None => {
+                return Ok(Credentials::default());
+            },
+            CredentialsType::Basic => {
+                return self.extract_basic().or_else(|e| {
+                    self.extract_internode().map_err(|_| e)
```
ikopylov, line 36: InterNode credentials is required for `Basic` authenticator, so its extraction can be integrated into `extract_basic`. Same for `extract_token`

bob-common/src/bob_client.rs
```diff
@@ -126,6 +127,7 @@ pub mod b_client {
         pub async fn ping(&self) -> PingResult {
             let mut client = self.client.clone();
             let mut req = Request::new(Null {});
+            self.set_credentials(&mut req);
```
ikopylov, line 130: `set_credentials` should be called in all gRPC API functions

bob-access/src/extractor.rs
Boneyan: To use this function inside extract_basic() and extract_token() it needs to be part of the trait (or should I create another trait, e.g. ExtractorInternode)

ikopylov: Ok, then just integrate code from `extract_internode` directly into `extract_basic` and `extract_token`. Tiny code duplication in this situation is not a big deal

bob-access/src/error.rs
```diff
@@ -0,0 +1,98 @@
...
+
+impl Error {
+    pub fn msg(&self) -> &'static str {
+        use Error::*;
+        match self {
+            _Unknown => "Unknown error",
+            InvalidToken(_) => "Invalid token",
+            Validation(_) => "Validation error",
+            Os(_) => "Os error",
+            UserNotFound => "User not found",
```
ikopylov, line 32: It is not safe to return for client this message, because it can be used to find out which users are registered and which are not. It is better to return `Unauthorized request` message here (same as for `UnauthorizedRequest`)

bob-access/src/credentials.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[derive(Debug, Clone)]
+pub enum CredentialsKind {
+    Basic { username: String, password: String },
+    Token(String),
+    InterNode(String),
+}
+
+impl CredentialsKind {
+    pub fn is_internode(&self) -> bool {
```
ikopylov, line 37: Same functions can be added for all `CredentialsKind` values (additionally for `Basic` and for `Token`)

ikopylov (approved):
LGTM

PR #221, opened 2021-06-14
Delete record api

### LINKED ISSUE #209: Make it possible to delete individual records
This feature is not consistent with Bob architecture and it cannot be done 100% safely and performant, but it is still required for some use-cases. Practical example: several invalid records were inserted to Bob and the client application is unable to process these records. In such case it would be great to be able to delete these records, even if that procedure would be slow and not 100% safe.

Possible way to implement deletion is to write special record at the end of the BLOB with the same key as original record and with special _deletion_ flag. After that the index file for that BLOB should be regenerated.

vovac12:
Closes #209
Draft until [#103](https://github.com/qoollo/pearl/pull/103) not merged


bob-backend/src/core.rs
```diff
@@ -356,4 +359,45 @@ impl Backend {
...
+    pub async fn delete(&self, key: BobKey, with_aliens: bool) -> Result<u64, Error> {
+        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
+        let mut total_count = 0;
+        if let Some(path) = disk_path {
+            trace!("DELETE[{}] try delete normal", key);
+            total_count += self
+                .delete_single(key, Operation::new_local(vdisk_id, path.clone()))
+                .await
+                .map_err(|e| {
+                    debug!("DELETE[{}] delete alien error: {}", key, e);
```
idruzhitskiy, line 372: Copypaste :) 
BTW I think you should create collection of operations (normal + aliens) and single execution function, so you can map-reduce it and return total count

bob/src/api/http.rs
```diff
@@ -627,6 +628,13 @@ fn internal(message: String) -> StatusExt {
     StatusExt::new(Status::InternalServerError, false, message)
 }
 
+#[delete("/data/<key>")]
+fn delete_records_by_key(bob: State<BobServer>, key: BobKey) -> Result<StatusExt, StatusExt> {
```
idruzhitskiy, line 632: Don't forget to add this to openapi

vovac12: Added

bob-backend/src/core.rs
vovac12: Fixed

pyakushin:
LGTM

bob-backend/src/pearl/group.rs
```diff
@@ -526,6 +526,40 @@ impl Group {
...
+        Ok(total_count)
+    }
+
+    async fn delete_common(holder: Holder, key: BobKey) -> Result<u64, Error> {
+        let result = holder.delete(key).await;
+        if let Err(e) = &result {
+            if !e.is_key_not_found() && !e.is_not_ready() {
+                holder.try_reinit().await?;
+                holder.prepare_storage().await?;
+                debug!("backend pearl group get common storage prepared");
```
idruzhitskiy, line 557: Typo, `delete common`

bob/src/grinder.rs
```diff
@@ -199,6 +199,26 @@ impl Grinder {
         self.counter.spawn_task(self.backend.clone());
         self.hw_counter.spawn_task();
     }
+
+    pub(crate) async fn delete(&self, key: BobKey, with_aliens: bool) -> Result<u64, Error> {
+        trace!(">>>- - - - - GRINDER DELETE START - - - - -");
```
ikopylov, line 204: It would be good to add `GRINDER_DELETE_COUNTER` to track number of deleted records. Also it would be good to add `GRINDER_DELETE_TIMER` to track delete time in Graphite

ikopylov (approved):
LGTM

PR #222, opened 2021-06-15
#211 Credentials extraction

### LINKED ISSUE #211: Authorization: implement extraction of user credentials from the request
Extract token or user/password from gRPC and REST requests and reject/accept that request based on on response from credential validator

vovac12:
Closes #211


bob-access/src/credentials.rs
```diff
@@ -1,2 +1,41 @@
...
+        password: impl Into<String>,
+    ) -> Self {
+        self.kind = Some(CredentialsKind::Basic {
+            username: username.into(),
+            password: password.into(),
+        });
+        self
+    }
+
+    pub fn with_token(mut self, token: impl Into<String>) -> Self {
```
idruzhitskiy, line 32: IMO there should be either separate builder for the credentials, or multiple constructors

bob-access/src/extractor.rs
```diff
@@ -14,7 +16,112 @@ impl StubExtractor {
...
+pub struct BasicExtractor {}
+
+impl<T> Extractor<TRequest<T>> for BasicExtractor {
+    fn extract(&self, req: &TRequest<T>) -> Result<Credentials, Error> {
+        let meta = req.metadata();
+        let username = meta
+            .get("username")
+            .ok_or_else(|| Error::credentials_not_provided("username"))?
+            .to_str()
+            .map_err(|e| Error::unknown(e.to_string()))?;
```
idruzhitskiy, line 34: Why unknown? Probably should be "parse error" or "conversion error"

bob-access/src/extractor.rs
```diff
@@ -14,7 +16,112 @@ impl StubExtractor {
...
+pub struct MultiExtractor {
+    basic_extractor: BasicExtractor,
+    token_extractor: TokenExtractor,
+}
+
+impl<T> Extractor<TRequest<T>> for MultiExtractor {
+    fn extract(&self, req: &TRequest<T>) -> Result<Credentials, Error> {
+        let basic_credentials = self.basic_extractor.extract(req);
+        let token_credentials = self.token_extractor.extract(req);
+        if basic_credentials.is_ok() && token_credentials.is_ok() {
```
idruzhitskiy, line 101: IMO pattern matching with `match (basic_credentials.is_ok(), token_credentials.is_ok()) { (true, true) => ... }` is more readable

bob-access/src/credentials.rs
vovac12: I added a `CredentialsBuilder`

bob-access/src/extractor.rs
vovac12: Replaced with `ConversionError`

bob-access/src/extractor.rs
vovac12: Done

pyakushin: @vovac12 You added this pattern only for `MultiExtractor`

pyakushin (approved):
LGTM, check 1 comment

vovac12:
@pyakushin Fixed

idruzhitskiy:
The "famous" match part in extractor impls can be factored out in smth like "find_single_credentials", but that's not necessary. Other than that - LGTM.

PR #224, opened 2021-06-15
Holders are checked in reversed order

Justarone:
This is needed to retrieve the last record with given key (this gives a capability to perform `update` operation properly with writing new record with given key)

P. S.: On Pearl side in [PR1](https://github.com/qoollo/pearl/pull/105) and [PR2](https://github.com/qoollo/pearl/pull/102) blobs are checked in reversed order and last-added record from blob is returned (so in this PRs capability added and holders should also be checked in reversed order to support this `update` operation completely)


pyakushin (approved):
LGTM

Issue #213, opened 2021-06-02
Authorization: implement storage for credentials, roles and users 


ikopylov:
Initial implementation: store these data in file (users.yaml)

ikopylov:
Closed via https://github.com/qoollo/bob/pull/217

PR #232, opened 2021-06-23
Add users storage to authenticator

Justarone:
Part of #213


bob-access/src/authenticator/users_storage.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn can_read_rest(&self) -> bool {
+        self.perms.read_rest
+    }
+
+    pub fn can_write_rest(&self) -> bool {
+        self.perms.write_rest
+    }
+}
+
+pub trait UsersStorageTrait: Default + Clone {
```
idruzhitskiy, line 55: IMO "Trait" is redundant in the name

pyakushin (changes requested):

config-examples/users.yaml
```diff
@@ -0,0 +1,24 @@
...
+  reader:
+    read: true
+    read_rest: true
+  writer:
+    write: true
+    write_rest: true
+
+users:
+  - username: admin_user
+    password: admin_pass
```
agend, line 17: Will password be stored as plain text?

bob-access/src/authenticator/users_storage/hashmap.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl UsersStorageTrait for UsersMap {
+    fn get_user<'a>(&'a self, username: &str) -> Result<&'a User, Error> {
+        self.inner.get(username).ok_or_else(|| Error::not_found())
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct ConfigUser {
```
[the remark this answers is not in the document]
Justarone: User and ConfigUser are slightly different. 
ConfigUser - structure, which represent user from config (aka login, password + **role** (that's important)). 
We have links: User <-> Role <-> Perms, and it's not a good idea to resolve User's perms in runtime, so I do this after I've parsed the config. So `User` = login + password + perms (resolved). So these are two different entities IMO

bob-access/src/authenticator/users_storage/hashmap.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut users = HashMap::new();
+    yaml_users.into_iter().try_for_each(|u| {
+        let perms = roles
+            .get(&u.role)
+            .ok_or_else(|| Error::validation(format!("Can't find role {}", u.role)))?
+            .clone();
+        let user = User::new(u.username.clone(), u.password, perms);
+        users.insert(u.username, user).map_or(Ok(()), |user| {
+            Err(Error::validation(format!(
+                "Users with the same username (first: {:?})",
```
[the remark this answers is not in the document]
Justarone: Hm..
`map_or` for Option will return:
1. default == first argument == `Ok(())` for None value (case when user added for the first time)
2. Some(user) --(map_func)--> Err(validation) for Some(user) value (case, when user added for 2+ time)

config-examples/users.yaml
Justarone: Hm, I've implemented simplest storage and it do just 2 things:
1. parses config
2. allows to get user by username with his password [as it was written in config] and perms

I think, we should store hashed password (in config) and set specific hash function, which will resolve correct passwords to that hashes, but I also think, that it may be weird to implement it on this storage level.
This config is just an example of what can be parsed with this storage.

PR #248, opened 2021-07-02
247 extend version

idruzhitskiy:


idruzhitskiy:
Requires https://github.com/qoollo/pearl/pull/107

bob-apps/bin/bobd.rs
```diff
@@ -102,7 +105,16 @@ fn spawn_signal_handler(
 
 fn get_matches<'a>() -> ArgMatches<'a> {
     App::new(env!("CARGO_PKG_NAME"))
-        .version(env!("CARGO_PKG_VERSION"))
+        .version(
```
pyakushin, line 108: I think it's better to use other arg/subcommand instead of version

bob/src/api/http.rs
```diff
@@ -87,9 +87,42 @@ pub(crate) struct DiskState {
...
+    build_time: String,
+}
+
+#[derive(Debug, Serialize)]
+pub(crate) struct VersionInfo {
+    bob_version: Version,
+    pearl_version: Version,
+}
+
+pub fn get_bob_version() -> String {
```
pyakushin, line 102: I suggest to move this helpers to separate mod, because you use them outside of `bob` crate

bob-apps/bin/bobd.rs
idruzhitskiy: Why? It's supposed to be the main version of the app...

bob/src/api/http.rs
idruzhitskiy: Done

bob-apps/bin/bobd.rs
ikopylov: The idea was to print all this info by executing `version` command

pyakushin (changes requested):
LGTM. Create PR for `Pearl` 106 issue and fix dep in Cargo.toml

pyakushin (approved):

PR #251, opened 2021-07-04
Prometheus exporter for metrics

### LINKED ISSUE #240: Publish metrics to Prometheus
Add a Prometheus as additional metrics target

Justarone:
Closes #240


config-examples/prometheus.yml
```diff
@@ -0,0 +1,10 @@
+global:
+  scrape_interval: 2s
+
+scrape_configs:
+  - job_name: 'bob'
+
+    scrape_interval: 2s
+
+    static_configs:
+      - targets: ['127.0.0.1:9000']
```
ikopylov, line 10: Can this config be integrated into `node.yaml`?

Probably, config restructuring is required:
```
metrics:
    graphite:
        ...
    prometheus:
        ...
```

bob-common/src/metrics/exporters/global_exporter/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+use rate_processor::RateProcessor;
+
+pub(crate) struct GlobalRecorder {
+    recorders: Vec<Box<dyn Recorder>>,
+    rate_processor: RateProcessor,
+}
+
+impl GlobalRecorder {
+    pub(crate) fn new(recorders: Vec<Box<dyn Recorder>>) -> Self {
+        assert!(recorders.len() != 0, "{}", NO_RECORDERS_MSG);
```
pyakushin, line 18: I think it's better to use `!vec.is_empty()` and `panic` instead of assertion

bob-common/src/metrics/exporters/global_exporter/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self {
+            recorders,
+            rate_processor: RateProcessor::run_task(),
+        }
+    }
+}
+
+// NOTE: first recorder is processed separately to avoid redundant clone (value is moved)
+impl Recorder for GlobalRecorder {
+    fn register_gauge(
```
pyakushin, line 28: You have 3 almost identical fns, maybe it better to do smth like this?
```
fn  register_with(&self, f: fn(...args), ...args) {
    for rec in self.recorders.iter() {
        f(&rec, ...args);
    }
} 
```

bob-common/src/metrics/exporters/global_exporter/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fn increment_counter(&self, key: &Key, value: u64) {
+        self.rate_processor.process(key, value);
+        for rec in self.recorders.iter() {
+            rec.increment_counter(key, value);
+        }
+    }
+
+    #[allow(clippy::cast_sign_loss)]
```
pyakushin, line 68: Looks like this lint is redundant

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::time::Duration;
+use std::{collections::HashMap, time::Instant};
+use tokio::sync::mpsc::{channel, Receiver, Sender};
+
+const RATE_INTERVAL: Duration = Duration::from_secs(5);
+
+type TimeStamp = Instant;
+
+pub(super) struct RateProcessor {
+    tx: Sender<(String, u64, TimeStamp)>,
```
pyakushin, line 12: I think the code would be more readable if you add struct for channel message

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
+const BUFFER_SIZE: usize = 10_485_760; // 10 Mb
```
pyakushin, line 1: Just note, it is not 10Mb, `memory_usage = BUFFER_SIZE * size_of(Message)`

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub(super) fn run_task() -> Self {
+        let (tx, rx) = channel(BUFFER_SIZE);
+        tokio::spawn(rate_processor_task(rx));
+        Self { tx }
+    }
+
+    pub(super) fn process(&self, key: &Key, value: u64) {
+        if let Err(e) = self
+            .tx
+            .try_send((key.name().to_owned(), value, Instant::now()))
```
pyakushin, line 25: Why you don't use async methods for channel here?

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            error!("Can't send counter to task with rates (reason: {})", e);
```
pyakushin, line 27: You can process TrySendError:
Full - try again or just drop
Closed - shutdown whole processor, receiver closed and won't be restored

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
...
+
+async fn rate_processor_task(mut rx: Receiver<(String, u64, TimeStamp)>) {
+    let mut counters_map = HashMap::new();
+    let mut base_ts = Instant::now();
+    if let Some((s, v, t)) = rx.recv().await {
+        process_counter(&mut counters_map, s, v);
+        base_ts = t;
+    }
+
+    loop {
```
pyakushin, line 40: You can run loop:
```rust
while let Some(msg) = rx.recv().await {
    //  loop body
};
```

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
```diff
@@ -0,0 +1,65 @@
...
+        if ts.duration_since(base_ts) > RATE_INTERVAL {
+            base_ts = ts;
+            commit_rates(&mut counters_map);
+        }
+        process_counter(&mut counters_map, k, v);
+    }
+}
+
+fn process_counter(counters_map: &mut HashMap<String, u64>, k: String, v: u64) {
+    let entry = counters_map.entry(k).or_insert_with(|| 0);
```
pyakushin, line 56: you don't do anything inside closure, use just `or_insert`

bob-common/src/metrics/exporters/graphite_exporter/mod.rs
```diff
@@ -113,28 +108,61 @@ impl GraphiteRecorder {
...
         )));
     }
 
     #[allow(clippy::cast_sign_loss)]
-    fn update_gauge(&self, key: Key, value: i64) {
+    fn update_gauge(&self, key: &Key, value: GaugeValue) {
+        let val = if let GaugeValue::Absolute(val) = value {
+            val
+        } else {
+            todo!("Diffs are not supported at the moment");
```
pyakushin, line 153: I think it's dangerous leave panic here

bob/src/grinder.rs
```diff
@@ -86,7 +86,7 @@ impl Grinder {
                 counter!(GRINDER_PUT_ERROR_COUNT_COUNTER, 1);
             }
 
-            timing!(GRINDER_PUT_TIMER, time.elapsed().as_nanos() as u64);
+            histogram!(GRINDER_PUT_TIMER, time.elapsed().as_nanos() as f64);
```
pyakushin, line 89: Maybe it's better to import macro `histogram as timing`?

config-examples/prometheus.yml
pyakushin: @ikopylov it is prometheus config

@Justarone Are you sure we need this in repo? 
If yes, add comment that this is not bob config file

pyakushin (changes requested):

bob-common/src/metrics/exporters/global_exporter/mod.rs
Justarone: It's trait implementation, and trait's structure is defined in `metrics` libs.
I don't understand you, you wan't to make register_`metric` functions to be wrappers above register_with or what?

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
Justarone: Yes, that's incorrect, it's size in messages

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
Justarone: This method is called from sync function (increment_counter), which in its turn defined in trait from metrics crate (I can create runtime and store it somewhere in global exporter, but there would be an overload; don't think, that's good solution)

bob-common/src/metrics/exporters/graphite_exporter/mod.rs
Justarone: Change on error log

bob-common/src/metrics/exporters/global_exporter/rate_processor.rs
Justarone: I'll do that, but this method is called from immutable method of global exporter (which is defined in trait from `metrics`) and if we want to do some shutdown logic on closed, we need:
1. Store rate processor in global exporter with some interior mutability wrapper (and this wrapper should be thread-safe, because I've already have had problems with `metrics` crate: their unsafe code makes possible `increment_counter` sync function to run concurrently)
2. Store state of processor via interior mutability pattern (also thread-safe implementation)
3. Unsafe code (and sync logic also, which should be with doubled attention)

Current version (for me) seems better: if error is `Closed`, then receiver was dropped due to rate_processor_task function end (we don't need to cleanup it) and `try_send` will just return error

I'll implement it with `is_active` atomic bool marker in processor

bob-common/src/metrics/exporters/global_exporter/mod.rs
pyakushin: Yes, you do the same operation 3 times but with different metrics. But maybe it unnecessary code complication

pyakushin (approved):
LGTM

PR #262, opened 2021-07-11
Blob recovery tool with verification

vovac12:


bob-tools/bin/brt.rs
```diff
@@ -250,6 +356,138 @@ fn try_main() -> AnyResult<()> {
...
+fn validate_blob(path: &str) -> AnyResult<()> {
+    let mut reader = BlobReader::from_path(path)?;
+    reader.read_header()?;
+    while !reader.is_eof() {
+        reader.read_record()?;
+    }
+    Ok(())
+}
+
+fn validate_blobs_recursive(path: &str, settings: &ValidateSettings) -> AnyResult<Vec<String>> {
```
idruzhitskiy, line 388: Isn't this and validate_index_recursive have almost the same structure and differ only in leaf processing? It's better to factor out dir tree processing then

bob-apps/bin/brt.rs
```diff
@@ -250,12 +367,216 @@ fn try_main() -> AnyResult<()> {
...
+    if settings.delete && !result.is_empty() {
+        if settings.confirm && !ask_confirmation("Delete invalid index files?")? {
+            return Ok(());
+        }
+
+        let mut fails = vec![];
+        for path in result {
+            match std::fs::remove_file(&path) {
+                Ok(_) => {
+                    log::info!("[{}] index file removed", path);
```
idruzhitskiy, line 545: In some places there are `log::info`, in others - `println!`, it's better to use only one method of output (preferably logs)

bob-apps/bin/brt.rs
```diff
@@ -250,12 +367,216 @@ fn try_main() -> AnyResult<()> {
...
+            return Ok(path);
+        }
+        i += 1;
+    }
+}
+
+fn validate_blob_command(settings: &ValidateBlobSettings) -> AnyResult<()> {
+    let result = validate_recursive(&settings.path, &settings.blob_suffix, validate_blob)?;
+
+    if !result.is_empty() {
```
idruzhitskiy, line 487: `validate_blob_command` and `validate_index_command` have plenty in common :) IMO you can extract part with printing blobs (`print_corrupted(v: Vec<..>, name: &str)`, so it's like `print_invalid(result, "blobs")`). Probably same for `print_fails` (or maybe they are the same function). An optionally there can be some function like `collect_fails`, but maybe there is too much to cusomtize between two functions

bob-apps/bin/brt.rs
```diff
@@ -299,45 +624,170 @@ fn validate_bytes(a: &[u8], checksum: u32) -> AnyResult<()> {
...
-                .value_of(OUTPUT_BLOB_OPT)
-                .expect("Required")
-                .to_string(),
-            validate_every: matches
-                .value_of(VALIDATE_EVERY_OPT)
-                .expect("Has default")
-                .parse()?,
-        })
+        match matches.subcommand() {
+            (RECOVERY_COMMAND, Some(matches)) => Ok(Settings::Recovery(RecoverySettings {
```
idruzhitskiy, line 659: Probably it's better to move each branch matches parsing to the corresponding structure constructor

bob-apps/bin/brt.rs
vovac12: Now `println` is only used to print the list of corrupted blobs or indexes, since `env_logger` prints messages to `stderr`, you can use the list of corrupted blobs in other scripts via pipe.

bob-apps/bin/brt.rs
vovac12: Done

bob-apps/bin/brt.rs
vovac12: Done

bob-apps/bin/blob_recovery_tool/command.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+}
+
+pub struct ValidateBlobCommand {
+    path: PathBuf,
+    blob_suffix: String,
+    backup_suffix: String,
+    validate_every: usize,
+    fix: bool,
+    confirm: bool,
```
pyakushin, line 81: It's not clear what's the purpose of this field. And why in args it's no-confirm

bob-apps/bin/blob_recovery_tool/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+
+pub(crate) fn validate_files_recursive<F>(
+    path: &Path,
+    suffix: &str,
+    mut function: F,
+) -> AnyResult<Vec<String>>
+where
+    F: FnMut(&Path) -> AnyResult<()> + Clone,
+{
+    log::info!("Start validation");
```
pyakushin, line 74: You can add
```rust
#[macro_use]
extern crate log;
```

bob-apps/bin/blob_recovery_tool/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+    let mut result = vec![];
+    let mut count = 0;
+    let mut last_log = Instant::now();
+    for_each_file_recursive(path, |entry_path| {
+        let entry_path = entry_path.as_os_str().to_str().unwrap().to_string();
+        if entry_path.ends_with(&suffix) {
+            match function(entry_path.as_ref()) {
```
pyakushin, line 81: I think `if let Err(e) = ...` is better

bob-apps/bin/blob_recovery_tool/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+{
+    if input.as_ref() == output.as_ref() {
+        return Err(anyhow::anyhow!(
+            "Recovering into same file is not supported"
+        ));
+    }
+    let should_validate_written = validate_every != 0;
```
pyakushin, line 185: I think it's better to rename `should_validate_written`, now it's unclear what this flag does

pyakushin: Maybe `need_validation` or smth

ikopylov:
> 
> 
> > BPlusTree pull request has changed the Index file format, so the tool should be updated to understand it
> 
> I think it's better to move readers to `pearl`(as you mentioned above) to resolve this issue, because BPlusTree index has a more complex structure. IMO it should be a separate task, because PR will be locked until related PR is not merged in `pearl`.

Yes, I agree. This should definitely be a separate task

PR #263, opened 2021-07-11
Disable aio in config if it doesn't start on system

### LINKED ISSUE #257: Cache AIO availability check results
Currently, AIO availability check is performed for every blob. This is redundant and it is better to check once and then reuse its results

Justarone:
Closes #257

### DIFF
--- a/bob-common/src/configs/node.rs
+++ b/bob-common/src/configs/node.rs
@@ -6,6 +6,7 @@ use super::{
 use crate::data::DiskPath;
 use futures::Future;
 use humantime::Duration as HumanDuration;
+use std::sync::{Arc, Mutex};
 use std::{
     cell::{Ref, RefCell},
     env::VarError,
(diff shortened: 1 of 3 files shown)


bob-common/src/configs/node.rs
```diff
@@ -233,7 +234,7 @@ pub struct Pearl {
     #[serde(default = "Pearl::default_hash_chars_count")]
     hash_chars_count: u32,
     #[serde(default = "Pearl::default_enable_aio")]
-    enable_aio: bool,
+    enable_aio: Arc<Mutex<bool>>,
```
ikopylov, line 237: Why not `AtomicBool`? It should be faster than `Mutex`

idruzhitskiy:
LGTM

bob-common/src/configs/node.rs
```diff
@@ -6,15 +6,20 @@ use super::{
...
     cell::{Ref, RefCell},
     env::VarError,
     fmt::Debug,
     net::SocketAddr,
+    sync::atomic::AtomicBool,
     time::Duration,
 };
 use tokio::time::sleep;
 
+const AIO_FLAG_ORDERING: Ordering = Ordering::SeqCst;
```
ikopylov, line 21: Too strict. In this case, a relaxed order is sufficient

PR #269, opened 2021-07-20
Add http_api_port to node config

### LINKED ISSUE #268: Add REST API port to node config
It should be possible to change the REST API port from config

vovac12:
Closes #268


bob-apps/bin/bobd.rs
```diff
@@ -55,10 +55,16 @@ async fn main() {
...
     bob.run_backend().await.unwrap();
     info!("Start API server");
-    let http_api_port = matches
+    let http_api_port = if let Some(port) = matches
         .value_of("http_api_port")
         .and_then(|v| v.parse().ok())
-        .expect("expect http_api_port port");
+    {
+        port
+    } else if let Some(port) = node.http_api_port() {
```
idruzhitskiy, line 63: You can use .or(node.http_api_port()) in the first if condition. And IMO it's better to factor out `matches.value_of("http_api_port").and_then(|v| v.parse().ok())` into some variable, like `requestedPort` or smth else

pyakushin (approved):

bob-apps/bin/bobd.rs
vovac12: I set default `http_api_port` value in config to 8000 and replace `if` with `unwrap_or`

ikopylov:
It would be useful to have an ability to configure bind address as well

ikopylov (changes requested):
We need the ability to configure bind address

vovac12:
@ikopylov Done

config-examples/node.yaml
```diff
@@ -32,6 +32,10 @@ cleanup_interval: 1h
 open_blobs_soft_limit: 2
 # hard limit for count of max blobs to remain in ram
 open_blobs_hard_limit: 10
+# http REST api port. Defaults to 8000
+http_api_port: 9889
+# http REST api address. Defaults to 0.0.0.0
+http_api_address: 127.0.0.1
```
ikopylov, line 38: I suggest to write "0.0.0.0" here by default

config-examples/node.yaml
```diff
@@ -32,6 +32,10 @@ cleanup_interval: 1h
 open_blobs_soft_limit: 2
 # hard limit for count of max blobs to remain in ram
 open_blobs_hard_limit: 10
+# http REST api port. Defaults to 8000
+http_api_port: 9889
```
ikopylov, line 36: Same here. It is better to write default values in example config

ikopylov (approved):
LGTM

PR #275, opened 2021-07-27
Possible metrics gap fix

### LINKED ISSUE #274: Metrics transferred to graphite contain gaps
In graphite all metrics periodically have gaps. It was noted that these gaps often occur under load

Justarone:
Seems like this is the reason why metrics have "gaps" on graphite side, but maybe this reason is not the only one
Closes #274

### DIFF
--- a/bob-common/src/metrics/exporter/send.rs
+++ b/bob-common/src/metrics/exporter/send.rs
@@ -1,11 +1,13 @@
 use log::{debug, trace};
 use std::collections::HashMap;
 use std::time::{Duration, Instant};
-use tokio::sync::mpsc::Receiver;
+use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::time::timeout;
 
 use super::retry_socket::RetrySocket;
 use super::{Metric, MetricInner, MetricKey, MetricValue, TimeStamp};
+
+const TCP_CHANNEL_SIZE: usize = 1024;
 
 // this function runs in other thread, so it would be better if it will take control of arguments
 // themselves, not just references
@@ -16,8 +18,7 @@ pub(super) async fn send_metrics(
     send_interval: Duration,
     prefix: String,
 ) {
-    let mut socket =
-        RetrySocket::new(address.parse().expect("Can't read address from String")).await;
+    let socket_sender = spawn_tcp_sender_task(address).await;
     let mut counters_map = HashMap::new();
     let mut gauges_map = HashMap::new();
     let mut times_map = HashMap::new();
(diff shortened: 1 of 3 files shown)


bob-common/src/metrics/exporter/send.rs
```diff
@@ -18,34 +18,58 @@ pub(super) async fn send_metrics(
...
             }
+            current_time = Instant::now();
         }
+        let mut res_string = String::new();
+        let ts = chrono::Local::now().timestamp();
+        flush_counters(&counters_map, &mut res_string, &prefix, ts).await;
+        flush_gauges(&gauges_map, &mut res_string, &prefix, ts).await;
+        flush_times(&mut times_map, &mut res_string, &prefix, ts).await;
+        if let Err(e) = socket_sender.send(res_string).await {
+            debug!("Can't send data to tcp sender task (reason: {})", e);
```
pyakushin, line 47: I think it should be `warn` level. Because if sending failed there something serious happened with receiver side

bob-common/src/metrics/exporter/send.rs
```diff
@@ -1,13 +1,13 @@
...
-use tokio::time::{interval, timeout};
+use std::time::{Duration, Instant};
+use tokio::sync::mpsc::{channel, Receiver, Sender};
+use tokio::time::timeout;
 
 use super::retry_socket::RetrySocket;
 use super::{Metric, MetricInner, MetricKey, MetricValue, TimeStamp};
 
-const METRICS_RECV_TIMEOUT: Duration = Duration::from_millis(100);
+const TCP_CHANNEL_SIZE: usize = 1024;
```
pyakushin, line 10: I suggest to name it differently, like `TCP_SENDER_QUEUE_SIZE` or `TCP_SENDER_BUFFER_SIZE` or smth like that

bob-common/src/metrics/exporter/send.rs
```diff
@@ -18,34 +18,58 @@ pub(super) async fn send_metrics(
...
+    let (tx, rx) = channel(TCP_CHANNEL_SIZE);
+    let socket = RetrySocket::new(address.parse().expect("Can't read address from String")).await;
+    tokio::spawn(tcp_sender_task(socket, rx));
+    tx
+}
+
+async fn tcp_sender_task(mut socket: RetrySocket, mut rx: Receiver<String>) {
+    while let Some(data) = rx.recv().await {
+        if let Err(e) = socket.write_all(data.as_bytes()).await {
+            debug!(
```
pyakushin, line 62: Maybe also `warn`. It's not critical system, but it would be helpful to now if metrics export fails

PR #283, opened 2021-08-10
Fix actual holder creation condition

Justarone:


ikopylov:
@pyakushin can you take a look?

PR #286, opened 2021-08-11
Add brt mode to reverse byte order

### LINKED ISSUE #281: Add a mode to `brt` that reverses the byte order of the keys
In this mode `brt` should recursively scan directories and copy all BLOBs to another location with the same directory structure and with reversed byte order of the keys. I think, that doing update in place is dangerous. `brt` should read BLOB parts to memory buffer (size is configurable), update byte order inside memory and then write the updated data

vovac12:
Closes #281


bob-apps/bin/blob_recovery_tool/utils.rs
```diff
@@ -185,13 +244,16 @@ where
     info!("Input blob header version: {}", header.version);
     let mut count = 0;
     while !reader.is_eof() {
-        match reader.read_record_with_skip_wrong() {
+        match reader
+            .read_record_with_skip_wrong()
```
ikopylov, line 248: This function is also used for migration, in which case it should not skip invalid entries. Migration must explicitly fail for broken BLOBs

bob-apps/bin/blob_recovery_tool/structs.rs
```diff
@@ -119,6 +134,21 @@ pub(crate) struct Record {
     pub data: Vec<u8>,
 }
 
+impl Record {
+    pub(crate) fn migrate(self, source: u32, target: u32) -> AnyResult<Self> {
+        match (source, target) {
+            (source, target) if source == target => Ok(self),
+            (1, 2) => self.mirgate_v1_to_v2(),
```
ikopylov, line 141: This is probably an error: current BLOB version is '1' and previous is '0'

bob-apps/bin/blob_recovery_tool/command.rs
```diff
@@ -248,13 +251,106 @@ impl ValidateIndexCommand {
...
+    input_path: PathBuf,
+    output_path: PathBuf,
+    blob_suffix: String,
+    validate_every: usize,
+    target_version: u32,
+}
+
+impl MigrateCommand {
+    fn run(&self) -> AnyResult<()> {
+        process_files_recursive(
```
ikopylov, line 264: It would be nice if `migrate` command can handle a single file. If `input_path` is a directory then the command recursively processes all BLOBs, if it is a specific file then it only processes that single file

ikopylov: Actually, the same behaviour would be good also for `validate-blob` and `validate-index`

bob-apps/bin/blob_recovery_tool/structs.rs
```diff
@@ -119,6 +134,21 @@ pub(crate) struct Record {
     pub data: Vec<u8>,
 }
 
+impl Record {
+    pub(crate) fn migrate(self, source: u32, target: u32) -> AnyResult<Self> {
+        match (source, target) {
+            (source, target) if source == target => Ok(self),
```
ikopylov, line 140: `if source >= target` is better

vovac12: Fixed

bob-apps/bin/blob_recovery_tool/command.rs
vovac12: Done

bob-apps/bin/blob_recovery_tool/structs.rs
vovac12: Fixed

bob-apps/bin/blob_recovery_tool/utils.rs
vovac12: Fixed. Also I made skipping optional via flag `--skip-wrong-record`

pyakushin (approved):

PR #295, opened 2021-08-17
255 rest api metrics

idruzhitskiy:


Justarone (commented):
Some minor issues discussed in team chat

bob-common/src/metrics/collector/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub mod recorder;
+pub mod snapshot;
+
+pub(crate) use self::recorder::MetricsRecorder;
+pub use self::snapshot::SharedMetricsSnapshot;
+
+pub(crate) fn establish_global_collector(
+    check_interval: Duration,
+) -> (MetricsRecorder, SharedMetricsSnapshot) {
+    const BUFFER_SIZE: usize = 1_048_576; // 1 Mb
```
Justarone, line 17: I had this comment somewhere before, but it confuses: that's not Mbs, that's max amount of elements of type `Metric` structs in channel buffer

bob-common/src/metrics/exporters/graphite_exporter/mod.rs
```diff
@@ -89,82 +59,13 @@ impl GraphiteBuilder {
...
-        )));
-    }
-
-    fn record_histogram(&self, key: &Key, value: f64) {
-        self.push_metric(Metric::Time(MetricInner::new(
-            key.name().to_owned(),
-            value as u64,
-            -1,
-        )));
+    // TODO move recorder and snapshot to other place?
```
Justarone, line 62: What do you mean? Move in project hierarchy?

bob-common/src/metrics/exporters/graphite_exporter/send.rs
```diff
@@ -179,12 +95,12 @@ async fn flush_gauges(
...
+    times_map: &HashMap<MetricKey, TimeEntry>,
     res_string: &mut String,
     prefix: &str,
     ts: i64,
 ) {
-    for (key, entry) in times_map.iter_mut() {
+    for (key, entry) in times_map.iter() {
         let mean_time = match entry.measurements_amount {
             0 => entry.mean.expect("No mean time provided"),
             val => entry.summary_time / val,
```
Justarone, line 106: I am not sure, but seems like we can take mean value here, because during snapshot creation mean value is calculated
UPD: it's [here](https://github.com/qoollo/bob/pull/295/files#diff-c891e00c677c35441c75d08c74b5df56bd0bfe50e426bb9dffaa2f2f50f33adcR44), but maybe it will be less obvious, so it would be better to leave a comment about this

Justarone (changes requested):

bob-common/src/metrics/exporters/graphite_exporter/mod.rs
idruzhitskiy: Forgot to remove this comment after refactoring, it's already done (different mod)

bob-common/src/metrics/collector/mod.rs
idruzhitskiy: Fixed comment

bob-common/src/metrics/exporters/graphite_exporter/send.rs
idruzhitskiy: You're right, we already have mean value in entry. I don't think we need another comment for this, as "mean" is self explanatory

PR #301, opened 2021-08-19
293-offload-bloom-filters

### LINKED ISSUE #293: Offload bloom filters when memory limit is reached
The memory limit parameter for Bloom filters must be added to config. When the total amount of memory occupied by Bloom filters exceeds the threshold value, the offload procedure should be triggered. It removes Bloom filters  from memory for the oldest BLOBs

vovac12:
Closes #293


bob-backend/src/pearl/disk_controller.rs
```diff
@@ -503,4 +503,16 @@ impl DiskController {
             group.close_unneeded_active_blobs(soft, hard).await;
         }
     }
+
+    pub(crate) async fn filter_memory_allocated(&self) -> usize {
+        let mut memory = 0;
+        for holder in self.groups.read().await.iter() {
```
idruzhitskiy, line 509: Maybe it's better to use `FuturesUnordered` here, so requests will be performed in parallel?

idruzhitskiy: And looks like it's `group`, not holder

bob-backend/src/pearl/group.rs
```diff
@@ -448,6 +449,14 @@ impl Group {
             _ => x.end_timestamp().cmp(&y.end_timestamp()),
         });
     }
+
+    pub(crate) async fn filter_memory_allocated(&self) -> usize {
+        let mut memory = 0;
+        for holder in self.holders().read().await.iter() {
```
idruzhitskiy, line 455: Same, `FuturesUnordered`

bob-backend/src/pearl/stuff.rs
```diff
@@ -98,4 +101,58 @@ impl Stuff {
...
+            );
+        }
+    }
+
+    async fn collect_holders(
+        iter: impl IntoIterator<Item = &Arc<DiskController>>,
+    ) -> Vec<(SimpleHolder, usize)> {
+        let mut res = vec![];
+        for dc in iter {
+            for group in dc.groups().read().await.iter() {
```
idruzhitskiy, line 146: Again, this cycle's iterations don't seem to be related to each other, so you can process them in parallel

bob/src/cleaner.rs
```diff
@@ -25,14 +28,26 @@ impl Cleaner {
...
+        soft: usize,
+        hard: usize,
+        filter_memory_limit: Option<usize>,
+    ) {
         let mut interval = interval(t);
         loop {
             interval.tick().await;
             backend.close_unneeded_active_blobs(soft, hard).await;
+            if let Some(limit) = filter_memory_limit {
+                log::warn!("Memory before: {}", backend.filter_memory_allocated().await);
```
idruzhitskiy, line 47: IMO it should be `info`, but it's debatable

bob/src/version_helpers.rs
```diff
@@ -12,9 +11,11 @@ pub fn get_bob_build_time() -> &'static str {
 }
 
 pub fn get_pearl_version() -> String {
-    pearl::get_pearl_version()
+    //pearl::get_pearl_version()
```
idruzhitskiy, line 14: ?

ikopylov: Same question

bob-backend/src/pearl/postprocessor.rs
```diff
@@ -0,0 +1,109 @@
...
+            filter_memory_limit,
+            ..Default::default()
+        }
+    }
+    pub(crate) async fn storage_prepared(&self, holder: impl Into<SimpleHolder>) {
+        let mut holders = self.holders.write().await;
+        let holder = holder.into();
+        let filter_memory = holder.filter_memory_allocated().await;
+        holders.insert(holder);
+        self.allocated_size
```
ikopylov, line 82: As I see, only this function have an access to `allocated_size` and it does not work in parallel. If I am right, `Atomic` here is overkill and can be replaced with a simple integer value

bob-backend/src/pearl/postprocessor.rs
```diff
@@ -0,0 +1,109 @@
...
+                    holder.offload_filter().await;
+                    let size_after = holder.filter_memory_allocated().await;
+                    log::debug!(
+                        "{} -> {}: {} freed",
+                        size_before,
+                        size_after,
+                        size_before.saturating_sub(size_after)
+                    );
+                    self.allocated_size
+                        .fetch_sub(size_before.saturating_sub(size_after), Ordering::Relaxed);
```
ikopylov, line 101: Could this just be `self.allocated_size = size_after`? What is the reason for this math?

bob/src/cleaner.rs
ikopylov: +1

ikopylov (changes requested):

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -503,4 +503,16 @@ impl DiskController {
...
+            memory += holder.filter_memory_allocated().await;
+        }
+        memory
+    }
+
+    pub(crate) fn groups(&self) -> Arc<RwLock<Vec<Group>>> {
```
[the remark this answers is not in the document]
vovac12: It make it visible only in current crate. Needed for stuff.rs

bob-backend/src/pearl/postprocessor.rs
```diff
@@ -0,0 +1,109 @@
...
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use crate::{
+    pearl::{holder::PearlSync, Holder},
+    prelude::*,
+};
+
+#[derive(Clone)]
+pub(crate) struct SimpleHolder {
```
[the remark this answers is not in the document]
vovac12: I created this struct to avoid of copying all fields of Holder, because it contains PearlConfig with a lot of fields

bob-backend/src/pearl/postprocessor.rs
vovac12: But this structure is cloned to different dcs, groups and holders, and they can call this method in parallel

bob-backend/src/pearl/core.rs
```diff
@@ -196,4 +203,27 @@ impl BackendStorage for Pearl {
             dc.close_unneeded_active_blobs(soft, hard).await;
         }
     }
+
+    async fn offload_old_filters(&self, limit: usize) {
+        Stuff::offload_old_filters(
```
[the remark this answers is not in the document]
vovac12: It's not my naming) I can rename it to Utils

bob-backend/src/pearl/core.rs
```diff
@@ -96,8 +101,10 @@ impl BackendStorage for Pearl {
         let start = Instant::now();
         let futs = FuturesUnordered::new();
         let alien_iter = once(&self.alien_disk_controller);
+        let postprocessor = PostProcessor::new(self.filter_memory_limit);
```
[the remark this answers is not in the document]
vovac12: Do you have any suggestions? Because I really don't have any other variants

bob-backend/src/pearl/core.rs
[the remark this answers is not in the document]
ikopylov: `Utils` is a little bit better, but still a bad name. I don't see the whole picture, so maybe @pyakushin or @idruzhitskiy can suggest a name

bob-backend/src/pearl/group.rs
```diff
@@ -76,19 +76,20 @@ impl Group {
...
         self.add_range(holders).await;
         debug!("{}: start holders", self);
-        self.run_pearls().await
+        self.run_pearls(pp).await
     }
 
     pub async fn remount(&self) -> AnyResult<()> {
         self.holders.write().await.clear();
-        self.run().await
+        self.run(Default::default()).await
```
ikopylov, line 84: This can lead to a problems. Why do you pass a default instance?

bob-backend/src/pearl/core.rs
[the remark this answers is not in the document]
ikopylov: For example, `BloomFilterMemoryLimiter` is much better, because it shows the purpose of this struct

bob-backend/src/pearl/postprocessor.rs
ikopylov: Ok

bob-backend/src/pearl/group.rs
vovac12: It's actually do nothing, but yeah it's not clear. Now I use another solution

ikopylov: Ok, approach with `Hooks` trait is much better

bob-backend/src/pearl/stuff.rs
```diff
@@ -129,6 +132,86 @@ impl Stuff {
...
+                current_size,
+                freed
+            );
+        }
+    }
+
+    async fn collect_holders(
+        iter: impl IntoIterator<Item = &Arc<DiskController>>,
+    ) -> Vec<(SimpleHolder, usize)> {
+        iter.into_iter()
```
ikopylov, line 175: This approach violates SOLID principle. It is better to expose `holder_iter()` function on each level, that provides iterator on holders and uses `holder_iter()` on level below. With such approach, here you can simply call top level `holder_iter()` and convert its result into `SimpleHolder`

bob-backend/src/pearl/hooks.rs
```diff
@@ -0,0 +1,125 @@
...
+                    let size_before = holder.filter_memory_allocated().await;
+                    holder.offload_filter().await;
+                    let size_after = holder.filter_memory_allocated().await;
+                    log::debug!(
+                        "{} -> {}: {} freed",
+                        size_before,
+                        size_after,
+                        size_before.saturating_sub(size_after)
+                    );
+                    self.allocated_size.store(size_after, Ordering::Relaxed);
```
ikopylov, line 117: I reread the code. My previous suggestion just to store the value is incorrect. `size_after` contains filter size of a single holder. You need to revert change and perform math: `allocated_size = allocated_size - (size_before - size_after)`

bob-backend/src/pearl/postprocessor.rs
ikopylov: Sorry, but my suggestion is incorrect

pyakushin (approved):

PR #306, opened 2021-08-19
Fix panic on load operation

### LINKED ISSUE #302: Panic on atomic operations with AcqRel Ordering
I started bobd with default configs and received following panic:
```
thread 'main' panicked at 'there is no such thing as an acquire/release load', /rustc/30a0a9b694cde95cbab863f7ef4d554f0f46b606/library/core/src/sync/atomic.rs:2359:23
```
As I found in this [issue comment](https://github.com/rust-lang/rust/issues/49127#issuecomment-373984912) this happens if you use AcqRel ordering in just store | load operations.

AcqRel ordering is used [here](https://github.com/qoollo/bob/blob/master/bob-common/src/metrics/exporters/global_exporter/rate_processor.rs#L11)

Justarone:
Closes #302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Bob versions changelog
 #### Fixed
 
 - Add brt to dockerfiles (https://github.com/qoollo/bob/pull/296)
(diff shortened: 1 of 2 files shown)


ikopylov:
LGTM

pyakushin (approved):

PR #307, opened 2021-08-20
298 build bob versions with 8 byte and 16 byte keys

### LINKED ISSUE #298: Build Bob versions with 8-byte and 16-byte keys
These versions should be published to Docker Hub and attached to GitHub release.
I suggest to use build matrix: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

alexprof:
Closes #298


.github/workflows/build-docker-images.yml
```diff
@@ -5,53 +5,85 @@ on:
...
           file: dockerfiles/ubuntu/Dockerfile
           push: true
-          tags: ${{ secrets.DOCKERHUB_USERNAME }}/bob:${{ env.VERSION }}-ubuntu20.10
-          build-args: BOB_COMMIT_HASH=${BOB_COMMIT_HASH}
\ No newline at end of file
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            BOB_COMMIT_HASH=${BOB_COMMIT_HASH}
+            BOB_KEY_SIZE=${BOB_KEY_SIZE}
```
ikopylov, line 89: Same here (`${{ matrix.bob_key_size }}`)

.github/workflows/build-packages.yml
```diff
@@ -117,42 +134,50 @@ jobs:
...
   upload-deb:
     runs-on: ubuntu-latest
     needs: build-deb
     environment: prod
+    strategy:
+      matrix:
+        bob_key_size: [8, 16]
+        include:
+          - bob_key_size: 16
+            asset_name_postfix: _key16
```
ikopylov, line 149: You can use conditional step that generates this name and save it to environment variable only when `bob_key_size != 8`

.github/workflows/build-packages.yml
```diff
@@ -64,41 +72,50 @@ jobs:
...
   upload-rpm:
     runs-on: ubuntu-latest
     needs: build-rpm
     environment: prod
+    strategy:
+      matrix:
+        bob_key_size: [8, 16]
+        include:
+          - bob_key_size: 16
+            asset_name_postfix: -key16
```
ikopylov, line 87: You can use conditional step that generates this name and save it to environment variable only when bob_key_size != 8

.github/workflows/build-packages.yml
```diff
@@ -168,27 +193,33 @@ jobs:
...
   upload-zip:
     runs-on: ubuntu-latest
     needs: zip-file
     environment: prod
+    strategy:
+      matrix:
+        bob_key_size: [8, 16]
+        include:
+          - bob_key_size: 16
+            asset_name_postfix: _key16
```
ikopylov, line 208: Same here: You can use conditional step that generates this name and save it to environment variable only when bob_key_size != 8

PR #309, opened 2021-08-20
add default config to dockerfiles

idruzhitskiy:


.gitignore
```diff
@@ -1,6 +1,6 @@
 # Generated by Cargo
 # will have compiled files and executables
-/target/
+**/target/
```
ikopylov, line 3: This can potentially ignore some valid `target` directories. Maybe it is better to specify full path here

dockerfiles/default-configs/cluster.yaml
```diff
@@ -0,0 +1,31 @@
...
+      # [str] node name
+    - name: local_node
+      # [ip:port] node address
+      address: 127.0.0.1:20000  
+      # [list] of physical node disks
+      disks:
+          # [str] name of disk, the node uses it to determine which vdisks belong to it
+        - name: disk1
+          # [str] mount point, it doesn't matter if it is a whole physical disk or a directory
+          path: /tmp/d1
```
ikopylov, line 20: Do not use `tmp` directory for data. This directory uses a special filesystem that can lead to unwanted behavior

ikopylov: I suggest to create and use '/bob/data/` directory

dockerfiles/default-configs/logger.yaml
```diff
@@ -0,0 +1,87 @@
...
+    # [pattern/json] encoder, default - [pattern]
+    encoder:
+    # [format] described in https://docs.rs/log4rs/0.9.0/log4rs/encode/pattern/index.html
+      pattern: "{d(%Y-%m-%d %H:%M:%S):<20} {M:>20.30}:{L:>3} {h({l})}    {m}\n"
+
+  # An appender named "requests" that writes to a file with a custom pattern encoder
+  requests:
+    kind: file
+    # [path] to the log file
+    path: "/log/logs.log"
```
ikopylov, line 26: I suggest to use `/bob/log/` directory

dockerfiles/default-configs/node.yaml
```diff
@@ -0,0 +1,72 @@
...
+#  seconds, second, sec, s
+#  minutes, minute, min, m
+#  hours, hour, hr, h
+#  days, day, d
+#  weeks, week, w
+#  months, month, M -- defined as 30.44 days
+#  years, year, y -- defined as 365.25 days
+
+# [file] logger config file
+log_config: /configs/logger.yaml
```
ikopylov, line 16: I suggest to use `/bob/configs/` directory. This way all files associated with bob will be in `/bob/` directory

dockerfiles/default-configs/node.yaml
```diff
@@ -0,0 +1,72 @@
...
+open_blobs_soft_limit: 2
+# hard limit for count of max blobs to remain in ram
+open_blobs_hard_limit: 10
+
+# used only for 'backend_type: pearl'
+pearl:
+  # optional, default = true, enables linux AIO
+  enable_aio: true
+  # path to logfile with info about disks states switches, default = "/tmp/bob_events.csv"
+  disks_events_logfile: "/tmp/bob_events.csv"
```
ikopylov, line 41: Do not use `/tmp` for data. It is better to use `/bob/events/` or `/bob/log/` here

.gitignore
idruzhitskiy: Specified it to ignore targets in bob* directories (as well as the one in the root directory)

dockerfiles/default-configs/logger.yaml
idruzhitskiy: /bob/log

dockerfiles/default-configs/cluster.yaml
idruzhitskiy: Done

PR #318, opened 2021-09-02
Nodes authentication

pyakushin:


bob-access/src/authenticator/basic.rs
```diff
@@ -5,29 +5,63 @@ use super::{users_storage::UsersStorage, Authenticator};
 #[derive(Debug, Default, Clone)]
 pub struct Basic<Storage: UsersStorage> {
     users_storage: Storage,
+    nodes: Vec<Credentials>,
```
ikopylov, line 8: It is better to use `HashMap` here

PR #323, opened 2021-09-06
Fix Prometheus exporter bug

### LINKED ISSUE #322: Prometheus exporter doesn't work after migration to global exporter
That's my fault: I expected, that prometheus exporter will behave like graphite one:
`install = build + set_boxed_recorder`
but in case of prometheus exporter `exporter` should also be set to make metrics scrapable

Justarone:
Closes #322

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+Prometheus exporter bug which occured after migration to global exporter scheme (#322)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-Prometheus exporter bug which occured after migration to global exporter scheme (#322)
+- Prometheus exporter bug which occured after migration to global exporter scheme (#322)
 
 #### Updated
 
--- a/bob-common/src/metrics/mod.rs
+++ b/bob-common/src/metrics/mod.rs
@@ -8,6 +8,7 @@ use std::{
     sync::Arc,
     time::{Duration, Instant},
 };
+use tokio::{pin, select};
 
 mod exporters;
 pub mod pearl;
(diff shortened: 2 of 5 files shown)


bob-common/src/metrics/mod.rs
```diff
@@ -228,20 +229,31 @@ fn install_global_recorder(recorders: Vec<Box<dyn Recorder>>) {
...
+        .expect("Failed to set Prometheus exporter");
+
+    tokio::spawn(async move {
+        pin!(exporter);
+        loop {
+            select! {
+                _ = &mut exporter => {}
+            }
+        }
+    });
```
ikopylov, line 255: This looks a little bit strange. Why is this needed?

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+Prometheus exporter bug which occured after migration to global exporter scheme (#322)
```
ikopylov, line 13: Please, add `-` before the text

bob/src/build_time.rs
```diff
@@ -1 +1 @@
-pub const BUILD_TIME: &str = "time-undefined";
\ No newline at end of file
+pub const BUILD_TIME: &str = "06-09-2021(22:56:27)";
```
ikopylov, line 1: This looks like an unintentional change

bob-common/src/metrics/mod.rs
Justarone: I just checked [source code of metrics-prometheus-exporter builder's `install` method](https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-prometheus/src/builder.rs#L202) and found it there. 

As I understood (hope that I understood right), this code is needed because exporter has to be polled to process requests from prometheus server, which it [promethes server] sends to scrape metrics. Without this code incoming 'scraping' requests won't be processed

ikopylov: As I see, a thread with busy loop started here, and that thread do nothing (`{}`). I don't understand, where the 'scraping' requests processing performed. 

@pyakushin maybe you can understand how this block of code works

Justarone: [in tokio docs](https://docs.rs/tokio/1.11.0/tokio/macro.select.html) you may find this pattern for select:
`<pattern> = <async expression> (, if <precondition>)? => <handler>,`
and in our case `<async expression>` is handler itself (*), so if it completed (**), that means, that request is handled, and `{}` in this case is some kind of postprocessing (which we don't need to do)

(*) - this logic can be found [here](https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-prometheus/src/builder.rs#L276). You see, that exporter itself - future (or 'async block'), which prepares some stuff and calls `server.serve(..)` at the end.

(**) You may read about this [here](https://rust-lang.github.io/async-book/06_multiple_futures/03_select.html#interaction-with-unpin-and-fusedfuture). That's not `select!` which is used here, because I use the one from tokio, but logic is the same.

Justarone: @ikopylov

ikopylov: Ok, now I understand how it works. Thanks!

idruzhitskiy (approved):

ikopylov:
LGTM

PR #358, opened 2021-11-04
355 hardware metrics issues

### LINKED ISSUE #355: Hardware metrics issues
Sever issues with hardware metrics:
1. `total_ram` metric actually shows used RAM, which is misleading (https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L67)
2. The same thing for `total_space`. It should print the total space on disks (https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L63)
3. `descr_amout` is calculated correctly, but drops to zero after a short time. Then the correct value appears again, which is again reset to zero (`value` is not updated here: https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L124)

pyakushin:
Closes #355

### DIFF
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -154,7 +154,7 @@ impl DescrCounter {
                 d.count() as u64 - 4 // exclude stdin, stdout, stderr and `read_dir` instance
             }
             Err(e) => {
-                warn!("failed to count descriptors: {}", e);
+                debug!("failed to count descriptors: {}", e);
                 0 // proc is unsupported
             }
         }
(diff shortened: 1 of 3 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -146,10 +149,14 @@ impl DescrCounter {
...
-        if let Ok(d) = d {
-            d.count() as u64 - 4 // exclude stdin, stdout, stderr and `read_dir` instance
-        } else {
-            0 // proc is unsupported
+        match d {
+            Ok(d) => {
+                d.count() as u64 - 4 // exclude stdin, stdout, stderr and `read_dir` instance
+            }
+            Err(e) => {
+                warn!("failed to count descriptors: {}", e);
```
ikopylov, line 157: If `/proc/` is not available, this message will clog up the log file. Since descriptors counting is not a critical feature, it is better to print this log on `debug` level

ikopylov:
@pyakushin please, add a line to CHANGELOG.md

PR #378, opened 2021-11-25
Add bloom_filter_memory_limit to example config

### LINKED ISSUE #377: Add `bloom_filter_memory_limit` parameter to node config example
https://github.com/qoollo/bob/blob/master/config-examples/node.yaml

vovac12:
Close #377

### DIFF
--- a/config-examples/node.yaml
+++ b/config-examples/node.yaml
@@ -36,6 +36,8 @@ open_blobs_hard_limit: 10
 http_api_port: 8000
 # http REST api address
 http_api_address: 0.0.0.0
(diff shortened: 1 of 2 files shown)


config-examples/node.yaml
```diff
@@ -36,6 +36,8 @@ open_blobs_hard_limit: 10
 http_api_port: 8000
 # http REST api address
 http_api_address: 0.0.0.0
+# memory limit for all bloom filters. Unlimited if not specified
+bloom_filter_memory_limit: 1000000000
```
ikopylov, line 40: I think, the default limit should be larger. Please, set it to 8Gb (8589934592)

PR #379, opened 2021-11-27
376 disk metrics fix

### LINKED ISSUE #376: Incorrect calculation of used and free disk space in hardware metrics
It looks like disks that don't belong to Bob are also counted.

Code for disk space metric is here: https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L81

Boneyan:
Fixed calculation of disk space metrics
Closes #376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+- Disk space metrics calculation fix (#376)
 
 #### Updated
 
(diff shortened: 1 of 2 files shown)


Boneyan:
that was accident click, don't bother

Justarone (approved):
Yes, seems like in case when 2 disks, for example, have paths "/", "/opt/disk1" and virtual disk has path "/opt/disk1/..." this may work incorrect (choose "/" instead of "/opt/disk1", didn't noticed that)

ikopylov (approved):
LGTM

PR #387, opened 2021-12-02
359 use of lsof

### LINKED ISSUE #359: Use 'lsof' to count descriptors
`lsof` is expected to be faster

Boneyan:
Descriptors count metric now tries lsof | wc before trying /proc/self/fd/
Closes #359

### DIFF
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -4,6 +4,7 @@ use bob_common::metrics::{
     USED_SPACE,
 };
 use std::path::{Path, PathBuf};
+use std::process;
 use sysinfo::{DiskExt, ProcessExt, System, SystemExt};
 
 const DESCRS_DIR: &str = "/proc/self/fd/";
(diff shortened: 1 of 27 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -148,6 +149,32 @@ impl DescrCounter {
...
+        let lsof_str = format!("lsof -a -p {} -d ^mem -d ^cwd -d ^rtd -d ^txt -d ^DEL", process::id());
+        match pipers::Pipe::new(&lsof_str)
+                          .then("wc -l")
+                          .finally() {
+            Ok(proc) => {
+                match proc.wait_with_output() {
+                    Ok(output) => {
+                        if output.status.success() {
+                            let count = String::from_utf8(output.stdout).unwrap();
+                            let count = count[..count.len() - 1].parse::<u64>().unwrap();
```
ikopylov, line 161: This can lead to a panic. Disk metrics is not important feature, so we should avoid panic

bob/src/hw_metrics_collector.rs
```diff
@@ -148,6 +149,32 @@ impl DescrCounter {
...
+                            return count - 5;
+                        } else {
+                            debug!("something went wrong (fs /proc will be used): {}",
+                                String::from_utf8(output.stderr).unwrap());
+                        }
+                    },
+                    Err(e) => {
+                        debug!("lsof output wait error (fs /proc will be used): {}", e);
```
ikopylov, line 169: If `lsof` is not available, we must not reattempt to use it in future calls. There should be a flag, that controls its usage and set to `false` on error

bob/src/hw_metrics_collector.rs
```diff
@@ -111,27 +112,29 @@ const CACHED_TIMES: usize = 10;
 struct DescrCounter {
     value: u64,
     cached_times: usize,
+    lsof_flag: bool
```
idruzhitskiy, line 115: `lsof_available` or `lsof_enabled` are better names IMO

bob/src/hw_metrics_collector.rs
```diff
@@ -148,6 +151,36 @@ impl DescrCounter {
...
+            let lsof_str = format!("lsof -a -p {} -d ^mem -d ^cwd -d ^rtd -d ^txt -d ^DEL", process::id());
+            match pipers::Pipe::new(&lsof_str)
+                            .then("wc -l")
+                            .finally() {
+                Ok(proc) => {
+                    match proc.wait_with_output() {
+                        Ok(output) => {
+                            if output.status.success() {
+                                let count = String::from_utf8(output.stdout).unwrap();
+                                if let Ok(count) = count[..count.len() - 1].parse::<u64>() {
```
idruzhitskiy, line 164: If parsing failed, we should notify user about changed format (and disable lsof)

bob/src/hw_metrics_collector.rs
```diff
@@ -148,6 +151,36 @@ impl DescrCounter {
         //  |  10.000   |      0.006     |
         //  with payload
         //  |  10.000   |      0.018     |
+        if self.lsof_flag {
```
idruzhitskiy, line 154: I think you should extract it to separate function, `count_descriptors_by_lsof(...) -> Option<u64>`

idruzhitskiy: And probably use `if !self.lsof_flag { return None }` in this function to reduce indent levels

idruzhitskiy:
LGTM

ikopylov:
LGTM

PR #394, opened 2021-12-08
393 bob ram metric

### LINKED ISSUE #393: Metrics: add metric to track RAM occupied by Bob process
In addition to the overall RAM metrics, it would be good to see how memory is occupied exclusively by Bob process

Boneyan:
Collecting bob ram metric via fs /proc
Closes #393

### DIFF
--- a/bob/Cargo.toml
+++ b/bob/Cargo.toml
@@ -26,6 +26,7 @@ lazy_static = "1.4.0"
 log = "0.4"
 metrics = { version = "0.17", features = ["std"] }
 mockall = "0.10.2"
+page_size = "0.4"
 prost = "0.8"
 ring = "0.16"
 rocket = { version = "0.5.0-rc.1", features = ["json", "uuid"] }
--- a/bob/Cargo.toml
+++ b/bob/Cargo.toml
@@ -26,7 +26,6 @@ lazy_static = "1.4.0"
 log = "0.4"
 metrics = { version = "0.17", features = ["std"] }
 mockall = "0.10.2"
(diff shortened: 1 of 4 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -69,6 +70,8 @@ impl HWMetricsCollector {
             debug!("used mem in mb: {}", used_mem);
             gauge!(USED_RAM, used_mem as f64);
             gauge!(FREE_RAM, (total_mem - used_mem) as f64);
+            let bob_ram = bytes_to_mb(Self::bob_ram());
```
ikopylov, line 73: here is no reason to get RAM for a process directly from procfs. The `Sysinfo` crate already has the ability to get process memory. 

Also, please, review the `sys.refresh_*` calls above. I think `refresh_all` is not needed, it is better to refresh only required parts

ikopylov:
LGTM

bob/Cargo.toml
```diff
@@ -36,7 +36,7 @@ serde_yaml = "0.8"
 stopwatch = "0.0.7"
 termion = "1.5"
 thiserror = "1.0"
-sysinfo = "0.21"
+sysinfo = "0.21.2"
```
ikopylov, line 39: Why this was updated?

ikopylov, line 39: Why this was updated?

Boneyan: It was updated for some reason during feature development, don't remember why

PR #403, opened 2021-12-23
400 bloom filters ram

### LINKED ISSUE #400: Add metric to track RAM used by the Bloom filters
This metric should only be updated when a new blob is created or when a dump has occurred

Boneyan:
Added bloom filters ram metric (updating on put & filter offload)
Closes #400

### DIFF
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -6,6 +6,8 @@ use crate::{
     stub_backend::StubBackend,
 };
 
+use bob_common::metrics::BLOOM_FILTERS_RAM;
+
 pub const BACKEND_STARTING: f64 = 0f64;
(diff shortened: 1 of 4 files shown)


vovac12:
Did you measured performance impact? Because seems like it's not good approach to call `filter_memory_allocated ` on every put request, because it scans all holders and blobs to calculate allocated size. Check if it possible to determine when new blob was created and update metrics in such case.

bob-backend/src/core.rs
```diff
@@ -228,6 +231,11 @@ impl Backend {
         self.put_single(key, data, operation).await
     }
 
+    async fn update_metrics(&self) {
```
ikopylov, line 234: It is better to explicitly indicate in the method name what metric it updates

ikopylov:
LGTM

PR #407, opened 2021-12-27
404 space in http api

### LINKED ISSUE #404: Add REST API method to get occupied disk space
This is required for OldPartitionRemover. Initial implementation can use data from metrics

idruzhitskiy:
closes #404

### DIFF
--- a/bob/src/api/http.rs
+++ b/bob/src/api/http.rs
@@ -262,7 +262,7 @@ async fn get_space_info(bob: &State<BobServer>) -> Result<Json<SpaceInfo>, Statu
             })
         })
         .ok_or(StatusExt::new(
-            Status::NotFound,
+            Status::from_code(503).expect("codes changed"),
             false,
             "Failed to get space info".to_string(),
         ))
--- a/bob/src/api/http.rs
+++ b/bob/src/api/http.rs
@@ -301,11 +301,7 @@ async fn nodes(bob: &State<BobServer>) -> Json<Vec<Node>> {
             .filter_map(|vd| {
                 if vd.replicas.iter().any(|r| r.node == node.name()) {
                     let mut vd = vd.clone();
-                    for i in 0..vd.replicas.len() {
-                        if vd.replicas[i].node != node.name() {
-                            vd.replicas.remove(i);
-                        }
-                    }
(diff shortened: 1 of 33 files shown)


bob/src/api/http.rs
```diff
@@ -236,6 +245,29 @@ async fn status(bob: &State<BobServer>) -> Json<Node> {
...
 
+#[get("/status/space")]
+async fn get_space_info(bob: &State<BobServer>) -> Result<Json<SpaceInfo>, StatusExt> {
+    let snapshot = bob.metrics().read().await;
+    let total_space = snapshot.gauges_map.get(TOTAL_SPACE).map(|g| g.value);
+    let used_space = snapshot.gauges_map.get(USED_SPACE).map(|g| g.value);
+    let free_space = snapshot.gauges_map.get(FREE_SPACE).map(|g| g.value);
+    total_space
+        .zip(used_space)
+        .zip(free_space)
```
ikopylov, line 256: Why do you use `zip` on scalar values? It would be easier to construct `SpaceInfo` directly from `total_space`, `used_space` and `free_space` values

idruzhitskiy: They are `Option`s, this way if any of them is `None` the whole result will be none, no need to unwrap/check individual values

bob/src/hw_metrics_collector.rs
```diff
@@ -1,20 +1,20 @@
...
-use libc::statvfs;
+use sysinfo::{DiskExt, ProcessExt, RefreshKind, System, SystemExt};
 
 const DESCRS_DIR: &str = "/proc/self/fd/";
 
-struct DisksMetrics {
-    total_space: u64,
-    used_space: u64,
-    free_space: u64,
+pub(crate) struct SpaceMetrics {
```
ikopylov, line 14: IMHO `DiskSpaceMetrics` is better

ikopylov:
One minor suggestion, otherwise LGTM

bob/src/hw_metrics_collector.rs
idruzhitskiy: Done

ikopylov (approved):
LGTM

PR #411, opened 2022-01-20
376 fix disk space metric

### LINKED ISSUE #376: Incorrect calculation of used and free disk space in hardware metrics

Boneyan:
Collector now correctly calculates used disk space
Closes #376

### DIFF
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -7,11 +7,15 @@ use std::path::{Path, PathBuf};
 use std::process;
 use std::os::unix::fs::MetadataExt;
 use sysinfo::{DiskExt, ProcessExt, System, SystemExt};
-//use std::process::Command;
 use libc::statvfs;
-use std::num::Wrapping;
 
 const DESCRS_DIR: &str = "/proc/self/fd/";
+
+struct DisksMetrics {
+    total_space: u64,
+    used_space: u64,
+    free_space: u64,
+}
 
 pub(crate) struct HWMetricsCollector {
     disks: HashMap<PathBuf, String>,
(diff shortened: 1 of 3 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -80,32 +83,41 @@ impl HWMetricsCollector {
...
+        let mut total = 0;
+        let mut used = 0;
+        let mut free = 0;
+        
+        for mount_point in disks.keys() {
+            let cm_p = Self::to_cpath(mount_point.as_path());
+            unsafe {
+                let mut stat: statvfs = std::mem::zeroed();
+                if statvfs(cm_p.as_ptr() as *const _, &mut stat) == 0 {
+                    let bsize = Wrapping(stat.f_bsize as u64);
```
idruzhitskiy, line 109: Why do you use wrappings here?

bob/src/hw_metrics_collector.rs
```diff
@@ -80,32 +83,41 @@ impl HWMetricsCollector {
...
-                (total + disk_total, free + disk_free)
-            })
+    fn space(disks: &HashMap<PathBuf, String>) -> (u64, u64, u64) {
+        let mut total = 0;
+        let mut used = 0;
+        let mut free = 0;
+        
+        for mount_point in disks.keys() {
+            let cm_p = Self::to_cpath(mount_point.as_path());
+            unsafe {
```
idruzhitskiy, line 106: Unsafe blocks should be as small as possible, so I think you should create function, wrapping unsafe call and returning `Option<statvfs>`, and work with the result in safe code

bob/src/hw_metrics_collector.rs
```diff
@@ -80,32 +83,41 @@ impl HWMetricsCollector {
...
+                    let bavail = Wrapping(stat.f_bavail as u64);
+                    let bfree = Wrapping(stat.f_bfree as u64);
+                    total += (bsize * blocks).0;
+                    free += (bsize * bavail).0;
+                    used += ((blocks - bfree) * bsize).0;
+                }
+            }
+        }
+
+        (total, free, used)
```
idruzhitskiy, line 120: It's better to use proper struct here, as all members are of the same type, and it's easy to misuse result

ikopylov:
LGTM

PR #417, opened 2022-01-24
180 folder creation feature

### LINKED ISSUE #180: Do not create automatically all required folders for Bob during startup
The current behavior conflicts with the disk fail/restore feature: if the disk was disconnected and the bob service was restarted, new folders will be created directly on the system disk, which is incorrect.

For initial folder structure initialization, Bob should be started with an additional command line parameter. In this mode it creates a folder structure, if not presented, during startup (in fact, this just reproduces the current behavior).

If Bob is started without that parameter (normal stat), it should treat the missing folders as if they were on the disk that was taken offline.

### LINKED ISSUE #398: Panic in metrics when data directory is not created
During startup, Bob should check if the data directory exists. If not then panic with meaningful description (see related issue: https://github.com/qoollo/bob/issues/180). Panic in metrics should not be observed by user

Boneyan:
Bobd checks if there is bob and alien folders. With -i flag bobd will create the folders, without the flag bobd will panic
Closes #180 
Closes #398

### DIFF
--- a/bob-apps/bin/bobd.rs
+++ b/bob-apps/bin/bobd.rs
@@ -1,11 +1,13 @@
 use bob::{
-    build_info::BuildInfo, init_counters, BobApiServer, BobServer, ClusterConfig, Factory, Grinder,
+    build_info::BuildInfo, init_counters, BobApiServer, BobServer, ClusterConfig, NodeConfig, Factory, Grinder,
     VirtualMapper,
 };
 use clap::{crate_version, App, Arg, ArgMatches};
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use tokio::runtime::Handle;
 use tonic::transport::Server;
+use std::path::PathBuf;
+use std::fs::create_dir;
 
 #[macro_use]
 extern crate log;
@@ -35,6 +37,8 @@ async fn main() {
     let node = cluster.get(node_config).await.unwrap();
 
     log4rs::init_file(node.log_config(), Default::default()).expect("can't find log config");
+
+    check_folders(&node, matches.is_present("init folders"));
 
     let mut mapper = VirtualMapper::new(&node, &cluster).await;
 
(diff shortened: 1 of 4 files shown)


bob-apps/bin/bobd.rs
```diff
@@ -191,5 +235,12 @@ fn get_matches<'a>() -> ArgMatches<'a> {
                 .long("port")
                 .takes_value(true),
         )
+        .arg(
+            Arg::with_name("init folders")
```
ikopylov, line 239: It is better to use the same name as specified in `long` section (`init_folders`)

bob-apps/bin/bobd.rs
```diff
@@ -191,5 +235,12 @@ fn get_matches<'a>() -> ArgMatches<'a> {
...
+                .help("Initializes bob and alien folders")
+                .short("i")
```
ikopylov, line 241: This is a special one-time flag, so I don't see any benefit in using a short name. It's better to use only the long name to improve command readability

bob-apps/bin/bobd.rs
```diff
@@ -144,6 +148,51 @@ fn spawn_signal_handler(
...
+            let bob_path = bob_path.as_path();
+            if !bob_path.is_dir() {
+                if init_flag {
+                    create_dir(bob_path).expect("Failed to create bob folder");
+                } else {
+                    if let Some(path_str) = bob_path.to_str() {
+                        error!("{} folder doesn't exist, try to use --init_folders flag", path_str);
+                        panic!("{} folder doesn't exist, try to use --init_folders flag", path_str);
+                    } else {
+                        error!("bob folder doesn't exist, try to use --init_folders flag");
```
ikopylov, line 170: Under what circumstances can we enter the `else` path?

bob-apps/bin/bobd.rs
```diff
@@ -144,6 +148,51 @@ fn spawn_signal_handler(
...
+            let alien_path = alien_path.as_path();
+            if !alien_path.is_dir() {
+                if init_flag {
+                    create_dir(alien_path).expect("Failed to create alien folder");
+                } else {
+                    if let Some(path_str) = alien_path.to_str() {
+                        error!("{} folder doesn't exist, try to use --init_folders flag", path_str);
+                        panic!("{} folder doesn't exist, try to use --init_folders flag", path_str);
+                    } else {
+                        error!("alien folder doesn't exist, try to use --init_folders flag");
```
ikopylov, line 187: Same question

bob-apps/bin/bobd.rs
Boneyan: We enter 'else' branch if path to root dir can't be turned into string

ikopylov: So the next obvious question: when can the path to root dir not be converted to string? This seems to be an impossible situation

Boneyan: Seems like it. So I'll remove 'if'

ikopylov (approved):
LGTM

PR #419, opened 2022-01-28
402 add grpc exists to bobp

### LINKED ISSUE #402: Add a mode to `bobp` that checks record existence
It should use `exists` method from gRPC protocol

Boneyan:
bobp now can test exist request (added new argument -s for packet size)
Closes #402


bob-apps/bin/bobp.rs
```diff
@@ -579,9 +651,69 @@ async fn put_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statis
     }
 }
 
+fn prepare_exist_keys(from: u64, count: u64, packet_size: u64, shuffle: bool) -> Vec<Vec<u64>> {
```
idruzhitskiy, line 654: I think it can be moved to `TaskConfig`, as all it uses comes from this struct

bob-apps/bin/bobp.rs
```diff
@@ -384,24 +416,39 @@ async fn print_errors_with_codes(stat: Arc<Statistics>) {
             println!("{:?} = {}", Code::from(code), count);
         }
     }
+    let guard = stat.exist_errors.lock().await;
+    if !guard.is_empty() {
+        println!("put errors:");
```
ikopylov, line 421: `exists errors`

bob-apps/bin/bobp.rs
```diff
@@ -438,46 +491,65 @@ fn print_periodic_stat(
...
             get_speed_values.push(get_spd);
         }
+        if exist_spd > 0.0 {
+            exist_speed_values.push(exist_spd);
+        }
         println!(
-            "put: {:>6} rps  | get {:>6} rps   | put err: {:5}     | get err: {:5}\r\n\
-            put: {:>6.2} kb/s | get: {:>6.2} kb/s | put lat: {:>6.2} ms | get lat: {:>6.2} ms",
+            "put: {:>6} rps  | get {:>6} rps   | exist {:>6} rps   | put err: {:5}     | get err: {:5}     | exist err: {:5}\r\n\
+            put: {:>6.2} kb/s | get: {:>6.2} kb/s | exist: {:>6.2} kb/s | put lat: {:>6.2} ms | get lat: {:>6.2} ms | exist lat: {:>6.2} ms",
```
ikopylov, line 535: The line becomes too long. I suggest to print information only about operations that was actually executed

bob-apps/bin/bobp.rs
```diff
@@ -384,24 +416,39 @@ async fn print_errors_with_codes(stat: Arc<Statistics>) {
...
     println!(
         "avg total: {} rps | total err: {}\r\n\
-        put: {:>6.2} kb/s | get: {:>6.2} kb/s\r\n\
-        put resp time, ms: {:>6.2} | get resp time, ms: {:>6.2}",
-        ((stat.put_total.load(Ordering::Relaxed) + stat.get_total.load(Ordering::Relaxed)) * 1000)
-            .checked_div(elapsed.as_millis() as u64)
-            .unwrap_or_default(),
-        stat.put_error_count.load(Ordering::Relaxed) + stat.get_error_count.load(Ordering::Relaxed),
+        put: {:>6.2} kb/s | get: {:>6.2} kb/s | exist: {:>6.2} kb/s\r\n\
+        put resp time, ms: {:>6.2} | get resp time, ms: {:>6.2} | exist resp time, ms: {:>6.2}",
```
ikopylov, line 446: Same here: it is better to print information only about operations that was actually executed

bob-apps/bin/bobp.rs
```diff
@@ -579,9 +651,69 @@ async fn put_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statis
     }
 }
 
+fn prepare_exist_keys(from: u64, count: u64, packet_size: u64, shuffle: bool) -> Vec<Vec<u64>> {
+    let mut keys = Vec::new();
+    let mut current_low = from;
+    for _ in 0..count {
```
ikopylov, line 657: `count` means a total count of keys (not the count of packages)

bob-apps/bin/bobp.rs
```diff
@@ -579,9 +651,69 @@ async fn put_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statis
...
+    }
+    keys
+}
+
+async fn exist_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statistics>) {
+    let mut client = net_conf.build_client().await;
+
+    let options = task_conf.find_get_options();
+    let send_size_bytes = task_conf.packet_size * (task_conf.key_size as u64);
+    let request_count = task_conf.count / task_conf.packet_size;
```
ikopylov, line 676: This operation will round the number of keys up to a multiple of the package size, which is incorrect

bob-apps/bin/bobp.rs
```diff
@@ -579,9 +651,69 @@ async fn put_worker(net_conf: NetConfig, task_conf: TaskConfig, stat: Arc<Statis
...
+            stat.save_single_thread_exist_time(&start.elapsed());
+            res
+        } else {
+            client.exist(request).await
+        };
+        stat.exist_size_bytes
+                    .fetch_add(send_size_bytes, Ordering::SeqCst);
+        match res {
+            Err(status) => stat.save_exist_error(status).await,
+            Ok(payload) => {
```
ikopylov, line 703: There should be an additional counter, that counts the number of presented keys out of the total requested keys (it should also be printed to console)

bob-apps/bin/bobp.rs
```diff
@@ -247,12 +308,14 @@ struct BenchmarkConfig {
     statistics: Arc<Statistics>,
     time: Option<Duration>,
     request_amount_bytes: u64,
+    request_count: u64,
```
ikopylov, line 311: Looks like `keys_count` is better name for this field

bob-apps/bin/bobp.rs
```diff
@@ -438,46 +529,70 @@ fn print_periodic_stat(
...
-
-        let put_error = stat.put_error_count.load(Ordering::Relaxed);
-        let get_error = stat.get_error_count.load(Ordering::Relaxed);
-        let put_spd = put_count_spd as f64 * k;
-        let get_spd = get_size.get_diff() as f64 / 1024.0;
-        if put_spd > 0.0 {
+        let d_put = put_count.get_diff();
+        let d_get = get_count.get_diff();
+        let d_exist = exist_count.get_diff();
+        if d_put > 0 {
```
ikopylov, line 549: Do I understand correctly that `d_put` is the number of `put` operations performed between two consecutive calls to `print_periodic_stat`? If yes, then this is incorrect condition, because if the performance of `put` operations is low, there can be 0 put requests executed in the time interval, but the statisitcs still should be printed.
It's better to check the `behaviour` parameter to see what to print

bob-apps/bin/bobp.rs
```diff
@@ -438,46 +535,73 @@ fn print_periodic_stat(
...
+            let exist_spd = exist_size.get_diff() as f64 / 1024.0 / sec;
+            println!("exist: {:>6} rps | err {:5} | {:>6.2} kb/s | lat {:>6.2} ms", 
+                exist_count_spd,
+                exist_error,
+                exist_spd,
+                finite_or_default(cur_st_exist_time / cur_st_exist_count / 1e9));
+
+            exist_speed_values.push(exist_spd);
+        }
+        println!("-------------");
```
ikopylov, line 600: This separator is redundant. Especially if all output information is placed in one line

Boneyan: In the current state output isn't placed in one line, but exist, get, put are printed in separate lines

bob-apps/bin/bobp.rs
```diff
@@ -787,6 +984,14 @@ fn get_matches() -> ArgMatches<'static> {
                 .short("k")
                 .default_value(option_env!("BOB_KEY_SIZE").unwrap_or("8")),
         )
+        .arg(
+            Arg::with_name("packet_size")
+                .help("size of packet in exist request")
+                .takes_value(true)
+                .long("size_p")
```
ikopylov, line 991: `packet_size` is a better name than `size_p` from readability perspective

bob-apps/bin/bobp.rs
ikopylov: In practice, only `put`, `get` and `exist` modes are used. 
`test` mode is rarely used and it is the only mode, in which multiple lines are printed.
This separator only creates visual noise in the output, so I suggest to remove it. If you want to separate output from different ticks, then you can add, for example, a number of seconds passed from the test start and print it in the beggining of each line. In groups you can print it only on the first line, and for remaning lines add a spaces to align output

bob-apps/bin/bobp.rs
```diff
@@ -550,8 +550,11 @@ fn print_periodic_stat(
     let print_put = (behavior_flags & PUT_FLAG) > 0;
     let print_get = (behavior_flags & GET_FLAG) > 0;
     let print_exist = (behavior_flags & EXIST_FLAG) > 0;
+    let multiline = (u8::from(print_exist) + u8::from(print_put) + u8::from(print_get)) > 1;
     while !stop_token.load(Ordering::Relaxed) {
         thread::sleep(pause);
+        let elapsed = start.elapsed().as_millis() as f64 / 1000.;
```
ikopylov, line 556: `as_secs()` is enough. Fraction of seconds is not needed, because pause interval is 1 second

bob-apps/bin/bobp.rs
```diff
@@ -550,8 +550,11 @@ fn print_periodic_stat(
...
+        print!("{:>8.1} ", elapsed);
```
ikopylov, line 557: 5 digits is more than enough

bob-apps/bin/bobp.rs
```diff
@@ -583,6 +590,10 @@ fn print_periodic_stat(
             get_speed_values.push(get_spd);
         }
         if print_exist {
+            if multiline {
```
ikopylov, line 593: Inconsistent prefix processing. This will not work with `put` + `exist` combination. I suggest to add a boolean flag, that keeps track of whether the first line was printed. With that flag you can check it at the beggining of every block (`print_put`, `print_get` and `print_exist`) and print the prefix if needed

ikopylov (approved):
LGTM

PR #422, opened 2022-01-29
420 error logging reduce

### LINKED ISSUE #420: Reduce logging intensity during disk disconnection
Currently, for every PUT request for that disk the following log with ERROR level is written:
`2022-01-28 19:20:56 bob_backend::core:245 ERROR PUT[D7A6440000000000][disk8] local failed: Error { ctx: Failed("Error {\n ctx: DCIsNotAvailable,\n}") }`

This message should be written on DEBUG level.
And at the ERROR level, the message should accumulate the number of errors for each disk and write it every 5 seconds. Something like:
`2022-01-28 19:20:56 bob_backend::core:245 ERROR PUT failed 100500 times on disk8.  Error { ctx: Failed("Error {\n ctx: DCIsNotAvailable,\n}") }`

Boneyan:
Same put error logs now agregate in one line every error_log_interval (in case of disk disconnection)
Closes #420


bob-common/src/configs/node.rs
```diff
@@ -481,6 +481,8 @@ pub struct Node {
     #[serde(skip)]
     disks_ref: Arc<Mutex<Vec<DiskPath>>>,
 
+    #[serde(default = "Node::default_error_log_interval")]
+    error_log_interval: String,
```
ikopylov, line 485: This parameter is redundant. It value will never change. It is better to remove the part related to the config and just declare a constant at the place of use

bob-backend/src/core.rs
```diff
@@ -149,7 +216,12 @@ impl Backend {
                 Arc::new(pearl)
             }
         };
-        Self { inner, mapper }
+
+        let error_logger = 
+            Arc::new(Mutex::new(BackendErrorLogger::new(config.error_log_interval())));
+        BackendErrorLogger::spawn_task(error_logger.clone());
```
ikopylov, line 222: Too much complexity for simple logging. This task can be done without additional thread by measuring time intervals. Example on C#: https://github.com/qoollo/dotNet-turbo/blob/master/src/Qoollo.Turbo/Common/EventTimingTracker.cs

bob-backend/src/core.rs
```diff
@@ -131,10 +134,74 @@ pub trait MetricsProducer: Send + Sync {
...
+            errors: HashMap::new(),
+            interval,
+        }
+    }
+
+    fn report_error(&mut self, action: BackendErrorAction) {
+        if let Some(count) = self.errors.get_mut(&action) {
+            *count += 1;
+        } else {
+            self.errors.insert(action, 1);
```
ikopylov, line 160: This is not thread safe. `task` function can access `errors` while this `insert` in progress

bob-backend/src/core.rs
ikopylov: For fast time measurement use this crate: https://docs.rs/coarsetime/latest/coarsetime/

bob-backend/src/core.rs
```diff
@@ -1,4 +1,7 @@
 use crate::prelude::*;
+use std::collections::HashMap;
+use std::sync::Mutex;
+use coarsetime::{Duration as CDuration,Instant};
```
ikopylov, line 4: Please, import `Instant` also with `C` prefix to see places where `coarsetime` is used

bob-backend/src/core.rs
```diff
@@ -131,10 +134,66 @@ pub trait MetricsProducer: Send + Sync {
...
 
+const ERROR_LOG_INTERVAL: u64 = 5000;
+
+#[derive(Hash, PartialEq, Eq, Debug)]
+enum BackendErrorAction {
+    PUT(String, String),
+}
+
+#[derive(Debug)]
+struct BackendErrorLogger {
```
ikopylov, line 145: I suggest to make this type a little bit more universal. Error type can be a generic parameter, allowing the type to be used in other places. Also it would nice to rename the struct to something more universal (e.g. `IntervalErrorLogger`). Additionally, this struct can be moved to a separate file

bob-backend/src/core.rs
```diff
@@ -131,10 +134,66 @@ pub trait MetricsProducer: Send + Sync {
...
+    errors: HashMap<BackendErrorAction, u64>,
+    interval: CDuration,
+    last_timestamp: CInstant,
+}
+
+impl BackendErrorLogger {
+    fn new() -> BackendErrorLogger {
```
ikopylov, line 152: `interval` should be passed as parameter

bob-backend/src/core.rs
```diff
@@ -131,10 +134,66 @@ pub trait MetricsProducer: Send + Sync {
...
+        BackendErrorLogger {
+            errors: HashMap::new(),
+            interval: CDuration::from_millis(ERROR_LOG_INTERVAL),
+            last_timestamp: CInstant::now(),
+        }
+    }
+
+    fn report_error(logger: Arc<Mutex<BackendErrorLogger>>, action: BackendErrorAction) {
```
ikopylov, line 160: All methods in this type should work directly with inner fields. `Arc` and `Mutex` are the things, that are created in the place of use when needed

ikopylov: Or you can take another approach: declare inner struct with HashMap and Timestamp and protect it with Mutex inside this struct

ikopylov: One problem: Mutex potentially can be under a high contention. Try to use Mutex from https://crates.io/crates/parking_lot crate. It implements spinning strategy, that should highly improve performance in our scenario

bob-backend/src/interval_logger.rs
```diff
@@ -0,0 +1,42 @@
+use coarsetime::{Duration as CDuration, Instant as CInstant};
+use std::{
+    collections::HashMap,
+    hash::Hash,
+    cmp::Eq,
+    fmt::Display
+};
+
+#[derive(Debug)]
+pub struct IntervalErrorLogger<E> {
```
ikopylov, line 10: Add an additional struct `IntervalErrorLoggerSafe<E>` that wraps `IntervalErrorLogger<E>` protected by Mutex. This structure will hide the synchronization behind the scenes and thus simplify the caller's code.

bob-backend/src/interval_logger.rs
```diff
@@ -0,0 +1,60 @@
...
+    interval: CDuration,
+    last_timestamp: CInstant,
+}
+
+impl<E: Hash + Eq + Display> IntervalErrorLogger<E> {
+    pub fn new(interval_ms: u64) -> Self {
+        Self {
+            errors: HashMap::new(),
+            interval: CDuration::from_millis(interval_ms),
+            last_timestamp: CInstant::now(),
```
ikopylov, line 22: I suggest subtracting the `interval` to ensure that the very first message is always logged

bob-backend/src/interval_logger.rs
```diff
@@ -0,0 +1,61 @@
...
+    last_timestamp: CInstant,
+}
+
+impl<E: Hash + Eq + Display> IntervalErrorLogger<E> {
+    pub fn new(interval_ms: u64) -> Self {
+        let interval = CDuration::from_millis(interval_ms);
+        Self {
+            errors: HashMap::new(),
+            interval,
+            last_timestamp: CInstant::now() - interval,
```
ikopylov, line 23: I was thinking that it would be nice to additionally subtract 1ms

ikopylov (approved):
LGTM

PR #433, opened 2022-02-21
BRT mode to get information about blobs and indexes

### LINKED ISSUE #356: `brt`: add command to display information about blob and index file
It would be nice to print the content of the BLOB file header with a special command. This will help diagnose problems. The same can be done for the index file

vovac12:
Closes #356


bob-apps/bin/blob_recovery_tool/command.rs
```diff
@@ -401,6 +405,91 @@ impl MigrateCommand {
     }
 }
 
+pub struct GetBlobInfoCommand {
+    path: PathBuf,
+}
+
+impl GetBlobInfoCommand {
```
ikopylov, line 412: This subcommand should have optional parameter `--full` to specify, that full scan of blob is required. In short mode (default one) it should print only information from BLOB header

bob-apps/Cargo.toml
```diff
@@ -52,6 +53,8 @@ features = ["signal"]
 
 [dependencies.pearl]
 version = "0.13.0"
+git = "https://github.com/qoollo/pearl"
```
ikopylov, line 56: This links should be updated, when new version of Pearl will be released

bob-apps/bin/blob_recovery_tool/command.rs
```diff
@@ -401,6 +406,105 @@ impl MigrateCommand {
...
+    path: PathBuf,
+    full: bool,
+}
+
+impl GetBlobInfoCommand {
+    fn run(&self) -> AnyResult<()> {
+        let collector = pearl::tools::BlobSummaryCollector::from_path(&self.path, self.full)?;
+        println!("Blob summary");
+        println!("Path: {:?}", self.path);
+        if self.full {
```
ikopylov, line 419: Move the printing of full info after printing the header info

ikopylov (approved):
LGTM

PR #436, opened 2022-02-23
408 user auth claims

### LINKED ISSUE #408: Authorization: add an ability to specify claims directly on user
We implement a claim-based policy in which role is just a group of claims. So the role on user should be optional, and an additional `claims` section should be supported to allow claims to be specified. If both role and claims are specified, then the role is used as the primary claim source, and the values in the `claims` section should override them.

Boneyan:
User's role is now optional. Claims can be set in user description, claims are also optional. If both claims and role are set, then claims will overwrite role's permissions
Closes #408


CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ Bob versions changelog
 
 ## [Unreleased]
 #### Added
-
+- Access management (#217)
```
ikopylov, line 7: Because #217 is already merged, it is better to remove this line from changelog

CHANGELOG.md
```diff
@@ -22,7 +23,7 @@ Bob versions changelog
 - Access management (#217)
   - Nodes authentication (#318)
   - Authorization parameters in bobp (#425)
-  - Http API authentication (#217)
+  - Http API authentication. (#217)
```
ikopylov, line 26: Please, revert this change

PR #443, opened 2022-03-12
304 password hash ability

### LINKED ISSUE #304: Authorization: add the ability to specify a password hash instead of bare password

Boneyan:
SHA512 hash of password with salt 'bob' can be specified instead of password
Closes #304


CHANGELOG.md
```diff
@@ -4,9 +4,15 @@ Bob versions changelog
 
 ## [Unreleased]
 #### Added
+- Access management
+  - Nodes authentication. (#318)
+  - Http API authentication.
+  - User's permissions can be set via role and/or claims (#408)
+  - SHA512 hash of password with salt 'bob' can be specified instead of password (#304)
```
ikopylov, line 11: Only this line should be left, and the rest should be removed

bob-access/src/authenticator/basic.rs
```diff
@@ -60,8 +62,22 @@ impl<Storage: UsersStorage> Basic<Storage> {
...
-                if user.password() == password {
-                    Ok(user.into())
+                if let Some(usr_password) = user.password() {
+                    if usr_password == password {
+                        Ok(user.into())
+                    } else {
+                        Err(Error::UnauthorizedRequest)
+                    }
+                } else if let Some(usr_hash) = user.hash() {
+                    let hash_str = format!("{}bob", password);
```
ikopylov, line 72: I think, we can make salt configurable by adding optional root parameter to `users.yaml`

bob-access/src/authenticator/basic.rs
```diff
@@ -60,8 +62,22 @@ impl<Storage: UsersStorage> Basic<Storage> {
...
+                    let mut hasher = Sha512::new();
+                    hasher.update(&hash_str.into_bytes());
+                    let hash = hasher.finalize();
```
ikopylov, line 75: Because in practice there will be a small number of users, it makes sense to store a limited number of hashes inside local cache (small HashMap, which size is limited to 32, without displacement)

ikopylov:
Please, recreate this request from scratch to remove irrelevant commits from history (i.e. create new branch, port all changes and create new PR)

PR #449, opened 2022-03-17
447 migration not updating version header

### LINKED ISSUE #447: `brt`: BLOB migration does not update version number in header

### LINKED ISSUE #448: `brt`: incorrect default target version for BLOB migration command
In the help the default target version is `2`, which is incorrect. The correct value is `1`

Boneyan:
Version in BlobHeader now changes during migration
Closes #447
Closes #448

### DIFF
--- a/bob-apps/bin/blob_recovery_tool/structs.rs
+++ b/bob-apps/bin/blob_recovery_tool/structs.rs
@@ -82,8 +82,8 @@ impl Validatable for BlobHeader {
 }
 
 impl BlobHeader {
-    pub(crate) fn migrate(mut self, target: u32) -> AnyResult<Self> {
-        self.version = target;
+    pub(crate) fn migrate(mut self, version: u32) -> AnyResult<Self> {
+        self.version = version;
         Ok(self)
     }
 }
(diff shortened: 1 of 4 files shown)


bob-apps/bin/blob_recovery_tool/structs.rs
```diff
@@ -81,6 +81,13 @@ impl Validatable for BlobHeader {
     }
 }
 
+impl BlobHeader {
+    pub(crate) fn migrate(mut self, target: u32) -> AnyResult<Self> {
```
idruzhitskiy, line 85: Why not `version` as argument?

bob-apps/bin/blob_recovery_tool/structs.rs
```diff
@@ -81,6 +81,13 @@ impl Validatable for BlobHeader {
     }
 }
 
+impl BlobHeader {
+    pub(crate) fn migrate(mut self, version: u32) -> AnyResult<Self> {
```
ikopylov, line 85: The signature should be the same as in `Record.migrate`. It should receive source_version and target_version as parameters. Also the behavior should be the same: you need to explicitly restrict possible migrations

ikopylov:
LGTM

PR #451, opened 2022-03-22
438 send memory fix

### LINKED ISSUE #438: High memory consumption inside docker
Strange behavior that only reproduces inside docker: Bob consumes too much memory and is killed by OutOfMemory. This becomes more obvious, when size of the record is large (e.g. 75 Mb). According to metrics the consumption is not related with Bloom filters or indices (number of BLOBs is too small for that). 

One assumption worth checking out is that the consumption could be due to disk cache not being flushed to the HDD. Docker could potentially treat this memory as belonging to the process itself. See this link: https://codefresh.io/docker-tutorial/docker-memory-usage/

idruzhitskiy:
Closes #438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-- Fix ping leading to sending too much requests (#438)
+- Fix ping and timeout leading to sending too much requests (#438)
 
 
 #### Updated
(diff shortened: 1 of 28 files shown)


ikopylov (approved):
LGTM

ikopylov:
Please, backport this entire PR to `release_v2.0.0`

PR #460, opened 2022-04-10
459 exact number of vdisks default

### LINKED ISSUE #459: CCG parameterization improvement
Currently, the number of vdisks is considered to be the least common multiple of the number of replicas and physical disks. If the `-exact` flag is set, then the specified number of vdisks is taken. This doesn't follow common usage patterns.

LCM should be removed and the number of vdisks passed by `-d` should always be used as the exact number of vdisks. Thus `-d` becomes a required parameter. In this way the `-exact` flag should be removed (because it becomes a default behaviour).

An alternative way to specify the number of vDisks must also be implemented, allowing you to configure the number of vDisks per physical disk. Formula to calculate precise number of vDisks will be: `vDisksCount = ceil((vDisksPerPhysicalDisk * physicalDiskCount) / replicasCount)`. `-d` and the new parameter should be mutually exclusive (if both specified then error should be displayed)

Boneyan:
Number of vdisks in ccg now is defined by -p or -d arg only, -exact arg removed.
Number of vdisks per disk in ccg can be specified via -p arg.
Closes #459


bob-apps/bin/config_cluster_generator/utils.rs
```diff
@@ -83,21 +83,10 @@ fn deserialize(content: String) -> AnyResult<ClusterConfig> {
...
 }
 
-/// greatest common divider
-pub fn gcd(a: usize, b: usize) -> usize {
-    debug!("gcd of {} and {}", a, b);
-    if a == 0 {
-        b
-    } else if b == 0 {
-        a
+pub fn ceil(a: usize, b: usize) -> usize {
```
ikopylov, line 86: https://doc.rust-lang.org/std/primitive.f64.html#method.ceil

Boneyan: Ceil for usize is available only on nightly https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil. Or should I covert usize to f64, calculate ceil and then convert result back to usize?

bob-apps/bin/ccg.rs
```diff
@@ -130,14 +130,15 @@ fn simple_gen(
...
-    vdisks_counts_match: bool,
+    vdisks_per_disk: bool,
     use_racks: bool,
 ) -> AnyResult<ClusterConfig> {
     let mut center = Center::from_cluster_config(&config, use_racks)?;
     center.validate()?;
-    if !vdisks_counts_match {
-        vdisks_count = vdisks_count.max(lcm(center.disks_count(), replicas_count));
+    if vdisks_per_disk {
+        vdisks_count = ceil(vdisks_count * center.disks_count(), replicas_count);
```
ikopylov, line 139: You should multiply by `vdisks_per_disk`, not by `vdisks_count`

bob-apps/bin/ccg.rs
```diff
@@ -180,23 +181,21 @@ fn get_replicas_count(matches: &ArgMatches) -> AnyResult<usize> {
...
-            s.parse()
-                .map_err(|err| anyhow!("get vdisks count: {}", err))
-        },
-    )
+fn get_vdisks_count(matches: &ArgMatches) -> AnyResult<usize> {
+    if let Some(s) = matches.value_of("vdisks_count") {
+        s.parse()
+            .map_err(|err| anyhow!("get vdisks count: {}", err))
+    } else if let Some(s) = matches.value_of("vdisks_per_disk") {
+        s.parse()
```
ikopylov, line 189: This conflicts with the function name because it returns the number of virtual disks per disk (not the number of virtual disks)

bob-apps/bin/config_cluster_generator/utils.rs
ikopylov: Didn't notice that you are working with `usize` only. The you can keep this code

ikopylov (approved):
LGTM

PR #465, opened 2022-04-14
364 clusterwide record delete

### LINKED ISSUE #364: Clusterwide single record deletion
Currently, the deletion is triggered through the REST API on a single node (since the REST API does not redirect commands to the entire cluster). It needs to be updated, so that the deletion is performed on the whole cluster

idruzhitskiy:
Closes #364


bob-common/src/bob_client.rs
```diff
@@ -172,6 +172,27 @@ pub mod b_client {
...
+            };
+            let mut req = Request::new(message);
+            self.set_credentials(&mut req);
+            self.set_timeout(&mut req);
+            self.metrics.delete_timer_stop(timer);
+            if client.delete(req).await.is_ok() {
+                Ok(NodeOutput::new(node_name, ()))
+            } else {
+                self.metrics.delete_error_count();
+                Err(NodeOutput::new(node_name, Error::timeout()))
```
ikopylov, line 192: Why `Error::timeout()` is used here? In the functions above the error returned from call is used (`Err(NodeOutput::new(node_name, e.into()))`)

bob-common/src/bob_client.rs
```diff
@@ -172,6 +172,27 @@ pub mod b_client {
...
+            self.metrics.delete_count();
+            let timer = BobClientMetrics::start_timer();
+            let message = DeleteRequest {
+                key: Some(BlobKey { key: key.into() }),
+                options: Some(options),
+            };
+            let mut req = Request::new(message);
+            self.set_credentials(&mut req);
+            self.set_timeout(&mut req);
+            self.metrics.delete_timer_stop(timer);
```
ikopylov, line 187: The timer should stop after the `delete` call is finished

bob/src/cluster/operations.rs
```diff
@@ -74,10 +86,30 @@ pub(crate) async fn put_at_least(
...
+    at_least: usize,
+    options: DeleteOptions,
+) -> (Tasks, Vec<NodeOutput<Error>>) {
+    call_at_least(target_nodes, at_least, |n| {
+        call_node_delete(key, options.clone(), n)
+    })
+    .await
+}
+
+pub(crate) async fn call_at_least(
```
ikopylov, line 107: Should it be `pub(crate)`?

bob/src/cluster/operations.rs
```diff
@@ -74,10 +86,30 @@ pub(crate) async fn put_at_least(
...
-        .cloned()
-        .map(|node| call_node_put(key, data.clone(), node, options.clone()))
-        .collect();
+    call_at_least(target_nodes, at_least, |n| {
+        call_node_put(key, data.clone(), n, options.clone())
+    })
+    .await
+}
+
+pub(crate) async fn delete_at_least(
```
ikopylov, line 95: `delete` operation is a little bit special. It should operate on every node and success only if the `delete` on every node was successful. Thus `at_least` parameter is not needed, nor is `at_least` in the function name

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,50 @@ impl Quorum {
         }
     }
 
+    async fn delete_at_least(&self, key: BobKey, with_aliens: bool) -> Result<(), Error> {
+        debug!("DELETE[{}] ~~~DELETE LOCAL NODE FIRST~~~", key);
+        let mut local_put_ok = 0_usize;
+        let mut remote_ok_count = 0_usize;
+        let mut at_least = self.quorum;
```
ikopylov, line 85: `delete` should be performed on all node, containing data with specified `key`. And if `with_aliens` is true, then on all node in cluster

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,50 @@ impl Quorum {
...
+        let (tasks, errors) = self
+            .delete_at_remote_nodes(key, with_aliens, at_least)
+            .await;
+        let all_count = self.mapper.get_target_nodes_for_key(key).len();
+        remote_ok_count += all_count - errors.len() - tasks.len() - local_put_ok;
+        failed_nodes.extend(errors.iter().map(|e| e.node_name().to_string()));
+        if remote_ok_count + local_put_ok >= self.quorum {
+            debug!("DELETE[{}] spawn {} background put tasks", key, tasks.len());
+            let q = self.clone();
+            tokio::spawn(q.background_delete(tasks, key, failed_nodes));
```
ikopylov, line 111: `delete` can be considered successful only if it successful on every node, thus running this operation in background is incorrect

bob/src/cluster/simple.rs
```diff
@@ -115,4 +132,11 @@ impl Cluster for Quorum {
         }
         Ok(exist)
     }
+
+    async fn delete(&self, key: BobKey, with_aliens: bool) -> Result<(), Error> {
+        self.perform_on_nodes(key, "DELETE", move |c| {
```
ikopylov, line 137: If `with_aliens` is true, then operation should be performed on all nodes in cluster (not just on nodes, containing specified key)

ikopylov:
Please, resolve merge conflicts

bob/src/cluster/operations.rs
idruzhitskiy: Since it'll use `call_at_least` internally, I'll leave `at_least` suffix here and pass count of target nodes on the call site

bob-common/src/metrics/mod.rs
```diff
@@ -168,6 +175,21 @@ impl BobClient {
             timer.elapsed().as_nanos() as f64
         );
     }
+
+    pub(crate) fn delete_count(&self) {
+        counter!(self.prefix.clone() + ".delete_count", 1);
```
ikopylov, line 180: This string concatenation executes for every request, but it is obviously that every time it gives the same result. I suggest to precalculate all strings in constructor. Probably, there should be a separate struct with counter names, stored inside `BobClient` and wrapped with `Box`

bob-grpc/proto/bob.proto
```diff
@@ -79,3 +86,9 @@ message BobError {
     int32 code = 1; // Error code in case of error
     string desc = 2; // Error desription
 }
+
+// Delete operation options
+message DeleteOptions {
+    bool force_node = 1; // Force operation to be served by node to which it comes
+    bool with_aliens = 2; // Remove data in aliens as well
```
ikopylov, line 93: I think, that we should always perform delete operation on aliens or at least inverse that flag, so that deleting in aliens will be enabled by default

bob/src/cluster/operations.rs
```diff
@@ -30,6 +30,18 @@ fn call_node_put(
...
 }
 
+fn call_node_delete(
+    key: BobKey,
+    options: DeleteOptions,
+    node: Node,
+) -> JoinHandle<Result<NodeOutput<()>, NodeOutput<Error>>> {
+    debug!("DELETE[{}] delete to {}", key, node.name());
+    let task = async move {
+        LinkManager::call_node(&node, |conn| conn.delete(key, options).boxed()).await
```
ikopylov, line 40: Tests build failed: `no method named 'delete' found for reference '&MockBobClient' in the current scope`. Looks like `delete` should be added to `MockBobClient`

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,38 @@ impl Quorum {
         }
     }
 
+    async fn delete_on_nodes(&self, key: BobKey, with_aliens: bool) -> Result<(), Error> {
+        debug!("DELETE[{}] ~~~DELETE LOCAL NODE FIRST~~~", key);
+        let mut local_put_ok = 0_usize;
```
ikopylov, line 83: Name copy-paste

bob/src/cluster/quorum.rs
```diff
@@ -124,6 +157,39 @@ impl Quorum {
...
+            key,
+            target_nodes.len(),
+        );
+        let count = target_nodes
+            .iter()
+            .filter(|n| n.name() != local_node)
+            .count();
+        let target_nodes = target_nodes
+            .into_iter()
+            .filter(|node| node.name() != local_node);
```
ikopylov, line 183: Probably it is better to remove one element from the vector than to repeat the filtering 2 times

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,38 @@ impl Quorum {
         }
     }
 
+    async fn delete_on_nodes(&self, key: BobKey, with_aliens: bool) -> Result<(), Error> {
+        debug!("DELETE[{}] ~~~DELETE LOCAL NODE FIRST~~~", key);
+        let mut local_put_ok = 0_usize;
+        let mut failed_nodes = Vec::new();
+        let res = delete_at_local_node(&self.backend, key, with_aliens).await;
```
ikopylov, line 85: You should check, that it is required to delete from local node before doing so

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,38 @@ impl Quorum {
...
+            error!("{}", e);
+            failed_nodes.push(self.mapper.local_node_name().to_owned());
+        } else {
+            local_put_ok += 1;
+            debug!("DELETE[{}] local node delete successful", key);
+        }
+
+        debug!("DELETE[{}] ~~~DELETE TO REMOTE NODES~~~", key);
+        let errors = self.delete_at_remote_nodes(key, with_aliens).await;
+        let all_count = self.mapper.get_target_nodes_for_key(key).len();
```
ikopylov, line 96: Total number of nodes depends on `with_aliens` flag. I suggest to return number of affected nodes from `delete_at_remote_nodes` inside tuple. In this case `all_count` can be calculated as `affected_nodes + if delete_on_local_node {1} else {0}`

bob/src/cluster/simple.rs
```diff
@@ -38,34 +38,33 @@ impl Quorum {
         }
         keys_by_nodes
     }
-}
 
-#[async_trait]
-impl Cluster for Quorum {
```
ikopylov, line 44: I just noticed, that types in `quorum.rs` and `simple.rs` have same name. How does it compile without errors?

bob/src/cluster/simple.rs
ikopylov: Looks like this was not resolved

bob/src/grinder.rs
```diff
@@ -209,23 +210,39 @@ impl Grinder {
...
-        );
-        if result.is_err() {
-            counter!(CLIENT_DELETE_ERROR_COUNT_COUNTER, 1);
-        }
-        timing!(CLIENT_DELETE_TIMER, sw.elapsed().as_nanos() as f64);
+        let result = if options.force_node {
+            counter!(CLIENT_DELETE_COUNTER, 1);
+            let sw = Stopwatch::start_new();
+            trace!(
+                "pass request to backend, /{:.3}ms/",
```
ikopylov, line 219: Add `delete` to text
```suggestion
                "pass delete request to backend, /{:.3}ms/",
```

bob/src/cluster/simple.rs
idruzhitskiy: Different modules, Quroum is renamed as SimpleQuorum

bob/src/cluster/quorum.rs
idruzhitskiy: Are you sure it's needed here, and not on lower levels?

ikopylov: This is needed to calculate `all_count` properly whithout call to `get_target_nodes_for_key` on this level.
Additionally, the function name by itself says, that it is supposed to work only for the local node. 
You could probably return an extra flag that tells whether it is required to run delete locally, but that would also require you to rename the function

bob/src/cluster/simple.rs
idruzhitskiy: True, for simple quorum it wasn't. Fixed

bob-common/src/metrics/mod.rs
idruzhitskiy: Although I've removed string concat, we still need to clone string every time (as metrics requires owned strings).

bob-grpc/proto/bob.proto
```diff
@@ -79,3 +86,9 @@ message BobError {
     int32 code = 1; // Error code in case of error
     string desc = 2; // Error desription
 }
+
+// Delete operation options
+message DeleteOptions {
+    bool force_node = 1; // Force operation to be served by node to which it comes
+    bool without_aliens = 2; // Do not remove data in aliens
```
ikopylov, line 93: This flag will become meaningless after issue #533 will be implemented, because working through aliens will become the default behavior (just like `put` request, which does not have such flag). Probably, it is better to remove it in this PR?

bob/src/cluster/quorum.rs
```diff
@@ -124,6 +154,37 @@ impl Quorum {
...
+        };
+        target_nodes.retain(|n| n.name() != local_node);
+        debug!(
+            "DELETE[{}] cluster quorum put remote nodes {} total target nodes",
+            key,
+            target_nodes.len(),
+        );
+        let count = target_nodes.len();
+        (
+            delete(
```
ikopylov, line 177: It would be nice to expand the name to distinguish it from the `self.delete` function

bob/src/cluster/quorum.rs
```diff
@@ -77,6 +78,35 @@ impl Quorum {
         }
     }
 
+    async fn delete_on_nodes(&self, key: BobKey, without_aliens: bool) -> Result<(), Error> {
+        debug!("DELETE[{}] ~~~DELETE LOCAL NODE FIRST~~~", key);
+        let res = delete_at_local_node(&self.backend, key, without_aliens).await;
```
ikopylov, line 83: When `without_aliens` is true, we should check, that the current node contains target vdisk for key. Otherwise the error will be returned (https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/core.rs#L295).
But probably we should remove  `without_aliens` flag at all.

bob/src/cluster/quorum.rs
ikopylov: When `without_aliens` is true, we should check, that the current node contains target vdisk for key. Otherwise the error will be returned (https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/core.rs#L295).
But probably we should remove  `without_aliens` flag at all.

bob/src/cluster/simple.rs
ikopylov: I don't see, that rename done

bob-common/src/metrics/mod.rs
```diff
@@ -111,16 +118,51 @@ pub type Timer = Instant;
...
 impl BobClient {
-    fn new(prefix: String) -> Self {
-        BobClient { prefix }
+    fn new(prefixed_names: Arc<PrefixedNames>) -> Self {
+        BobClient { prefixed_names }
     }
 
     pub(crate) fn put_count(&self) {
-        counter!(self.prefix.clone() + ".put_count", 1);
+        counter!(self.prefixed_names.put_count.clone(), 1);
```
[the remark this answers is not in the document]
idruzhitskiy: It will requre `'static` strings to construct, and they can be either compile-time defined or cloned

bob-grpc/proto/bob.proto
idruzhitskiy: Sure, I'll remove it

bob/src/cluster/simple.rs
idruzhitskiy: Renamed here https://github.com/qoollo/bob/blob/master/bob/src/cluster/mod.rs#L10

bob-common/src/metrics/mod.rs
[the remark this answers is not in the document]
ikopylov: It has conversion from `String`: https://docs.rs/metrics/0.20.1/metrics/struct.KeyName.html#method.from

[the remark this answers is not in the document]
idruzhitskiy: It will require updating metrics crate to `0.18.0` version, which breaks some code (method changes). I think it's better to make metrics update separate ticket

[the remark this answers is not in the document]
ikopylov: Agree with that. Created a separate issue: https://github.com/qoollo/bob/issues/628

ikopylov:
`delete_alien` in `disk_controller.rs` writes `warn` log when group is not found, but if there are no aliens for target `vdisk` the group will not be found and that is correct: https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/disk_controller.rs#L625

I think, that the log level should be changed to `debug` in this particular case

ikopylov:
Please, fix merge conflicts

ikopylov (approved):
LGTM

PR #474, opened 2022-04-24
New metric collector fix

### LINKED ISSUE #461: `iowait` and `rps` disk metrics are not working due to an `iostat` output parsing error
Different versions of `iostat` has slightly different output, thus it is better to get metrics directly from procfs files

Boneyan:
CPU iowait & disks iowait, iops are now collected via procfs
Closes #461

### DIFF
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -195,21 +195,14 @@ impl CPUStatCollector {
     }
 
     fn stat_cpu_line() -> Result<String, String> {
-        if let Ok(mut stat_file) = File::open(CPU_STAT_FILE) {
-            let mut contents = String::new();
-            if stat_file.read_to_string(&mut contents).is_err() {
-                return Err("Failed to read stat file".into());
+        let lines = file_contents(CPU_STAT_FILE)?;
+        for stat_line in lines {
+            let mut parts = stat_line.split_whitespace();
+            if let Some("cpu") = parts.next() {
+                return Ok(stat_line);
             }
-            for stat_line in contents.lines() {
-                let mut parts = stat_line.split_whitespace();
-                if let Some("cpu") = parts.next() {
-                    return Ok(stat_line.into())
-                }
-            }
-            Err("Can't find cpu stat".into())
-        } else {
-            Err("Can't open stat file".into())
         }
+        Err("Can't find cpu stat".into())
     }
 
     fn iowait(&mut self) -> Result<f64, CommandError> {
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -202,6 +202,8 @@ impl CPUStatCollector {
                 let mut lines = output
                     .lines()
                     .skip_while(|line| line.find("avg-cpu:").is_none());
+                // skip headers
+                lines.next();
                 // find iowait column on the next line
                 if let Some(line) = lines.next() {
                     let mut values = line.split_whitespace();
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -12,7 +12,9 @@ use std::io::{Read};
 use sysinfo::{DiskExt, ProcessExt, RefreshKind, System, SystemExt};
 
 const DESCRS_DIR: &str = "/proc/self/fd/";
-const STAT_FILE: &str = "/proc/stat";
+const CPU_STAT_FILE: &str = "/proc/stat";
+const DISK_STAT_FILE: &str = "/proc/diskstats";
+const UPTIME: &str = "/proc/uptime";
 
 pub(crate) struct DiskSpaceMetrics {
     pub(crate) total_space: u64,
@@ -193,7 +195,7 @@ impl CPUStatCollector {
     }
 
     fn stat_cpu_line() -> Result<String, String> {
-        if let Ok(mut stat_file) = File::open(STAT_FILE) {
+        if let Ok(mut stat_file) = File::open(CPU_STAT_FILE) {
             let mut contents = String::new();
             if stat_file.read_to_string(&mut contents).is_err() {
                 return Err("Failed to read stat file".into());
@@ -253,7 +255,7 @@ impl CPUStatCollector {
 }
 
 struct DiskStatCollector {
-    iostat_avl: bool,
+    procfs_avl: bool,
     disk_metric_data: HashMap<String, String>,
 }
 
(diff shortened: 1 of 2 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -309,30 +364,33 @@ impl DiskStatCollector {
...
                 Err(CommandError::Primary(e))
             }
         }
     }
 }
 
+fn file_contents(path: &str) -> Result<Vec<String>, String> {
+    if let Ok(mut file) = File::open(path) {
+        let mut contents = String::new();
+        if file.read_to_string(&mut contents).is_err() {
```
ikopylov, line 385: https://doc.rust-lang.org/std/fs/fn.read_to_string.html

bob/src/hw_metrics_collector.rs
```diff
@@ -179,52 +184,71 @@ enum CommandError {
...
+        }
     }
 
-    fn parse_iostat_result(iostat_str: &str) -> Result<f64, std::num::ParseFloatError> {
-        iostat_str.replace(',', ".").parse()
+    fn stat_cpu_line() -> Result<String, String> {
+        let lines = file_contents(CPU_STAT_FILE)?;
+        for stat_line in lines {
+            let mut parts = stat_line.split_whitespace();
+            if let Some("cpu") = parts.next() {
```
ikopylov, line 201: It is better to use `starts_with` to find the line and only after the line is found, perform string splitting

bob/src/hw_metrics_collector.rs
```diff
@@ -179,52 +184,71 @@ enum CommandError {
...
+                    for i in 1..parts.len() {
+                        match parts[i].parse::<f64>() {
+                            Ok(val) => {
+                                sum += val;
+                                if i == 5 {
+                                    f_iowait = val;
+                                }
+                            },
+                            Err(_) => {
+                                err = Some("Can't parse stat".to_string());
```
ikopylov, line 229: Please, write path to the file in error messages (`Can't parse /proc/stat`)

bob/src/hw_metrics_collector.rs
```diff
@@ -179,52 +184,71 @@ enum CommandError {
...
+                                break;
+                            }
                         }
                     }
+                    if err.is_none() {
+                        return Ok(f_iowait * 100. / sum);
+                    }
+                } else {
+                    err = Some("Stat format changed".into());
```
ikopylov, line 238: Same here

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +266,83 @@ impl DiskStatCollector {
...
             let lsp: Vec<&str> = i.split_whitespace().collect();
-            if lsp.len() == 21 {
-                if let Some(metric_prefix) = self.disk_metric_data.get(lsp[0]) {
-                    if let (Ok(rs), Ok(ws)) = (
-                        // readops per s count is in 1st column
-                        Self::parse_iostat_result(lsp[1]),
-                        // writeops per s count is in 7th column
-                        Self::parse_iostat_result(lsp[7]),
+            if lsp.len() >= 8 {
+                if let Some(metric_prefix) = self.disk_metric_data.get(lsp[2]) {
```
ikopylov, line 280: Please add a comment on what is supposed to be read at index 2

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +266,83 @@ impl DiskStatCollector {
...
+                    if let (Ok(r_ios), Ok(w_ios)) = (
+                        // successfull reads count is in 3rd column
+                        lsp[3].parse::<f64>(),
+                        // successfull writes count is in 7th column
+                        lsp[7].parse::<f64>(),
                     ) {
-                        let iops = rs + ws;
+                        let iops = (r_ios + w_ios) / uptime;
```
ikopylov, line 287: We need the current number of iops, not the average since the system was started. The easiest way is to calculate the delta between this call and the previous one, with an additional check that the elapsed time is greater than 1 second

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +266,83 @@ impl DiskStatCollector {
...
-                        let gauge_name = format!("{}_iowait", metric_prefix);
-                        gauge!(gauge_name, iowait);
+                        if lsp.len() >= 11 {
+                            if let (Ok(w_ticks), Ok(r_ticks)) = (
+                                // time spend on write is in 10th column
+                                lsp[10].parse::<f64>(),
+                                // time spend on read is in 6th column
+                                lsp[6].parse::<f64>(),
+                            ) {
+                                let iowait = r_ticks / r_ios + w_ticks / w_ios;
```
ikopylov, line 299: Same problem here. Also, you should check that `r_ios` and `w_ios` is greater than 0 (otherwise you will get a NaN)

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +266,83 @@ impl DiskStatCollector {
...
+        
+                                let gauge_name = format!("{}_iowait", metric_prefix);
+                                gauge!(gauge_name, iowait);
+                            } else {
+                                return Err(CommandError::Primary("Can't parse diskstat values to float".to_string()));
```
ikopylov, line 304: Please, write full path to the file in the error message

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +184,100 @@ enum CommandError {
     Primary(String),
 }
 struct CPUStatCollector {
-    iostat_avl: bool,
+    procfs_avl: bool,
 }
 
+const CPU_IOWAIT_COLUMN: usize = 5;
```
ikopylov, line 190: This const is only used in one function, so I suggest declaring it directly inside the function body.

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +294,105 @@ impl DiskStatCollector {
...
+            read_time: DiffContainer::new(),
+            write_time: DiffContainer::new(),
         }
     }
 
     fn collect_and_send_metrics(&mut self) -> Result<(), CommandError> {
-        let iostat_lines = self.collect_iostat()?;
-        for i in iostat_lines {
+        let uptime = self.uptime()?;
+        let elapsed = self.uptime.diff(uptime);
```
ikopylov, line 309: We don't need to parse `uptime` file to get elapsed time. It is simpler and faster to use `Instant` and `Duration` types

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +294,105 @@ impl DiskStatCollector {
...
+                    if let (Ok(r_ios), Ok(w_ios)) = (
+                        // successfull reads count is in 3rd column
+                        lsp[3].parse::<f64>(),
+                        // successfull writes count is in 7th column
+                        lsp[7].parse::<f64>(),
                     ) {
-                        let iops = rs + ws;
+                        let d_r_ios = self.reads.diff(r_ios);
+                        let d_w_ios = self.writes.diff(w_ios);
+                        let iops = (d_r_ios + d_w_ios) / elapsed;
```
ikopylov, line 324: If `elapsed` is too small, the approximation will be incorrect. I suggest to introduce threshold (500ms), under which the diff is calculated not from the previous value, but from the pre-previous one. And the previous value is also not updated below this threshold. In this way we can guarantee that approximation will be estimated by the meaningful interval (> 500ms). 

Pseudocode:
```
if elapsed < 500 {
    let d_r_ios = r_ios - self.prev_prev_reads;
    let iops = (d_r_ios + d_w_ios) / elapsed_from_prev_prev;
} 
else {
    let d_r_ios = r_ios - self.prev_reads;
    self.prev_prev_reads = self.prev_reads;
    self.prev_reads = r_ios;
    let iops = (d_r_ios + d_w_ios) / elapsed;
}
```

bob/src/hw_metrics_collector.rs
```diff
@@ -242,52 +294,105 @@ impl DiskStatCollector {
...
-                        gauge!(gauge_name, iowait);
+                        if lsp.len() >= 11 {
+                            if let (Ok(w_ticks), Ok(r_ticks)) = (
+                                // time spend on write is in 10th column
+                                lsp[10].parse::<f64>(),
+                                // time spend on read is in 6th column
+                                lsp[6].parse::<f64>(),
+                            ) {
+                                let d_r_ticks = self.read_time.diff(r_ticks);
+                                let d_w_ticks = self.write_time.diff(w_ticks);
```
ikopylov, line 337: Same idea here

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +183,165 @@ enum CommandError {
...
+    fn tick(&mut self) -> Option<TimeMoment> {
+        let now = Instant::now();
+        if let Some(last) = self.last {
+            let elapsed = now.duration_since(last);
+            if elapsed > DIFFCONTAINER_THRESHOLD {
+                self.update(now);
+                return Some(TimeMoment::Last(elapsed));
+            } else if let Some(prelast) = self.prelast {
+                let elapsed = now.duration_since(prelast);
+                self.update(now);
```
ikopylov, line 294: We don't need to update the time here because it could lead to a situation where the time elapsed since the prelast is less than `DIFFCONTAINER_THRESHOLD`.

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +183,165 @@ enum CommandError {
...
+        match elapsed {
+            TimeMoment::Last(_) => {
+                let diff = new - self.last;
+                self.prelast = self.last;
+                self.last = new;
+                diff
+            },
+            TimeMoment::Prelast(_) => {
+                let diff = new - self.prelast;
+                self.prelast = self.last;
```
ikopylov, line 329: We don't need to update values if the diff calculated from prelast value. This logic is tightly coupled with the update logic in `TimeContainer`, so I suggest to combine them into a single struct

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +183,165 @@ enum CommandError {
...
+            TimeMoment::Prelast(dur) => dur.as_millis() as f64,
+        }
+    }
+}
+
+impl TimeContainer {
+    fn new() -> Self {
+        Self {
+            last: None,
+            prelast: None,
```
ikopylov, line 276: Actually, I just realized that we don't need to keep track of `prelast` values, because we don't need to provide values on every update. If we don't provide values, then metrics library will simply report a previous ones to graphite/prometheus. So the logic can be simplified: we just check, that the elapsed time since the last update is greater than `DIFFCONTAINER_THRESHOLD`. If it is then we publish new values and update the last to the current ones, if it is not then do nothing.

I apologize that I previously suggested to implement more complex logic and that now I ask you to simplify it

bob/src/hw_metrics_collector.rs
```diff
@@ -360,62 +333,78 @@ impl DiskStatCollector {
...
+    }
+
     fn collect_and_send_metrics(&mut self) -> Result<(), CommandError> {
         let diskstats = self.diskstats()?;
-        if let Some(elapsed) = self.time.tick() {
+        let now = Instant::now();
+        let elapsed = now.duration_since(self.upd_timestamp);
+        if elapsed > DIFFCONTAINER_THRESHOLD {
+            self.upd_timestamp = now;
+            let elapsed = elapsed.as_millis() as f64 / 1000.;
```
ikopylov, line 383: There is a function for that: `as_secs_f64` (https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.as_secs_f64)

ikopylov (approved):
One minor comment, otherwise LGTM

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +182,138 @@ enum CommandError {
...
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+struct DiskStats {
+    reads: u128,
+    writes: u128,
+    read_time: u128,
+    write_time: u128,
```
ikopylov, line 283: Why do you use `u128`? I checked the documentation and these values are stored by the OS in `unsigned long`, which is `u64` in Rust.

https://www.kernel.org/doc/Documentation/admin-guide/iostats.rst

bob/src/hw_metrics_collector.rs
```diff
@@ -235,54 +323,94 @@ impl DiskStatCollector {
...
+                            let gauge_name = format!("{}_iops", ds.prefix);
+                            gauge!(gauge_name, iops);
+
+                            if diff.extended {
+                                let mut iowait = 0.;
+                                if diff.reads > 0 {
+                                    iowait += diff.read_time as f64 / diff.reads as f64;
+                                }
+                                if diff.writes > 0 {
+                                    iowait += diff.write_time as f64 / diff.writes as f64;
```
ikopylov, line 401: I just notice, that this calculation of `iowait` is incorrect. `iowait` is the percentage of CPU time spent waiting for an I/O operation to complete. What you calculate is just an average time of single I/O operation. 

I suggest to calculate disk utilization instead (`util`). It is the time spent doing I/Os (column 13) divided by the elapsed time in milliseconds

ikopylov (changes requested):
Changes are required

bob/src/hw_metrics_collector.rs
```diff
@@ -179,53 +182,136 @@ enum CommandError {
...
+    extended: bool,
 }
 
+impl std::ops::Sub for DiskStats {
+    type Output = DiskStats;
+    fn sub(self, rhs: DiskStats) -> DiskStats {
+        DiskStats {
+            reads: self.reads - rhs.reads,
+            writes: self.writes - rhs.writes,
+            io_time: self.io_time - rhs.io_time,
```
ikopylov, line 292: These 3 subtracts can overflow and it is a valid behavior. I just checked that Rust panic on overflow with normal subtract, so you need to use operation supporting overflow (`wrapping_sub`)

bob/src/hw_metrics_collector.rs
```diff
@@ -235,54 +321,82 @@ impl DiskStatCollector {
...
+            // successfull reads count is in 3rd column
+            parts[3].parse::<u64>(),
+            // successfull writes count is in 7th column
+            parts[7].parse::<u64>(),
+        ) {
+            new_ds.reads = r_ios;
+            new_ds.writes = w_ios;  
+            if new_ds.extended {
+                // time spend doing i/o operations is in 13th column
+                if let Ok(io_time) = parts[13].parse::<u64>() {
```
ikopylov, line 354: Indices are zero based, so column 13 is at index 12
```suggestion
                if let Ok(io_time) = parts[12].parse::<u64>() {
```

ikopylov (approved):
LGTM

PR #480, opened 2022-04-30
logging reduce & interval logger backport

Boneyan:
Put error logs now agregate in one line every 5 sec (in case of disk disconnection)


ikopylov (approved):
LGTM

PR #492, opened 2022-05-26
466 index memory limit

### LINKED ISSUE #466: Add a configuration parameter to limit the memory occupied by indexes
Currently we have `open_blobs_soft_limit` and `open_blobs_hard_limit` parameters, but they worked on vdisk level only. For configuration simplicity and for better memory control it would be nice to introduce global memory limit for all indexes in memory

idruzhitskiy:
Closes #466

### DIFF
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -226,6 +226,29 @@ impl BackendStorage for Pearl {
         }
     }
 
+    async fn close_oldest_active_blob(&self) -> bool {
+        let mut oldest: Option<Holder> = None;
+        for dc in self.disk_controllers.iter() {
+            if let Some(holder) = dc.find_oldest_inactive_holder().await {
+                if holder.end_timestamp()
+                    < oldest
+                        .as_ref()
+                        .map(|h| h.end_timestamp())
+                        .unwrap_or(u64::MAX)
+                {
+                    oldest = Some(holder);
+                }
+            }
+        }
+
+        if let Some(h) = oldest {
+            h.close_active_blob().await;
+            true
+        } else {
+            false
+        }
+    }
+
     async fn delete(&self, op: Operation, key: BobKey) -> Result<u64, Error> {
         debug!("DELETE[{}] from pearl backend. operation: {:?}", key, op);
         let dc_option = self
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -226,7 +226,7 @@ impl BackendStorage for Pearl {
         }
     }
 
-    async fn close_oldest_active_blob(&self) -> bool {
+    async fn close_oldest_active_blob(&self) -> Option<usize> {
         let mut oldest: Option<Holder> = None;
         for dc in self.disk_controllers.iter() {
             if let Some(holder) = dc.find_oldest_inactive_holder().await {
@@ -243,9 +243,9 @@ impl BackendStorage for Pearl {
 
         if let Some(h) = oldest {
             h.close_active_blob().await;
-            true
+            Some(h.index_memory().await)
         } else {
-            false
+            None
         }
     }
 
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -474,6 +474,27 @@ impl Group {
         String::from_utf8(hex).unwrap()
     }
 
+    pub(crate) async fn find_oldest_inactive_holder(&self) -> Option<Holder> {
+        let holders_lock = self.holders();
+        let holders = holders_lock.read().await;
+        let mut result: Option<Holder> = None;
+        for holder in holders.iter() {
+            if let Some(is_empty) = holder.active_blob_is_empty().await {
+                if !is_empty && holder.is_outdated() {
+                    if holder.end_timestamp()
+                        < result
+                            .as_ref()
+                            .map(|h| h.end_timestamp())
+                            .unwrap_or(u64::MAX)
+                    {
+                        result = Some(holder.clone());
+                    }
+                }
+            }
+        }
+        result
+    }
+
     pub(crate) async fn close_unneeded_active_blobs(&self, soft: usize, hard: usize) {
         let holders_lock = self.holders();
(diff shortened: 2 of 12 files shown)


bob/src/cleaner.rs
```diff
@@ -29,23 +32,42 @@ impl Cleaner {
...
-            backend.close_unneeded_active_blobs(soft, hard).await;
+            if soft.is_some() || hard.is_some() {
+                let soft = soft.unwrap_or(1);
+                let hard = hard.unwrap_or(10);
+                backend.close_unneeded_active_blobs(soft, hard).await;
+            }
             if let Some(limit) = bloom_filter_memory_limit {
                 backend.offload_old_filters(limit).await;
             }
+            if let Some(limit) = index_memory_limit {
```
ikopylov, line 58: Two ways to offload index memory are better to place side by side and before bloom filters offloading

ikopylov, line 58: Two ways to offload index memory are better to place side by side and before bloom filters offloading

ikopylov:
Can you replace SystemTime for tracking blob access time to something more performant in this request?

ikopylov:
Can you replace SystemTime for tracking blob access time to something more performant in this request?

bob-backend/src/pearl/core.rs
```diff
@@ -226,6 +226,29 @@ impl BackendStorage for Pearl {
...
+                        .unwrap_or(u64::MAX)
+                {
+                    oldest = Some(holder);
+                }
+            }
+        }
+
+        if let Some(h) = oldest {
+            h.close_active_blob().await;
+            Some(h.index_memory().await)
```
ikopylov, line 246: Will this give the correct result when the blob is already closed? Maybe `index_memory` call should be right before `close_active_blob`

bob-backend/src/pearl/group.rs
```diff
@@ -474,17 +474,40 @@ impl Group {
         String::from_utf8(hex).unwrap()
     }
 
+    pub(crate) async fn find_oldest_inactive_holder(&self) -> Option<Holder> {
+        let holders_lock = self.holders();
+        let holders = holders_lock.read().await;
+        let mut result: Option<Holder> = None;
+        for holder in holders.iter() {
```
ikopylov, line 481: Is there a guarantee that `holders` are ordered by timestamp? If yes, then we can stop on the first holder. If no, then `close_unneeded_active_blobs` has a bug, thinking that they are ordered

bob-backend/src/pearl/group.rs
```diff
@@ -474,17 +474,40 @@ impl Group {
...
+            if let Some(is_empty) = holder.active_blob_is_empty().await {
+                if !is_empty && holder.is_outdated() {
```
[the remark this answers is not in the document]
ikopylov: Additionally, I think, there is no reason to check for `is_empty`

config-examples/node.yaml
```diff
@@ -55,6 +55,8 @@ http_api_port: 8000
 http_api_address: 0.0.0.0
 # [size] memory limit for all bloom filters. Unlimited if not specified
 bloom_filter_memory_limit: 8 GiB
+# [size] memory limit for all indexes. Unlimited if not specified
+index_memory_limit: 2 GiB
```
ikopylov, line 59: Set `8 GiB` by default

dockerfiles/default-configs/node.yaml
```diff
@@ -32,6 +32,10 @@ cleanup_interval: 1h
 open_blobs_soft_limit: 2
 # hard limit for count of max blobs to remain in ram
 open_blobs_hard_limit: 10
+# [size] memory limit for all bloom filters. Unlimited if not specified
+bloom_filter_memory_limit: 8 GiB
+# [size] memory limit for all indexes. Unlimited if not specified
+index_memory_limit: 2 GiB
```
ikopylov, line 38: Same here

bob-backend/src/pearl/group.rs
idruzhitskiy: `close_unneded_active_blobs` uses custom sort on filtered holders, there is no assumption of holders ordering. Here we do not assume this too

bob-backend/src/pearl/group.rs
```diff
@@ -474,14 +474,36 @@ impl Group {
         String::from_utf8(hex).unwrap()
     }
 
+    pub(crate) async fn find_oldest_inactive_holder(&self) -> Option<Holder> {
+        let holders_lock = self.holders();
+        let holders = holders_lock.read().await;
+        let mut result: Option<Holder> = None;
+        let period = self.settings.timestamp_period_as_secs() / 2;
+        for holder in holders.iter() {
+            if holder.active_blob_is_empty().await.is_some() {
```
ikopylov, line 483: That is a weird way to check if an active blob is present :) I suggest to add a separate method `holder.has_active_blob()`

bob-backend/src/pearl/group.rs
```diff
@@ -474,14 +474,36 @@ impl Group {
...
+
     pub(crate) async fn close_unneeded_active_blobs(&self, soft: usize, hard: usize) {
         let holders_lock = self.holders();
         let holders = holders_lock.read().await;
 
         let mut total_open_blobs = 0;
         let mut close = vec![];
         for h in holders.iter() {
-            if !h.active_blob_is_empty().await {
+            if h.active_blob_is_empty().await.is_some() {
```
ikopylov, line 506: Same here

bob-backend/src/pearl/group.rs
```diff
@@ -474,14 +474,36 @@ impl Group {
         String::from_utf8(hex).unwrap()
     }
 
+    pub(crate) async fn find_oldest_inactive_holder(&self) -> Option<Holder> {
+        let holders_lock = self.holders();
+        let holders = holders_lock.read().await;
+        let mut result: Option<Holder> = None;
+        let period = self.settings.timestamp_period_as_secs() / 2;
```
ikopylov, line 481: I suggest to add 10 seconds to this value to better handle cases, when timestamp period is too small

ikopylov (approved):
LGTM

PR #511, opened 2022-06-05
508 publish available ram

### LINKED ISSUE #508: metrics: Publish available ram to `free_ram` metric
Currently, `total_ram - used_ram` is publish to `free_ram` metric. But some memory is reserved and cannot be used for allocation and that memory is not accounted by `used_ram`. We need to publish [available_memory](https://docs.rs/sysinfo/0.23.13/sysinfo/trait.SystemExt.html#tymethod.available_memory) to `free_ram` and `total_ram - free_ram` to `used_ram`

Boneyan:
Publish available ram metric instead of calculated free ram, used ram is now calculated
Closes #508

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Bob versions changelog
 
 
 #### Changed
+- Publish available ram metric instead of calculated free ram, used ram is now calculated (#508)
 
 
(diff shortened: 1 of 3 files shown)


ikopylov (approved):
LGTM

PR #517, opened 2022-06-15
501 occupied disk metric

### LINKED ISSUE #501: Add a metric to calculate disk space occupied solely by Bob
Bob keeps files open and keeps track of them in internal data structures, so it would be fairly easy to calculate total size of all files tracked by Bob. This metric should be cumulative, and besides, it would be nice to have the same metric for every physical disk

idruzhitskiy:
Closes #501


ikopylov:
LGTM, but should be updated after Pearl release

PR #524, opened 2022-06-25
518 implement semi automatic integration tests

Ziktorin:


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install compose
+        uses: KengoTODA/actions-setup-docker-compose@main
+        with:
+          version: '2.6.0'
+      - name: Get version data
+        run: |
+          echo "VERSION=$(cat bob-apps/Cargo.toml | grep ^version | head -1 | awk '{print $3}' | sed -e 's/"//g' | sed -e 's/-/./g')" >> $GITHUB_ENV
+          echo "VERSION_NO_FORMATTING=$(cat bob-apps/Cargo.toml | grep ^version | head -1 | awk '{print $3}' | sed -e 's/"//g')" >> $GITHUB_ENV
+      - name: Get tools
+        uses: dawidd6/action-download-artifact@v2
```
ikopylov, line 37: There is a default action for downloading artifacts: `actions/download-artifact@v2`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+      - name: Install deps
+        run: |
+          /usr/bin/pip3 install -r ./requirements.txt
+        working-directory: ./integration-tests
+      - name: Install compose
+        uses: KengoTODA/actions-setup-docker-compose@main
```
ikopylov, line 29: Have you checked that docker compose is not installed by default?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+  deploy-and-test:
+    strategy:
+      matrix:
+        key: [8, 16]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checks-out repo
+        uses: actions/checkout@v2
+      - name: Install python
+        uses: actions/setup-python@v4
```
ikopylov, line 21: Have you checked that python is not installed by default?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          version: '2.6.0'
+      - name: Get version data
+        run: |
+          echo "VERSION=$(cat bob-apps/Cargo.toml | grep ^version | head -1 | awk '{print $3}' | sed -e 's/"//g' | sed -e 's/-/./g')" >> $GITHUB_ENV
+          echo "VERSION_NO_FORMATTING=$(cat bob-apps/Cargo.toml | grep ^version | head -1 | awk '{print $3}' | sed -e 's/"//g')" >> $GITHUB_ENV
+      - name: Get tools
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: build-packages.yml
```
ikopylov, line 39: You should build binaries as a separate job of this workflow, because there is no guarantee, that `build-packages.yml` will be run every time this workflow is run.

Additionally, I think it is better to build docker image, not the binaries

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          workflow_conclusion: success
+          branch: v${{ env.VERSION_NO_FORMATTING }}
+          name: bob-binaries-${{ matrix.key }}
+          check_artifacts:  true
+      - name: Generate config files for docker and bob
+        env:
+          AMOUNT_OF_NODES: "4"
```
ikopylov, line 46: You can use input parameters to allow customization by the user (https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs)

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          ./generate_from_templates.py -a ${{ env.AMOUNT_OF_NODES }} -v qoollo/bob:${{ env.VERSION }}-key${{ matrix.key }} --path ${{ env.STORAGE_PATH }}
+        working-directory: ./integration-tests
+      - name: Deploy docker containers
+        env:
+          PATH_TO_CONFIGS: ${{ env.STORAGE_PATH }}/generated_configs
+          REPLICAS: "2"
+          DISKS: "4"
```
ikopylov, line 54: Same here

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          ./deploy.py --path ${{ env.PATH_TO_CONFIGS }} -d ${{ env.DISKS }} -r ${{ env.REPLICAS }}
+        working-directory: ./integration-tests
+      - name: Run write/read tests
+        env:
+          PAYLOAD: "10240"
+          COUNT: "10000"
```
ikopylov, line 61: Same here

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
+name: integration-tests
+on:
```
ikopylov, line 2: You should allow manual start of this workflow: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

ikopylov: Additionally, I suggest to run this workflow when PR is approved: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review

integration-tests/Templates/node_template.yml.j2
```diff
@@ -0,0 +1,37 @@
...
+quorum: {{ quorum }}         # min count of successful operations on replicas to consider operation successful
+operation_timeout: {{ operation_timeout }}           # timeout for every GRPC operation, eg: [1y 2M 4w 8d 16h 32min 64s 128ms 256us]
+check_interval: {{ check_interval }}              # interval for checking connections , eg: [1y 2M 4w 8d 16h 32min 64s 128ms 256us]
+cluster_policy: {{ cluster_policy }}              # simple - without checking status [simple, quorum]
+backend_type: {{ backend_type }}                 # type of the backend [in_memory, stub, pearl]
+cleanup_interval: {{ cleanup_interval }}               # interval for checking for blobs cleanup
+open_blobs_soft_limit: {{ open_blobs_soft_limit }}           # soft limit for count of max blobs to remain in ram
+open_blobs_hard_limit: {{ open_blobs_hard_limit }}           # hard limit for count of max blobs to remain in ram
+init_par_degree: 4   # degree of parallelism in disks initialization
+http_api_port: 8000
```
ikopylov, line 13: I'd suggest to specify `http_api_port` as a parameter too

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,108 @@
...
+help='sets bloom filter buffer size in bits count, best value ~= max_data_in_blob.')
+parser.add_argument('--root-dir-name', dest='root_dir_name', type=str, default='bob')
+parser.add_argument('--alien-root-dir-name', dest='alien_root_dir_name', type=str, default='/bob/data/d1/alien')
+parser.add_argument('-t', dest='timestamp_period', type=str, default='1m', help='period when new pearl directory created.')
+parser.add_argument('--create-pearl-wait-delay', dest='create_pearl_wait_delay', type=str, default='100ms', help='each thread will wait this period if another thread creating pearl.')
+parser.add_argument('--metrics-name', dest='metrics_name', type=str, default='bob', help='add base name for metrics.')
+parser.add_argument('--graphite-enabled', dest='graphite_enabled', type=str, default='true', choices=['true', 'false'])
+parser.add_argument('--prometheus-enabled', dest='prometheus_enabled', type=str, default='false', choices=['true', 'false'])
+parser.add_argument('--path', dest='path', type=str, help='sets path to directory where configs will be generated.', default='/tmp')
```
ikopylov, line 42: Too many arguments. IMHO it is better to get optional arguments from environment variables

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,108 @@
...
+
+args = parser.parse_args()
+
+path = pathified(args.path) + '/generated_configs'
+
+os.makedirs(path, exist_ok=True, mode=0o777)
+
+for filename in os.listdir(path):
+    try:
+        os.remove(f"{path}/{filename}")
```
ikopylov, line 52: This is dangerous. I suggest to check, that folder is empty. If it is not, then stop with error

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,108 @@
...
+    backend_type=args.backend_type, cleanup_interval=args.cleanup_interval, open_blobs_soft_limit=args.open_blobs_soft_limit,
+    open_blobs_hard_limit=args.open_blobs_hard_limit, bloom_filter_memory_limit=args.bloom_filter_memory_limit, 
+    auth_type=args.auth_type, index_memory_limit=args.index_memory_limit, enable_aio=args.enable_aio, 
+    disks_events_logfile=args.disks_events_logfile, max_blob_size=args.max_blob_size, allow_duplicates=args.allow_duplicates, 
+    max_data_in_blob=args.max_data_in_blob, blob_file_name_prefix=args.blob_file_name_prefix,
+    fail_retry_timeout=args.fail_retry_timeout, alien_disk=args.alien_disk, 
+    bloom_filter_max_buf_bits_count=args.bloom_filter_max_buf_bits_count, root_dir_name=args.root_dir_name, 
+    alien_root_dir_name=args.alien_root_dir_name, timestamp_period=args.timestamp_period,
+    create_pearl_wait_delay=args.create_pearl_wait_delay, metrics_name=args.metrics_name, graphite_enabled=args.graphite_enabled,
+    prometheus_enabled=args.prometheus_enabled))
```
ikopylov, line 87: This will be hard to maintain. Please, find out how to translate all arguments into `render` parameters without listing them all explicitly

.github/workflows/integration-tests.yml
ikopylov: I see, that you use `ccg` and `bobp` binaries, so you should have 2 additional jobs: 
1. Build binaries
2. Build docker image

integration-tests/deploy.py
```diff
@@ -0,0 +1,89 @@
...
+    os.chmod(path='../ccg', mode=0o771)
+    os.chmod(path='../bobp', mode=0o771)
+except OSError as e:
+    print(e)
+
+try:
+    pr = subprocess.check_output(shlex.split(f'../ccg new -i {args.path}/cluster.yaml.bobnet -o {args.path}/cluster.yaml.bobnet {args_str.rstrip()}'))
+    if str(pr).find('ERROR') != -1:
+        print(pr)
+        sys.exit()
```
ikopylov, line 51: You should specify an error code, because it is not a normal exit

integration-tests/deploy.py
```diff
@@ -0,0 +1,89 @@
...
+except d_err.NotFound:
+    ipam_config = d_types.IPAMConfig(pool_configs=[d_types.IPAMPool(subnet='172.21.0.0/24', gateway='172.21.0.1')])
+    bobnet = client.networks.create('bob_net', driver='bridge', ipam=ipam_config, attachable=True)
+
+start_path = os.getcwdb()
+good_path = os.path.abspath(args.path)
+
+if not 'cluster.yaml.bobnet' in os.listdir(good_path):
+    print('Cluster config not found in the specifed directory.')
+    sys.exit()
```
ikopylov, line 39: Error code required

integration-tests/deploy.py
```diff
@@ -0,0 +1,89 @@
...
+
+try:
+    os.chmod(path='../ccg', mode=0o771)
+    os.chmod(path='../bobp', mode=0o771)
+except OSError as e:
+    print(e)
```
ikopylov, line 45: Error code required

integration-tests/deploy.py
```diff
@@ -0,0 +1,89 @@
...
+
+try:
+    pr = subprocess.check_output(shlex.split(f'../ccg new -i {args.path}/cluster.yaml.bobnet -o {args.path}/cluster.yaml.bobnet {args_str.rstrip()}'))
+    if str(pr).find('ERROR') != -1:
+        print(pr)
+        sys.exit()
+except subprocess.CalledProcessError:
+    print(pr.stderr)
+    sys.exit()
```
ikopylov, line 54: Error code required

integration-tests/deploy.py
```diff
@@ -0,0 +1,89 @@
...
+try:
+    client.networks.get('bob_net')
+    d_cli.compose.up(detach=True)
+    print('Containers are running!')
+except d_err.NotFound:
+    print('docker network not found')
+    sys.exit()
+
+try:
+    os.chdir(start_path)
```
ikopylov, line 77: Please, add comments to this file

integration-tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+        args_str += f'{key} {final_args.get(key)} '
+
+
+#write/read tests
+try:
+    p = subprocess.check_output(shlex.split(f'../bobp -b put {args_str.rstrip()}'))
+    print(str(p))
+except subprocess.CalledProcessError as e:
+    print(e)
+    sys.exit()
```
ikopylov, line 30: Error code here and in other places

integration-tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+args = parser.parse_args()
+final_args = {'-c':args.count, '-l':args.payload, '-h':args.node, '-s':args.start, '-e':args.end, '-t':args.threads, '--mode':args.mode, '-k':args.keysize}
+
+args_str = str()
+for (key) in final_args:
+    if final_args.get(key) != None:
+        args_str += f'{key} {final_args.get(key)} '
+
+
+#write/read tests
```
ikopylov, line 24: Add `exists` stage too

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          workflow: build-packages.yml
+          workflow_conclusion: success
+          branch: v${{ env.VERSION_NO_FORMATTING }}
+          name: bob-binaries-${{ matrix.key }}
+          check_artifacts:  true
+      - name: Generate config files for docker and bob
+        env:
+          AMOUNT_OF_NODES: "4"
+        run: |
+          ./generate_from_templates.py -a ${{ env.AMOUNT_OF_NODES }} -v qoollo/bob:${{ env.VERSION }}-key${{ matrix.key }} --path ${{ env.STORAGE_PATH }}
```
ikopylov, line 48: Docker image should be built within this workflow. You can't use the image from DockerHub because it does not contain the current changes

ikopylov (changes requested):
Changes are required

ikopylov:
@Ziktorin  Please, merge master and resolve merge conflicts. Also, you should update `node.yaml` template, because new parameter was added to config (`index_memory_limit`)

integration-tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+try:
+    p = subprocess.check_output(shlex.split(f'../bobp -b put {args_str.rstrip()}'))
+    print(str(p))
+except subprocess.CalledProcessError as e:
+    print(e)
+    sys.exit()
+
+
+try:
+    p = subprocess.check_output(shlex.split(f'../bobp -b get {args_str.rstrip()}'))
```
ikopylov, line 34: I suggest using a different node for reading, not the one to which the write was made

integration-tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+args_str = str()
+for (key) in final_args:
+    if final_args.get(key) != None:
+        args_str += f'{key} {final_args.get(key)} '
+
+
+#write/read tests
+try:
+    p = subprocess.check_output(shlex.split(f'../bobp -b put {args_str.rstrip()}'))
+    print(str(p))
```
ikopylov, line 27: You should parse `bobp` output and check that there were no errors. Without this, the test is useless

integration-tests/tests.py
ikopylov, line 34: You should parse `bobp` output and check that there were no errors. Without this, the test is useless

.github/workflows/integration-tests.yml
Ziktorin: It seems that `actions/download-artifact@v2` does not support downloading artifacts from other workflows, so it is necessary to use this one instead.

ikopylov: There is no need to download artifacts from other workflow. The proper way is to build all binaries within this workflow

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
+name: integration-tests
+on:
+  push:
+    branches: [ "518-implement-semi-automatic-integration-tests" ]
+  pull_request:
+    branches: [ "518-implement-semi-automatic-integration-tests" ]
+  pull_request_review:
```
ikopylov, line 7: Workflow should be started only when PR approved. See `Running a workflow when a pull request is approved` section in documentation

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+    branches: [ "518-implement-semi-automatic-integration-tests" ]
+  workflow_dispatch:
+    inputs:
+      amount_of_nodes:
+        description: 'Amount of nodes in bob cluster for tests'
+        required: true
+        default: '4'
+      storage_path:
```
ikopylov, line 15: This parameter can be removed. It is pointless to configure path

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+      disks:
+        description: 'Vdisks count'
+        required: true
+        default: '4'
+      replicas:
+        description: 'Amount of vdisks replicas'
+        required: true
+        default: '2'
+      payload:
+        description: 'Test payload'
```
ikopylov, line 28: `Playload size`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Checks-out repo
+        uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+      - name: Setup default variables for auto runs
+        run: |
+          INPUT=${{ github.event.inputs.amount_of_nodes }}
+          echo "AMOUNT_OF_NODES=${INPUT:-"4"}" >> $GITHUB_ENV
```
ikopylov, line 54: Add some prefix to env variables name (e.g. `BOB_`) to avoid collisions

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+          echo "BOB_COMMIT_HASH=$(echo ${GITHUB_SHA} | cut -c1-8)" >> $GITHUB_ENV
+          echo "BOB_KEY_SIZE=${{ matrix.key }}" >> $GITHUB_ENV
+          sudo apt-get update && sudo apt-get install -y --no-install-recommends musl-tools
+      - name: Install latest stable
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: ${{ env.TARGET }}
+          default: true
+      - name: build binaries
```
ikopylov, line 84: This should be a separate job within workflow

integration-tests/tests.py
Ziktorin: What would cause failed test of `exists` in this case? Non-matching amount of put keys in write stage to `--count` of `exists`, or `exists` errors that are output at the end of execution similar to that of `put`/`get` stages?

integration-tests/generate_from_templates.py
Ziktorin: It seems that getting some variables as enviromental and others as script arguments would make code messier at render step. At this point, rendering variables for templates is done with one line: 
`f.write(template.render(vars(args), node_number=item))` 
and taking them from enviromental variables would demand explicitly equating them to their template counterpart, i.e: 
`f.write(template.render(log_config=${BOB_LOG_CONFIG}, user_config=${BOB_USER_CONFIG}, ...))`
Maybe it would be better left as it is?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+  #pull_request_review:
+  #  types: [submitted]
+  #  branches: [ "518-implement-semi-automatic-integration-tests" ]
+  workflow_dispatch:
+    inputs:
+      bob_amount_of_nodes:
+        description: 'Amount of nodes in bob cluster for tests'
+        required: true
+        default: '4'
+      bob_disks:
```
ikopylov, line 14: `bob_vdisk_count`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+        description: 'Vdisks count'
+        required: true
+        default: '4'
+      bob_replicas:
```
ikopylov, line 18: `bob_replicas_amount`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+        description: 'Amount of vdisks replicas'
+        required: true
+        default: '2'
+      bob_payload:
```
ikopylov, line 22: `bob_payload_size`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+        description: 'Payload size'
+        required: true
+        default: '10240'
+      bob_count:
```
ikopylov, line 26: `bob_record_count`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
+name: integration-tests
+on:
+  push:
+    branches: [ "518-implement-semi-automatic-integration-tests" ] #REMOVE WHEN ON MASTER 
+  #pull_request_review:
+  #  types: [submitted]
+  #  branches: [ "518-implement-semi-automatic-integration-tests" ]
+  workflow_dispatch:
+    inputs:
+      bob_amount_of_nodes:
```
ikopylov, line 10: `bob_nodes_amount`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+        required: true
+        default: '10240'
+      bob_count:
+        description: 'Amount of records to process during test'
+        required: true
+        default: '10000'
+
+jobs:
+  build-binaries:
+    # if: github.event.review.state == 'approved'
```
ikopylov, line 33: This is incorrect. You should run job when this is not review submit or when it is an approved review submit. Thus you should add a check for event type

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: bob-binaries-${{ matrix.key }}
+          path: |
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/bobd
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/bobp
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/ccg
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/brt
+  build-docker-images:
```
ikopylov, line 69: Please, add empty lines to separate jobs from each other

integration-tests/generate_from_templates.py
ikopylov: We decided to keep things as is

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,103 @@
...
+parser.add_argument('--fail-retry-timeout', dest='fail_retry_timeout', type=str, default='100ms', help='retry to reinit pearl backend after fail.')
+parser.add_argument('--alien-disk', dest='alien_disk', type=str, default='d1')
+parser.add_argument('--bloom-filter-max-buf-bits-count', dest='bloom_filter_max_buf_bits_count', type=int, default=10000, 
+help='sets bloom filter buffer size in bits count, best value ~= max_data_in_blob.')
+parser.add_argument('--root-dir-name', dest='root_dir_name', type=str, default='bob')
+parser.add_argument('--alien-root-dir-name', dest='alien_root_dir_name', type=str, default='/bob/data/d1/alien')
+parser.add_argument('-t', dest='timestamp_period', type=str, default='1m', help='period when new pearl directory created.')
+parser.add_argument('--create-pearl-wait-delay', dest='create_pearl_wait_delay', type=str, default='100ms', help='each thread will wait this period if another thread creating pearl.')
+parser.add_argument('--metrics-name', dest='metrics_name', type=str, default='bob', help='add base name for metrics.')
+parser.add_argument('--graphite-enabled', dest='graphite_enabled', type=str, default='true', choices=['true', 'false'])
```
ikopylov, line 41: It is better to use `false` as a default value here

ikopylov, line 41: It is better to use `false` as default value here

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,103 @@
...
+parser.add_argument('--prometheus-enabled', dest='prometheus_enabled', type=str, default='false', choices=['true', 'false'])
+parser.add_argument('--path', dest='path', type=str, help='sets path to directory where configs will be generated.', default='/tmp')
+
+
+args = parser.parse_args()
+
+path = pathified(args.path) + '/generated_configs'
```
ikopylov, line 48: https://www.geeksforgeeks.org/python-os-path-join-method/

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,103 @@
...
+
+os.makedirs(path, exist_ok=True, mode=0o777)
+
+
+if len(os.listdir(path)) == 0:
+    for filename in os.listdir(path):
```
ikopylov, line 54: Why are you removing files when you've already checked, that the directory is empty?

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,103 @@
...
+            os.remove(f"{path}/{filename}")
+        except OSError:
+            shutil.rmtree(f"{path}/{filename}")
+else:
+    sys.exit('Directory not empty.')
+
+try:
+    original_umask = os.umask(0)
+    for i in range(args.amount_of_nodes):
+        os.makedirs(path + f'/data{i}/d1', exist_ok=True, mode=0o777)
```
ikopylov, line 65: Use `os.path.join()`

integration-tests/tests.py
```diff
@@ -0,0 +1,60 @@
...
+    if behaviour in {'put', 'get'}:
+        try:
+            print(f'Running bobp -b {str(behaviour)} {args.rstrip()}')
+            p = subprocess.check_output(shlex.split(f'./bobp -b {behaviour} {args.rstrip()}'))
+            print(str(p))
+            if f'{behaviour} errors:' in str(p) or f'panicked' in str(p):
+                sys.exit(f'{behaviour} test failed, see output')
+        except subprocess.CalledProcessError as e:
+            sys.exit(str(e.stderr))
+    elif behaviour == 'exist':
```
ikopylov, line 15: Add `else` at the end that prints error about unknown behavior. This will simplify detecting problems in future, if behavior names will be changed

integration-tests/tests.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            print(f'Running bobp -b {str(behaviour)} {args.rstrip()}')
+            p = subprocess.check_output(shlex.split(f'./bobp -b {behaviour} {args.rstrip()}'))
+            print(str(p))
```
ikopylov, line 19: These 3 lines above are common for both branches. I suggest to make them common by moving them out of `if..else` statement

integration-tests/tests.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_run_args(mode, args):
+    return {'-c':args.count, '-l':args.payload, '-h':f'{args.node}', '-s':args.start, '-e':args.end, '-t':args.threads, '--mode':args.mode, '-k':args.keysize, '-p':test_run_config.get(mode)}
+
+test_run_config = {'put':'20000', 'get':'20001'}
+
+parser = argparse.ArgumentParser(description='This script launches bob tests with given configuration.')
+parser.add_argument('-c', dest='count', type=int, help='amount of entries to process', required=True)
+parser.add_argument('-l', dest='payload', type=int, help='payload in bytes', required=True)
+parser.add_argument('-n', dest='node', type=str, help='target node address', required=True)
```
ikopylov, line 41: I suggest to pass nodes list and choose different nodes for `put`, `get` and `exist` tests

integration-tests/tests.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python3
+
+import subprocess, argparse, shlex, sys, re
+
+def run_tests(behaviour, args):
+    if behaviour in {'put', 'get'}:
+        try:
+            print(f'Running bobp -b {str(behaviour)} {args.rstrip()}')
+            p = subprocess.check_output(shlex.split(f'./bobp -b {behaviour} {args.rstrip()}'))
+            print(str(p))
```
ikopylov, line 10: There are `\n` text in the output instead of normal line ending

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+        with:
+          name: bob-binaries-${{ matrix.key }}
+          path: |
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/bobd
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/bobp
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/ccg
+            target/${{ env.TARGET }}/${{ env.BUILD_MODE }}/brt
+
+  build-docker-images:
+    if: github.event.review.state == 'approved' || github.event.type != 'PullRequestReviewEvent'
```
ikopylov, line 70: Test runs on every comment to this PR. This should be fixed. Please recheck, that you write the correct event type

ikopylov:
Integration tests sometimes failed on `exist`: https://github.com/qoollo/bob/actions/runs/3175984263/jobs/5174881507
Please, check that are no aliens appeared during the test

ikopylov:
Please, look at unresolved comments in this PR. You missed some of them

.github/workflows/integration-tests.yml
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: added

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checks-out repo
+        uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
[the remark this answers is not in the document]
Ziktorin: changed to 3.x

ikopylov:
Test

ikopylov (approved):
Test

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,178 @@
...
+        required: true
+        default: '10240'
+      bob_record_count:
+        description: 'Amount of records to process during test'
+        required: true
+        default: '10000'
+
+jobs:
+  build-binaries:
+    if: ${{ github.event.review.state == 'approved' || (github.event_name != 'pull_request_review' && github.event_name != 'push') }}
```
ikopylov, line 34: Why did you discarded `push` events?
Also the condition is not good: you can check for review state only if it is a `pull_request_review` event. 
So the condition should be something like this: `github.event_name != 'pull_request_review' || github.event.review.state == 'approved'`

ikopylov (approved):
Test trigger

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,178 @@
...
+            /tmp/bob-image-key${{ matrix.key }}.tar
+
+  deploy-and-test:
+    if: ${{ github.event.review.state == 'approved' || (github.event_name != 'pull_request_review' && github.event_name != 'push') }}
+    needs: [build-binaries, build-docker-images]
+    strategy:
+      matrix:
+        key: [8, 16]
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
```
ikopylov, line 121: Looks like `TARGET` is not used in this job

integration-tests/deploy.py
```diff
@@ -0,0 +1,88 @@
...
+#run docker containers
+try:
+    client.networks.get('bob_net')
+    d_cli.compose.up(detach=True)
+    print(f'Services are initilized:\n{d_cli.container.list()}')
+except d_err.NotFound:
+    sys.exit('Docker network not found')
+
+try:
+    if len(d_cli.container.list()) < os.environ['BOB_NODES_AMOUNT']:
```
ikopylov, line 81:
```
Traceback (most recent call last):
  File "./deploy.py", line 81, in <module>
    if len(d_cli.container.list()) < os.environ['BOB_NODES_AMOUNT']:
TypeError: '<' not supported between instances of 'int' and 'str'
Error: Process completed with exit code 1.
```

ikopylov (changes requested):
Changes are required

integration-tests/tests.py
```diff
@@ -0,0 +1,67 @@
...
+parser.add_argument('-t', dest='threads', type=int, help='amount of working threads', default=1)
+parser.add_argument('--mode', dest='mode', type=str, help='random or normal', choices=['random', 'normal'], default='normal')
+parser.add_argument('-k', dest='keysize', type=int, help='size of binary key (8 or 16)', choices=[8, 16], default=8)
+
+parsed_args = parser.parse_args()
+
+test_run_config = dict()
+iter = 0
+try:
+    for item in ['put', 'get', 'exist']:
```
ikopylov, line 49: `for item in run_options:`

integration-tests/tests.py
```diff
@@ -0,0 +1,67 @@
...
+        test_run_config[item]=str(20000+(iter % int(os.environ['BOB_NODES_AMOUNT'])))
```
ikopylov, line 50: I don't see the place where this dictionary is used

ikopylov (approved):
test run

ikopylov (approved):
test run 2

ikopylov (approved):
test run 3

ikopylov (approved):
test run 4

ikopylov (approved):
test run 5

ikopylov (changes requested):
Changes are required

ikopylov (approved):
Run test 1

ikopylov (approved):
Run test 2

ikopylov (approved):
Run test 3

ikopylov (approved):
Run test 4

ikopylov (approved):
Run test 5

integration-tests/tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+import subprocess, argparse, shlex, sys, re, os
+
+run_options = ['put','get','exist']
+
+test_run_config = dict()
+iter = 0
+try:
+    for item in run_options:
+        test_run_config[item]=str(20000+(iter % int(os.environ['BOB_NODES_AMOUNT']))) #used in get_run_args()
```
ikopylov, line 11: This is the only place in this script, where env variable is used. This is a bit strange. I suggest to pass `min_port` and `max_port` as command line arguments and used them to generate target ports

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,182 @@
...
+        run: |
+          ./generate_from_templates.py -a ${{ env.BOB_NODES_AMOUNT }} -v qoollo/bob:${{ env.VERSION }}-key${{ matrix.key }} --path /tmp
+        working-directory: ./integration-tests
+      - name: Deploy docker containers
+        run: |
+          ./deploy.py --path /tmp/generated_configs -d ${{ env.BOB_VDISK_COUNT }} -r ${{ env.BOB_REPLICAS_AMOUNT }}
+        working-directory: ./integration-tests
+      - name: Run write/read/exist tests
+        run: |
+          ./tests.py -l ${{ env.BOB_PAYLOAD_SIZE }} -c ${{ env.BOB_RECORD_COUNT }} -f ${{ env.BOB_FIRST_INDEX }} -n 0.0.0.0 -k ${{ matrix.key }}
```
ikopylov, line 181: `-n 0.0.0.0`
How that works? Shouldn't it be 127.0.0.1 here?

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+import os, argparse, sys, subprocess, shlex, docker,  requests, json
+from python_on_whales import docker as d_cli
+from docker import errors as d_err
+from docker import types as d_types
+from retry import *
+
+
+def request_metrics(port):
+    try:
+        return requests.get(f"http://0.0.0.0:{port}/metrics")
```
ikopylov, line 11: Again, I expect `0.0.0.0` should not work here. According to my expectation it should be `127.0.0.1`. Why does it work?

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+    except requests.RequestException:
+        sys.exit('Failed to get backend status.')
+    
+         
+@retry((UserWarning), tries=5, delay=1, backoff=2, max_delay=5)
+def ensure_backend_up():
+    try:
+        for item in range(int(os.environ['BOB_NODES_AMOUNT'])):
+            response = request_metrics(f"800{item}")
```
ikopylov, line 20: Hm. What if the number of nodes is >= 10? You should use `8000 + item`

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+    try:
+        return requests.get(f"http://0.0.0.0:{port}/metrics")
+    except requests.RequestException:
+        sys.exit('Failed to get backend status.')
+    
+         
+@retry((UserWarning), tries=5, delay=1, backoff=2, max_delay=5)
+def ensure_backend_up():
+    try:
+        for item in range(int(os.environ['BOB_NODES_AMOUNT'])):
```
ikopylov, line 19: I suggest to pass min and max port as arguments and not use env variable, because this is the only place where it is needed and that behavior is not expected

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+#run docker containers
+try:
+    client.networks.get('bob_net')
+    d_cli.compose.up(detach=True)
+    print(f'Services are initilized:\n{d_cli.container.list()}')
+except d_err.NotFound:
+    sys.exit('Docker network not found')
+
+try:
+    if len(d_cli.container.list()) < int(os.environ['BOB_NODES_AMOUNT']):
```
ikopylov, line 110: Again: I suggest to avoid using env variables. You can pass `min_port` and `nodes_amount` as the arguments to that script

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+        sys.exit('One or more bob docker containers are not running.')
+except KeyError:
+    sys.exit('Nodes amount is not set.')
+except ValueError:
+    sys.exit('Amount of nodes has unexpected value.')
+
+#ensure bob initilized in container
+ensure_backend_up()
```
ikopylov, line 118: I remember we talked that you would add `sleep` here to ensure for all connections between nodes are established

integration-tests/deploy.py
```diff
@@ -0,0 +1,121 @@
...
+except OSError as e:
+    sys.exit(e)
+
+#run cluster generation
+try:
+    pr = subprocess.check_output(shlex.split(f'./ccg new -i {args.path}/cluster.yaml.bobnet -o {args.path}/cluster.yaml.bobnet {args_str.rstrip()}'))
+    if str(pr).find('ERROR') != -1:
+        sys.exit(str(pr))
+except subprocess.CalledProcessError:
+    sys.exit(pr.stderr)
```
ikopylov, line 89: I don't know python well, but looks like `pr` is out of scope here, so there will not be access to `stderr`

ikopylov (approved):
Test trigger

ikopylov (approved):
test trigger 2

ikopylov (approved):
test trigger 3

ikopylov (approved):
test trigger 4

ikopylov (approved):
test trigger 5

ikopylov (changes requested):
Stop tests

.github/workflows/integration-tests.yml
Ziktorin: It works as it is, but i will change it to be 127.0.0.1. I assume that it would in fact be more correct to use 127.0.0.1 here

integration-tests/deploy.py
Ziktorin: I checked it, it is in scope, stderr is handled correctly

integration-tests/tests.py
Ziktorin: fixed

integration-tests/tests.py
```diff
@@ -0,0 +1,64 @@
...
+            else:
+                print(f"{exists[0]} of {exists[1]} keys")   
+        else:
+            sys.exit('Unknown behaviour.')     
+    except subprocess.CalledProcessError as e:
+        sys.exit(str(e.stderr))
+    except Exception as e:
+        sys.exit(str(e))
+
+def get_run_args(mode, args):
```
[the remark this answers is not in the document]
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: fixed

integration-tests/deploy.py
Ziktorin: fixed

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        default: '100000'
+jobs:
+  build-binaries:
+    if: ${{ github.event_name != 'pull_request_review' || github.event.review.state == 'approved' }}
+    strategy:
+      matrix:
+        key: [8, 16]
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
+      BUILD_MODE: "release"
```
ikopylov, line 55: For tests we should build `debug` version

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+          echo "VERSION=$(cat bob-apps/Cargo.toml | grep ^version | head -1 | awk '{print $3}' | sed -e 's/"//g' | sed -e 's/-/./g')" >> $GITHUB_ENV
+          echo "BOB_COMMIT_HASH=$(echo ${GITHUB_SHA} | cut -c1-8)" >> $GITHUB_ENV
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: qoollo/bob
+          tags: |
+            ${{ env.VERSION }}-key${{ matrix.key }}
+      - name: Build image
```
ikopylov, line 106: Here we should also build debug version!

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        uses: actions-rs/toolchain@v1.0.6
+        with:
+          toolchain: stable
+          target: ${{ env.TARGET }}
+          default: true
+      - name: Build binaries
+        uses: actions-rs/cargo@v1.0.1
+        with:
+          command: build
+          args: --target=${{ env.TARGET }}
```
ikopylov, line 75: `--profile=${{ BUILD_MODE }}`
Unfortunately, `debug` should converted to `dev` here

ikopylov: Here is the hack:
`--profile=${{ (env.BUILD_MODE == 'debug' && 'dev') || env.BUILD_MODE }}`

.github/workflows/integration-tests.yml
ikopylov: I read more about profiles. It is better to use `test` profile, than `debug`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        default: '100000'
+jobs:
+  build-binaries:
+    if: ${{ github.event_name != 'pull_request_review' || github.event.review.state == 'approved' }}
+    strategy:
+      matrix:
+        key: [8, 16]
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
+      BUILD_MODE: "debug"
```
ikopylov, line 55: I read more about profiles. It is better to use `test` profile, than `debug`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        uses: actions-rs/toolchain@v1.0.6
+        with:
+          toolchain: stable
+          target: ${{ env.TARGET }}
+          default: true
+      - name: Build binaries
+        uses: actions-rs/cargo@v1.0.1
+        with:
+          command: build
+          args: --target=${{ env.TARGET }} --profile=${{ (env.BUILD_MODE == 'test' && 'dev') || env.BUILD_MODE }}
```
ikopylov, line 75: Why did you change comparison condition?
```suggestion
          args: --target=${{ env.TARGET }} --profile=${{ (env.BUILD_MODE == 'debug' && 'dev') || env.BUILD_MODE }}
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+      - name: List availible images
+        run: |
+          docker image ls
+      - name: Generate config files for docker and bob
+        run: |
+          ./generate_from_templates.py -a ${{ env.BOB_NODES_AMOUNT }} -v qoollo/bob:${{ env.VERSION }}-key${{ matrix.key }} --path /tmp -min_port ${{ env.BOB_MIN_PORT }}
+        working-directory: ./integration-tests
+      - name: Deploy docker containers
+        run: |
+          ./deploy.py --path /tmp/generated_configs -d ${{ env.BOB_VDISK_COUNT }} -r ${{ env.BOB_REPLICAS_AMOUNT }} -nodes_amount ${{ env.BOB_NODES_AMOUNT }} --bobp_actions_offset ${{ env.BOB_BOBP_ACTIONS_OFFSET }}
```
ikopylov, line 191: As I see `env.BOB_BOBP_ACTIONS_OFFSET` is set within `generate_from_templates.py` script, which is not obvious. I suggest to pass argument `--check-interval 5000ms` to `generate_from_templates.py` and pass the same hardcoded value to this parameter. And remove this environment variable

ikopylov, line 191: Also, `--bobp_actions_offset` is a super bad name for parameter. I suggest to rename it to `--cluster_start_waiting_time`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
```
ikopylov, line 259: Same here

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        working-directory: ./integration-tests
+      - name: Run alien tests
+        run: |
+          ./tests_aliens.py -l ${{ env.BOB_PAYLOAD_SIZE }} -c ${{ env.BOB_RECORD_COUNT }} -f ${{ env.BOB_FIRST_INDEX }} -n 127.0.0.1 -k ${{ matrix.key }}
+        working-directory: ./integration-tests
+  deploy-and-test-chain:
```
ikopylov, line 265: Add new line before this to visually separate jobs

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+      - name: List availible images
+        run: |
+          docker image ls
+      - name: Generate config files for docker and bob
+        run: |
+          ./generate_from_templates.py -a ${{ env.BOB_NODES_AMOUNT }} -v qoollo/bob:${{ env.VERSION }}-key${{ matrix.key }} --path /tmp -min_port ${{ env.BOB_MIN_PORT }}
+        working-directory: ./integration-tests
+      - name: Deploy docker containers
+        run: |
+          ./deploy.py --path /tmp/generated_configs -d ${{ env.BOB_VDISK_COUNT }} -r ${{ env.BOB_REPLICAS_AMOUNT }} -nodes_amount ${{ env.BOB_NODES_AMOUNT }} --bobp_actions_offset ${{ env.BOB_BOBP_ACTIONS_OFFSET }}
```
ikopylov, line 326: Same here

.github/workflows/integration-tests.yml
ikopylov, line 265: Name is not good. Better to name `deploy-and-test-with-bobt`.
Also, I think the part `deploy-and-` is redundant. You can name the tests like this:
1. `test-put-get-exist`
2. `test-aliens`
3. `test-with-bobt`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        working-directory: ./integration-tests
+      - name: Run chained operations tests
```
ikopylov, line 328:
```suggestion
      - name: Run tests with bobt
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        run: |
+          ./tests_chain.py -c ${{ env.BOB_BOBT_COUNT }} -s ${{ env.BOB_BOBT_FIRST_INDEX }} -e ${{ env.BOB_BOBT_LAST_INDEX }}
```
ikopylov, line 330: It would be good to change the script name as well

integration-tests/bob_backend_timer.py
```diff
@@ -0,0 +1,31 @@
...
+        return requests.get(f"http://127.0.0.1:{port}/metrics")
+    except requests.RequestException:
+        raise UserWarning('Failed to get metrics, retrying...')
+    
+         
+@retry((UserWarning), tries=5, delay=1, backoff=2, max_delay=5)
+def ensure_backend_up(bob_nodes_amount):
+    try:
+        for item in range(bob_nodes_amount):
+            response = request_metrics(8000 + item)
```
ikopylov, line 15: Pass min REST API port as the parameter to this function

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,105 @@
...
+parser.add_argument('-min_port', dest='min_port', type=int, required=True, help='Port of the first bob container.')
+
+args = parser.parse_args()
+
+path = os.path.join(pathified(args.path),'generated_configs')
+
+#get the offset for bobp operations and set in as env 
+bobp_actions_offset = float(str(Quantity(args.check_interval, scale='s')).replace(' s', '')) + 1
+with open(os.getenv('GITHUB_ENV'), 'a') as file:
+    file.write(f"BOB_BOBP_ACTIONS_OFFSET={bobp_actions_offset}")
```
ikopylov, line 54: This is not obvious. As I wrote before, it is better to remove this and pass hardcoded values as parameters

integration-tests/generate_from_templates.py
```diff
@@ -0,0 +1,105 @@
...
+
+os.makedirs(path, exist_ok=True, mode=0o777)
+
+
+if len(os.listdir(path)) != 0:
+    sys.exit('Directory not empty.')
```
ikopylov, line 60: Please, add `path` to the message (it will simplify problems diagnostics)

integration-tests/tests.py
```diff
@@ -0,0 +1,67 @@
...
+def run_tests(behaviour, args):
+    try:
+        print(f'Running bobp -b {str(behaviour)} {args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b {behaviour} {args.rstrip()}')).decode('ascii')
+        print(str(p))
+        if behaviour in {'put', 'get'}:
+            if f'{behaviour} errors:' in str(p) or f'panicked' in str(p):
+                sys.exit(f'{behaviour} test failed, see output')
+            if behaviour == 'put':
+                sleep(30)
```
ikopylov, line 17: 1. This function is not a good place to add any delays. In terms of code structure, this function is responsible for running a single test step. It is better to add delay at the higher level
2. 30 seconds is too much. Why this amount was chosen?

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+
+import subprocess, argparse, shlex, sys, os
+from time import sleep
+from python_on_whales import docker as d_cli
+from python_on_whales.exceptions import *
+from retry import *
+from bob_backend_timer import ensure_backend_up
+
+try:
+    bob_nodes_amount_string = os.environ['BOB_NODES_AMOUNT']
```
ikopylov, line 11: Pass nodes amount as a parameter to this script (like it is done in `tests.py`)

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+parser.add_argument('--mode', dest='mode', type=str, help='random or normal', choices=['random', 'normal'], default='normal')
+parser.add_argument('-k', dest='keysize', type=int, help='size of binary key (8 or 16)', choices=[8, 16], default=8)
+
+parsed_args = parser.parse_args()
+
+#get container object mapping to ports
+container_dict = {}
+try:
+    for i in range(int(bob_nodes_amount_string)):
+        port_num = str(20000+i)
```
ikopylov, line 41: Pass `min_port` as an argument (like it is done in `tests.py`)

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+        #run put
+        print(f'Running bobp -b put {bobp_args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b put {bobp_args.rstrip()}')).decode('ascii')
+        print(str(p))
+        if f'put errors:' in str(p) or f'panicked' in str(p):
+            sys.exit(f'Put test failed, see output.')
+        written_count += dict_args.get('-c')
+        #stops one
+        sleep(30)
+        d_cli.container.stop(container_dict[str(20000 + i)])
```
ikopylov, line 65: Use `min_port` arg

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+        print(f'Starting node on port {key}...')
+        d_cli.container.start(value)
+except DockerException as e:
+    sys.exit(e.stderr)
+
+try:
+    ensure_backend_up(int(bob_nodes_amount_string))
+except ValueError:
+    sys.exit('Amount of nodes has unexpected value.')
+
```
ikopylov, line 85: Add constant delay too, like it is done in `deploy.py`. Amount of delay can be passed as an argument

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+try:
+    dict_args = make_run_args(parsed_args, 0)
+    dict_args['-c'] = str(written_count)
+    bobp_args = args_to_str(dict_args)
+    print(f'Running bobp -b get {bobp_args.rstrip()}')
+    p = subprocess.check_output(shlex.split(f'./bobp -b get {bobp_args.rstrip()}')).decode('ascii')
+    print(str(p))
+    if f'get errors:' in str(p) or f'panicked' in str(p):
```
ikopylov, line 93: It is better to check for success, not for errors. Because if the output of `bobp` someday will change, the test start passing by even on errors. I suggest to search for `total err: 0`

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,97 @@
...
+    upper_bound = int(bob_nodes_amount_string)
+    for i in range(1, upper_bound):
+        #make correctly formatted args 
+        dict_args = make_run_args(parsed_args, written_count)
+        bobp_args = args_to_str(dict_args)
+        #run put
+        print(f'Running bobp -b put {bobp_args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b put {bobp_args.rstrip()}')).decode('ascii')
+        print(str(p))
+        if f'put errors:' in str(p) or f'panicked' in str(p):
```
ikopylov, line 60: Same here:
It is better to check for success, not for errors. Because if the output of `bobp` someday will change, the test start passing by even on errors. I suggest to search for `total err: 0`

integration-tests/tests.py
```diff
@@ -0,0 +1,67 @@
...
+
+run_options = ['put','get','exist']
+
+def run_tests(behaviour, args):
+    try:
+        print(f'Running bobp -b {str(behaviour)} {args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b {behaviour} {args.rstrip()}')).decode('ascii')
+        print(str(p))
+        if behaviour in {'put', 'get'}:
+            if f'{behaviour} errors:' in str(p) or f'panicked' in str(p):
```
ikopylov, line 14: Same here:
It is better to check for success, not for errors. Because if the output of `bobp` someday will change, the test start passing by even on errors. I suggest to search for `total err: 0`

integration-tests/tests_chain.py
```diff
@@ -0,0 +1,41 @@
...
+    print(str(p))
+    #find all of summaries in output
+    found_summaries = re.search(r'\bFinal\ssummary:\s[0-9]{1,}\/[0-9]{1,}\b', str(p))
+    #exit if no summaries
+    if not found_summaries:
+        sys.exit('No bobt output found.')
+    #find the last Summary and get its values
+    summary = found_summaries.group(0).replace('Final summary: ', '').split('/')
+    if summary[0] != summary[1]:
+        sys.exit('Test failed, captured summary has incomplete score.')
```
ikopylov, line 37: Please, write `summary[0]` and `summary[1]` values in the text of error

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        required: true
+        default: '10000'
+      bob_first_index:
+        description: 'Index of the first record to proceed during tests'
+        required: true
+        default: '0'
+      bob_bobt_count:
+        description: 'Amount of iterations for chain operation test'
+        required: true
+        default: '10000'
```
ikopylov, line 38: Count for `bobt` tests should be `~ 5 * (last_index - first_index)`.
```suggestion
        default: '100000'
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+      bob_bobt_first_index:
+        description: 'Bottom boundry for chain test'
+        required: true
+        default: '0'
+      bob_bobt_last_index:
+        description: 'Upper boundry for chain test'
+        required: true
+        default: '100000'
```
ikopylov, line 46: Count for `bobt` tests should be `~ 5 * (last_index - first_index)`.
```suggestion
        default: '10000'
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,331 @@
...
+        required: true
+        default: '2'
+      bob_payload_size:
+        description: 'Payload size'
+        required: true
+        default: '10240'
+      bob_record_count:
+        description: 'Amount of records to process during test'
+        required: true
+        default: '10000'
```
ikopylov, line 30:
```suggestion
        default: '100000'
```

.github/workflows/integration-tests.yml
ikopylov: Better to build with `test` profile

integration-tests/tests.py
Ziktorin: This is outdated sleep for testing the inconsistent get/exist results. I will remove it.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,338 @@
...
+jobs:
+  build-binaries:
+    if: ${{ github.event_name != 'pull_request_review' || github.event.review.state == 'approved' }}
+    strategy:
+      matrix:
+        key: [8, 16]
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
+      BUILD_MODE: "debug"
```
ikopylov, line 55:
```suggestion
      BUILD_MODE: "test"
```

.github/workflows/integration-tests.yml
ikopylov: Depends on https://github.com/qoollo/bob/issues/705

.github/workflows/integration-tests.yml
ikopylov: About Rust build profiles (https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles):
Rust has 4 default build profiles: `dev`, `release`, `test`, `bench`.
Additionally, Rust allows to declare custom profiles, but at the moment we don't have those.
Best profile for integration tests is `test` (it even has the name, indicating that).

There is no profile with name `debug` in Rust. Instead it has `dev` profile, that essentially the same.
That is the reason behind `debug` -> `dev` transformation.

Additionally, profile name affect the output directory for binaries. For all profiles, except `dev` the output directory equals to the profile name. But for `dev` profile, unfortunately, the directory is called `debug`. 
Example:
1. `release` -> `./target/x86_64-unknown-linux-musl/release/`
2. `test` -> `./target/x86_64-unknown-linux-musl/test`
3. `dev` -> `./target/x86_64-unknown-linux-musl/debug`

.github/workflows/integration-tests.yml
```diff
@@ -68,20 +68,28 @@ jobs:
...
           target: ${{ env.TARGET }}
           default: true
+      - name: Estimate output binaries directory
+        run: |
+          echo "BUILD_PROFILE_DIR=$(case "$BUILD_PROFILE" in\
+          ("dev") echo "debug";;\
+          ("test") echo "debug";;\
+          ("bench") echo "release";;\
+          (*) echo "$BUILD_PROFILE";;\
+          esac)" >> $GITHUB_ENV
```
ikopylov, line 78: Have you tested, that this syntax works properly?

ikopylov: Please, add on the next line command to print value of `BUILD_PROFILE_DIR`

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,346 @@
...
+      - name: Build binaries
+        uses: actions-rs/cargo@v1.0.1
+        with:
+          command: build
+          args: --target=${{ env.TARGET }} --profile=${{ env.BUILD_PROFILE }}
```
ikopylov, line 83: Why did you revert the profile name conversion? 
```suggestion
          args: --target=${{ env.TARGET }} --profile=${{ (env.BUILD_MODE == 'debug' && 'dev') || env.BUILD_MODE }}
```

bob-common/src/mapper.rs
```diff
@@ -7,11 +7,7 @@ use crate::{
     node::{Id as NodeId, Node},
 };
 use futures::{stream::FuturesUnordered, StreamExt};
-use std::{
```
ikopylov, line 10: Hm. You made something wrong. There should no be changes in the source code

CHANGELOG.md
```diff
@@ -6,7 +6,6 @@ Bob versions changelog
 #### Added
 - 'Exists' method through HEAD request in REST API (#691)
 - Support for 'exists' method in bobt (#691)
-- Round robin algorithm for alien nodes selection (#570)
```
ikopylov, line 9: Omg. Why did you revert this?

ikopylov:
> Revert "Merge branch 'master' into 518-implement-semi-automatic-integ
Reset this change. It reverts the code changes

.github/workflows/integration-tests.yml
Ziktorin: This convertion is not needed anymore. We set build profile and estimate build profile directory on previous steps, so there is no need to transform anything.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+      bob_record_count:
+        description: 'Amount of records to process during test'
+        required: true
+        default: '100000'
+      bob_first_index:
+        description: 'Index of the first record to proceed during tests'
+        required: true
+        default: '0'
+      bob_bobt_count:
+        description: 'Amount of iterations for chain operation test'
```
ikopylov, line 36:
```suggestion
        description: 'Amount of iterations for bobt test'
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+        required: true
+        default: '100000'
+      bob_bobt_first_index:
+        description: 'Bottom boundry for chain test'
```
ikopylov, line 40:
```suggestion
        description: 'Bottom boundary for key in bobt test'
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+        required: true
+        default: '0'
+      bob_bobt_last_index:
+        description: 'Upper boundry for chain test'
```
ikopylov, line 44:
```suggestion
        description: 'Upper boundary for key in bobt test'
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+        default: '10000'
+jobs:
+  build-binaries:
+    if: ${{ github.event_name != 'pull_request_review' || github.event.review.state == 'approved' }}
+    strategy:
+      matrix:
+        key: [8, 16]
+    env:
+      TARGET: "x86_64-unknown-linux-musl"
+      BUILD_PROFILE: "test"
```
ikopylov, line 55: I have added `integration-test` profile here: https://github.com/qoollo/bob/pull/713
Please, merge master and update build profile to `integration-test`

integration-tests/tests_aliens.py
```diff
@@ -0,0 +1,96 @@
...
+        bobp_args = args_to_str(dict_args)
+        #run put
+        print(f'Running bobp -b put {bobp_args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b put {bobp_args.rstrip()}')).decode('ascii')
+        print(str(p))
+        if not 'total err: 0' in str(p):
+            sys.exit(f'Put test failed, see output.')
+        written_count += dict_args.get('-c')
+        #stops one
+        sleep(30)
```
ikopylov, line 61: Reduce this sleep to 10 seconds

.github/workflows/integration-tests.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           ("dev") echo "debug";;\
           ("test") echo "debug";;\
           ("bench") echo "release";;\
+          ("integration-test") echo "debug";;\
```
ikopylov, line 77: You don't need to change the conversion logic. `integration-test` builds into the folder with the same name as the name of profile

ikopylov (approved):
LGTM

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+          INPUT="${{ github.event.inputs.bob_nodes_amount }}"
+          echo "BOB_NODES_AMOUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_vdisk_count }}"
+          echo "BOB_VDISK_COUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_replicas_amount }}"
+          echo "BOB_REPLICAS_AMOUNT=${INPUT:-"2"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_payload_size }}"
+          echo "BOB_PAYLOAD_SIZE=${INPUT:-"10240"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_record_count }}"
+          echo "BOB_RECORD_COUNT=${INPUT:-"10000"}" >> $GITHUB_ENV
```
ikopylov, line 165:
```suggestion
          echo "BOB_RECORD_COUNT=${INPUT:-"100000"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
```
ikopylov, line 235:
```suggestion
          echo "BOB_RECORD_COUNT=${INPUT:-"20000"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+      - name: Setup default variables for auto runs
+        run: |
+          INPUT="${{ github.event.inputs.bob_nodes_amount }}"
+          echo "BOB_NODES_AMOUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_vdisk_count }}"
+          echo "BOB_VDISK_COUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_replicas_amount }}"
+          echo "BOB_REPLICAS_AMOUNT=${INPUT:-"2"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_payload_size }}"
+          echo "BOB_PAYLOAD_SIZE=${INPUT:-"10240"}" >> $GITHUB_ENV
```
ikopylov, line 233:
```suggestion
          echo "BOB_PAYLOAD_SIZE=${INPUT:-"4096"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+          INPUT="${{ github.event.inputs.bob_nodes_amount }}"
+          echo "BOB_NODES_AMOUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_vdisk_count }}"
+          echo "BOB_VDISK_COUNT=${INPUT:-"4"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_replicas_amount }}"
+          echo "BOB_REPLICAS_AMOUNT=${INPUT:-"2"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_bobt_first_index }}"
+          echo "BOB_BOBT_FIRST_INDEX=${INPUT:-"0"}" >> $GITHUB_ENV
+          INPUT="${{ github.event.inputs.bob_bobt_last_index }}"
+          echo "BOB_BOBT_LAST_INDEX=${INPUT:-"100000"}" >> $GITHUB_ENV
```
ikopylov, line 305:
```suggestion
          echo "BOB_BOBT_LAST_INDEX=${INPUT:-"10000"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
...
+          INPUT="${{ github.event.inputs.bob_bobt_count }}"
+          echo "BOB_BOBT_COUNT=${INPUT:-"10000"}" >> $GITHUB_ENV
```
ikopylov, line 307:
```suggestion
          echo "BOB_BOBT_COUNT=${INPUT:-"100000"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
ikopylov, line 233:
```suggestion
          echo "BOB_PAYLOAD_SIZE=${INPUT:-"4096"}" >> $GITHUB_ENV
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,348 @@
+name: integration-tests
+on:
+  push:
+    branches: [ "518-implement-semi-automatic-integration-tests", "642-integration-tests-add-tests-for-aliens", "648-integrationtests-add-consistency-test-for-random-putgetdelete-requests-with-bobt-tool" ] #REMOVE WHEN ON MASTER
```
ikopylov, line 4: Remove this

ikopylov (approved):
LGTM

ikopylov (changes requested):
Test reject

ikopylov:
Test comment

integration-tests/Templates/compose_template.yml.j2
```diff
@@ -0,0 +1,24 @@
+version: "3.7"
```
ikopylov, line 1: Test comment

ikopylov (commented):
Test review

ikopylov (approved):
LGTM

PR #535, opened 2022-07-14
408 user auth claims

Boneyan:
User's role is now optional. Claims can be set in user description, claims are also optional. If both claims and role are set, then claims will overwrite role's permissions
Closes https://github.com/qoollo/bob/issues/408


ikopylov (approved):
LGTM

PR #536, opened 2022-07-14
304 password hash ability

Boneyan:
SHA512 hash of password with salt 'bob' can be specified instead of password
Closes https://github.com/qoollo/bob/issues/304


bob-access/src/authenticator/users_storage/config.rs
```diff
@@ -1,20 +1,40 @@
...
     pub(super) role: String,
 }
 
 #[derive(Debug, Deserialize)]
-pub(super) struct ConfigUsers {
+pub struct ConfigUsers {
     pub(super) roles: HashMap<String, Perms>,
     pub(super) users: Vec<ConfigUser>,
+    #[serde(default = "ConfigUsers::default_password_salt")]
+    pub password_salt: String,
```
ikopylov, line 23: `pub(super)`?

Boneyan: pub(super) limits field to be visible only inside /user_storage, so in case of password_salt it needs to be 'pub' (or can I make the field private and create getter method?)

ikopylov: I just noticed, that you started using `ConfigUsers` outside of the current module. This is a bad design and should be reverted. You can store `password_salt` inside `UsersMap` struct and add getter to that struct

bob-access/src/authenticator/users_storage/config.rs
```diff
@@ -26,7 +37,8 @@ pub(super) fn parse_users(
         let &perms = roles
             .get(&u.role)
             .ok_or_else(|| Error::Validation(format!("Can't find role {}", u.role)))?;
-        let user = User::new(u.username.clone(), u.password, perms);
+        let hash = u.password_hash.map(|h| Vec::from_hex(h).expect("Invalid sha512 hash"));
```
ikopylov, line 40: Add additional check, that `password_hash` has 64 bytes size (512 / 8)

Boneyan: Code already will panic if hash's length is not 512 bits

ikopylov (approved):
LGTM

PR #543, opened 2022-07-19
437 support default user

### LINKED ISSUE #437: Authorization: support special user `default`
If `default` user is listed in `users.yaml` file, then all requests without auth data in header should be associated with it

Boneyan:
If no credentials provided and default user specified in config, then request will have default user permissions
Closes #437


bob-access/src/extractor.rs
```diff
@@ -40,23 +40,26 @@ impl<T: Extractor> ExtractorExt for T {
...
-            }
+            },
+            (None, None) => {
+                if let Some(node_name) = self.get_header("node_name")? {
+                    let creds = builder
+                        .with_nodename(node_name)
+                        .build();
+                    return Ok(creds);
+                }
+                let creds = builder
```
ikopylov, line 57: Please, add comment here: `Fallback to special "default" user, when no credentials were provided`

ikopylov:
Please, resolve merge conflicts

ikopylov (approved):
LGTM

PR #545, opened 2022-07-26
363 improve deletion performance

### LINKED ISSUE #363: Single record deletion performance improvements
Deleting a record requires rebuilding the index file. Currently, the index file is overwritten immediately. It is better to load the index file into memory, update it and leave it on until the background dumping logic dumps it

idruzhitskiy:
Closes #363

### DIFF
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -255,7 +255,7 @@ impl BackendStorage for Pearl {
         let mut min_modification = u64::MAX;
         for dc in self.disk_controllers.iter() {
             if let Some(holder) = dc.find_least_modified_freeable_holder().await {
-                let modification = holder.last_modification().await;
+                let modification = holder.last_modification();
                 if modification < min_modification {
                     least_modified = Some(holder);
                     min_modification = modification;
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -101,7 +101,7 @@ impl MetricsProducer for Pearl {
             .iter()
             .chain(once(&self.alien_disk_controller))
             .cloned()
-            .map(|dc| async move { dc.index_memory().await })
+            .map(|dc| async move { dc.active_index_memory().await })
             .collect();
         let cnt = futs.fold(0, |cnt, dc_cnt| ready(cnt + dc_cnt)).await;
         cnt
@@ -242,7 +242,7 @@ impl BackendStorage for Pearl {
         }
 
         if let Some(h) = oldest {
-            let memory = h.index_memory().await;
+            let memory = h.active_index_memory().await;
             h.close_active_blob().await;
             Some(memory)
         } else {
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -114,6 +114,21 @@ impl MetricsProducer for Pearl {
         }
         cnt
     }
+
+    async fn disk_used_by_disk(&self) -> HashMap<DiskPath, u64> {
+        let futs: FuturesUnordered<_> = self
+            .disk_controllers
+            .iter()
+            .chain(once(&self.alien_disk_controller))
+            .cloned()
+            .map(|dc| async move { (dc.disk().clone(), dc.disk_used().await) })
+            .collect();
+        futs.fold(HashMap::new(), |mut m, (n, u)| {
+            m.entry(n).and_modify(|s| *s = *s + u).or_insert(u);
+            ready(m)
+        })
+        .await
+    }
 }
 
 #[async_trait]
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -250,6 +250,28 @@ impl BackendStorage for Pearl {
         }
     }
 
+    async fn free_least_used_resources(&self) -> Option<usize> {
+        let mut least_modified: Option<Holder> = None;
+        let mut min_modification = u64::MAX;
+        for dc in self.disk_controllers.iter() {
+            if let Some(holder) = dc.find_least_modified_freeable_holder().await {
+                let modification = holder.last_modification().await;
+                if modification < min_modification {
+                    least_modified = Some(holder);
+                    min_modification = modification;
+                }
+            }
+        }
+
+        if let Some(h) = least_modified {
+            let memory = h.freeable_resources_memory().await;
+            h.free_freeable_resources().await;
+            Some(memory)
+        } else {
+            None
+        }
+    }
+
     async fn delete(&self, op: Operation, key: BobKey) -> Result<u64, Error> {
         debug!("DELETE[{}] from pearl backend. operation: {:?}", key, op);
         let dc_option = self
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -264,9 +264,8 @@ impl BackendStorage for Pearl {
         }
 
         if let Some(h) = least_modified {
-            let memory = h.index_memory().await;
-            h.free_excess_resources().await;
-            Some(memory)
+            let freed = h.free_excess_resources().await;
+            Some(freed)
         } else {
             None
         }
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -101,7 +101,7 @@ impl MetricsProducer for Pearl {
             .iter()
             .chain(once(&self.alien_disk_controller))
             .cloned()
-            .map(|dc| async move { dc.active_index_memory().await })
+            .map(|dc| async move { dc.index_memory().await })
             .collect();
         let cnt = futs.fold(0, |cnt, dc_cnt| ready(cnt + dc_cnt)).await;
         cnt
@@ -264,8 +264,8 @@ impl BackendStorage for Pearl {
         }
 
         if let Some(h) = least_modified {
-            let memory = h.freeable_resources_memory().await;
-            h.free_freeable_resources().await;
+            let memory = h.index_memory().await;
+            h.free_excess_resources().await;
             Some(memory)
         } else {
             None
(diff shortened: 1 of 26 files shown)


bob/src/cleaner.rs
```diff
@@ -22,27 +24,74 @@ impl Cleaner {
...
+                    if *lock {
+                        *lock = false;
+                        should_clean_up = true;
+                    }
+                }
+                if should_clean_up {
+                    let mut memory = backend.index_memory().await;
+                    info!("Memory before closing old active blobs: {:?}", memory);
+                    while memory > limit {
+                        if let Some(freed) = backend.close_oldest_active_blob().await {
```
ikopylov, line 75: As far as I remember, we track the time of the last access to the pearl storage. If this is true, then it would be a good to implement `close_least_recently_used_blob` method and use it here. Closing the least recently used storages is better suited for deletion scenario

bob/src/cleaner.rs
```diff
@@ -22,27 +24,74 @@ impl Cleaner {
...
+                let mut should_clean_up = false;
+                {
+                    let mut lock = cleaner.index_cleanup_requested.lock().await;
+                    if *lock {
+                        *lock = false;
+                        should_clean_up = true;
+                    }
+                }
+                if should_clean_up {
+                    let mut memory = backend.index_memory().await;
```
ikopylov, line 72: I think, it would be good to have 2 memory limits: soft and hard. Here the hard limit fits best, and soft limit is better for usual clean-up task. 
From the config perspective, the current `index_memory_limit` is basically a hard limit. Additional new parameter (`index_memory_limit_soft`) should be added. This new parameter should be `Optional` and by default calculated as `0.9 * index_memory_limit`

bob/src/cleaner.rs
```diff
@@ -22,27 +24,74 @@ impl Cleaner {
...
+        t: Duration,
+        index_memory_limit: Option<usize>,
+    ) {
+        let mut interval = interval(t);
+        loop {
+            interval.tick().await;
+            if let Some(limit) = index_memory_limit {
+                let mut should_clean_up = false;
+                {
+                    let mut lock = cleaner.index_cleanup_requested.lock().await;
```
ikopylov, line 65: `Notify` is better for this scenario: https://docs.rs/tokio/1.9.0/tokio/sync/struct.Notify.html
`notify_waiters` can be used to unlock the task: https://docs.rs/tokio/1.9.0/tokio/sync/struct.Notify.html#method.notify_waiters
After the notification is received, an additional small delay can be added to prevent immediate cleanup

bob/src/cleaner.rs
```diff
@@ -52,19 +101,7 @@ impl Cleaner {
...
-                    if let Some(freed) = backend.close_oldest_active_blob().await {
-                        memory = memory - freed;
-                        debug!("closed index, freeing {:?} bytes", freed);
-                    } else {
-                        break;
-                    }
-                }
-                info!("Memory after closing old active blobs: {:?}", memory);
-            }
+            cleaner.request_index_cleanup().await;
```
ikopylov, line 104: It is better to execute the code of clean-up here directly, and its execution should be synced with `fast_cleaner_task` with `Mutex`. And obviously, `close_unneeded_active_blobs` should also be protected by that mutex

bob/src/cleaner.rs
```diff
@@ -52,19 +101,7 @@ impl Cleaner {
...
             if let Some(limit) = bloom_filter_memory_limit {
                 backend.offload_old_filters(limit).await;
```
ikopylov, line 106: An additional question to investigate: how does loading/offloading index affect bloom filters? Does index loading restore offloaded bloom filter? Will this restored bloom filter show up in the bloom filter hierarchy and be unloaded by calling `offload_old_filters`?

bob/src/cleaner.rs
```diff
@@ -22,27 +24,74 @@ impl Cleaner {
...
+                    info!("Memory before closing old active blobs: {:?}", memory);
+                    while memory > limit {
+                        if let Some(freed) = backend.close_oldest_active_blob().await {
+                            memory = memory - freed;
+                            debug!("closed index, freeing {:?} bytes", freed);
+                        } else {
+                            break;
+                        }
+                    }
+                    info!("Memory after closing old active blobs: {:?}", memory);
```
ikopylov, line 82: It is better to write here both memory before and memory after unloading

bob/src/cleaner.rs
```diff
@@ -22,27 +24,74 @@ impl Cleaner {
...
+                {
+                    let mut lock = cleaner.index_cleanup_requested.lock().await;
+                    if *lock {
+                        *lock = false;
+                        should_clean_up = true;
+                    }
+                }
+                if should_clean_up {
+                    let mut memory = backend.index_memory().await;
+                    info!("Memory before closing old active blobs: {:?}", memory);
```
ikopylov, line 73: It is better to write this message on `debug` level

bob/src/cleaner.rs
ikopylov: Additional question:  `close_oldest_active_blob` probably only handles storages with active blob, but indexes loading don't do anything with active blobs, and if I'm right, this function won't offload indexes as expected

bob/src/cleaner.rs
idruzhitskiy: 1. Index loading restore offloaded bloom filter from disk.
2. Yes, they will be freed as filters cleanup work on all blobs, not only active.

bob/src/cleaner.rs
idruzhitskiy: Now we have separate functions for closing all indexes in least modified holder

bob-backend/src/pearl/core.rs
```diff
@@ -101,7 +101,7 @@ impl MetricsProducer for Pearl {
             .iter()
             .chain(once(&self.alien_disk_controller))
             .cloned()
-            .map(|dc| async move { dc.index_memory().await })
+            .map(|dc| async move { dc.active_index_memory().await })
```
ikopylov, line 104: For metrics we should report full memory occupied by indixes (in active and in inactive blobs)

bob-backend/src/pearl/core.rs
```diff
@@ -242,14 +242,36 @@ impl BackendStorage for Pearl {
...
+                if modification < min_modification {
+                    least_modified = Some(holder);
+                    min_modification = modification;
+                }
+            }
+        }
+
+        if let Some(h) = least_modified {
+            let memory = h.freeable_resources_memory().await;
+            h.free_freeable_resources().await;
```
ikopylov, line 268: This function can return the amount of freed memory. In this way the separate function `freeable_resources_memory` is not needed

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -538,14 +538,14 @@ impl DiskController {
         cnt
     }
 
-    pub(crate) async fn index_memory(&self) -> usize {
+    pub(crate) async fn active_index_memory(&self) -> usize {
```
ikopylov, line 541: Looks like this function should return full memory occupied by indexes, because it is used only in metrics

bob-backend/src/pearl/holder.rs
```diff
@@ -79,9 +79,14 @@ impl Holder {
         storage.blobs_count().await
     }
 
-    pub async fn index_memory(&self) -> usize {
+    pub async fn active_index_memory(&self) -> usize {
```
ikopylov, line 82: Looks like full index memory should be returned here

bob-backend/src/pearl/holder.rs
```diff
@@ -79,9 +79,14 @@ impl Holder {
...
         let storage = self.storage.read().await;
-        storage.index_memory().await
+        storage.active_index_memory().await
+    }
+
+    pub async fn freeable_resources_memory(&self) -> usize {
```
ikopylov, line 87: As I see, we only need to know the fact, that holder has some freeable resources (the size of occupied memory is not needed). Thus I suggest to replace this function with `have_excess_resources` that returns `bool`

bob-backend/src/pearl/holder.rs
```diff
@@ -25,7 +25,7 @@ pub struct Holder {
     disk_path: PathBuf,
     config: PearlConfig,
     storage: Arc<RwLock<PearlSync>>,
-    last_write_ts: Arc<RwLock<u64>>,
+    last_modification: Arc<RwLock<u64>>,
```
ikopylov, line 28: `RwLock` here protects only a variable, so there is no need to use async version of it. I suggest to use sync `RwLock` from `parking_lot` crate (https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html). It will be more efficient

bob-backend/src/pearl/holder.rs
```diff
@@ -79,11 +81,20 @@ impl Holder {
...
+        storage.active_index_memory().await
+    }
+
     pub async fn index_memory(&self) -> usize {
         let storage = self.storage.read().await;
         storage.index_memory().await
     }
 
+    pub async fn has_excess_resources(&self) -> bool {
+        self.index_memory().await > 0
```
ikopylov, line 95: I think we should not treat active blob indexes as excessive, thus it is better to use `inactive_index_memory` function here

bob/src/cleaner.rs
```diff
@@ -50,24 +99,41 @@ impl Cleaner {
...
+                index_cleanup(&backend, limit).await;
             }
             if let Some(limit) = bloom_filter_memory_limit {
                 backend.offload_old_filters(limit).await;
             }
         }
     }
 }
+
+async fn old_index_cleanup(backend: &Arc<Backend>, soft: usize, hard: usize) {
```
ikopylov, line 116: `index_cleanup_by_count_limit`

bob/src/cleaner.rs
```diff
@@ -50,24 +99,41 @@ impl Cleaner {
...
+    backend.close_unneeded_active_blobs(soft, hard).await;
+}
+
+async fn index_cleanup(backend: &Arc<Backend>, limit: usize) {
```
ikopylov, line 120: `index_cleanup_by_memory_limit`

ikopylov (approved):
LGTM

PR #546, opened 2022-07-30
540 auth error convertion

### LINKED ISSUE #540: Authorization error is not converted to an internal error struct
Conversion code is missed: https://github.com/qoollo/bob/blob/master/bob-common/src/error.rs#L143

Boneyan:
Added conversion of "unauthorized" status into internal error 
Closes #540

### DIFF
--- a/bob-common/src/error.rs
+++ b/bob-common/src/error.rs
@@ -156,6 +156,7 @@ impl From<Status> for Error {
                 "Failed" => Some(Self::failed(rest_words(words, length))),
                 "Internal" => Some(Self::internal()),
                 "PearlChangeState" => Some(Self::pearl_change_state(rest_words(words, length))),
+                "Unauthorized" => Some(Self::internal()),
                 _ => None,
             },
         }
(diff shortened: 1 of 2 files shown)


bob-common/src/error.rs
```diff
@@ -156,6 +156,7 @@ impl From<Status> for Error {
                 "Failed" => Some(Self::failed(rest_words(words, length))),
                 "Internal" => Some(Self::internal()),
                 "PearlChangeState" => Some(Self::pearl_change_state(rest_words(words, length))),
+                "Unauthorized" => Some(Self::internal()),
```
ikopylov, line 159: Seems like `internal` is a bad choice. Probably, separate error kind should be used for authorization errors

ikopylov:
@Boneyan please, run the test when `Unauthorized` error generated during the work, and check that everything works normally (no panic, data is written to aliens when this error occured)

ikopylov (approved):
LGTM
Need a test

Boneyan:
I ran the test, all works fine

Issue #529, opened 2022-07-08
Authorization: inter-node authentication does not work in Docker Swarm


ikopylov:

ikopylov:
Panic: `thread 'main' panicked at 'failed to resolve hostname', bob-apps/bin/bobd.rs:167:21`

PR #547, opened 2022-07-30
530 internode auth fix

### LINKED ISSUE #530: Authorization: inter-node authentication does not work for some nodes in cluster

Boneyan:
Fixed internode authorization error
Closes #530


bob-common/src/node.rs
```diff
@@ -82,12 +82,12 @@ impl Node {
         self.conn.read().await.clone()
     }
 
-    pub async fn check(&self, client_fatory: &Factory) -> Result<(), String> {
+    pub async fn check(&self, client_factory: &Factory, local_node_name: &String) -> Result<(), String> {
```
ikopylov, line 85: This change seems to break the idea of factory pattern. `local_node_name` should be part of `Factory`, not the separate argument

ikopylov:
@Boneyan Please, check whether these changes fixed the same bug in Docker Swarm (https://github.com/qoollo/bob/issues/529)

ikopylov (approved):
LGTM 
Need a test inside Swarm

PR #554, opened 2022-08-06
548 internode auth same machine

### LINKED ISSUE #548: Authorization: fix inter-node auth when several nodes are on the same machine (have same ip address)

Boneyan:
Internode auth works properly with nodes on same ip
Closes #548


ikopylov (approved):
LGTM

PR #583, opened 2022-08-25
581 print full error in exist function

### LINKED ISSUE #581: Print full error information and propagate it to the caller in `exist` function
Currently, there are many errors `blob index get any failed` in log. We should see details
https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/holder.rs#L309

ikopylov:
Closes #581


idruzhitskiy (commented):
LGTM

PR #613, opened 2022-09-08
591 avoid unnecessary Stopwatch::start_new

### LINKED ISSUE #591: Avoid Stopwatch start_new if log level is higher than trace
This is a micro optimization.
There are several places, where Stopwatch is used and the time measurement goes only to logs on trace level.
Examples:
https://github.com/qoollo/bob/blob/master/bob/src/server.rs#L112
https://github.com/qoollo/bob/blob/master/bob/src/grinder.rs#L82
https://github.com/qoollo/bob/blob/master/bob-backend/src/core.rs#L242

In this places we can avoid Stopwatch start if log level is higher than trace

lostl1ght:
Closes #591

### DIFF
--- a/bob/src/server.rs
+++ b/bob/src/server.rs
@@ -171,7 +171,10 @@ where
             return Err(Status::permission_denied("READ permission required"));
         }
         trace!("- - - - - SERVER GET START - - - - -");
-        let sw = Stopwatch::start_new();
+        let mut sw = Stopwatch::new();
+        if log_enabled!(log::Level::Debug) {
+            sw.start()
+        }
         trace!(
             "process incoming get request /{:.3}ms/",
             sw.elapsed().as_secs_f64() * 1000.0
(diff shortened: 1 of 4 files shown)


idruzhitskiy (approved):

ikopylov:
I think we don't really need this micro-optimization. Please revert everything except changes where `sw.elapsed()` was evaluated outside of the logging functions. And update the description in the CHANGELOG accordingly

bob/src/server.rs
```diff
@@ -200,8 +203,7 @@ where
                 "grinder finished request processing /{:.3}ms/",
                 sw.elapsed().as_secs_f64() * 1000.0
             );
-            let elapsed = sw.elapsed_ms();
-            debug!("GET[{}]-OK dt: {}ms", key, elapsed);
+            debug!("GET[{}]-OK dt: {}ms", key, sw.elapsed_ms());
```
ikopylov, line 206: Keep this change

bob/src/server.rs
```diff
@@ -236,8 +241,7 @@ where
             .exist(&keys, &options)
             .await
             .map_err::<Status, _>(|e| e.into())?;
-        let elapsed = sw.elapsed();
-        debug!("EXISTS-OK dt: {:?}", elapsed);
+        debug!("EXISTS-OK dt: {:?}", sw.elapsed());
```
ikopylov, line 244: And this one too

ikopylov:
This optimization is not significant

PR #615, opened 2022-09-08
595 special handling of a single holder result

### LINKED ISSUE #595: Special handling of a single holder result in `Group::find_actual_holder`
This is a micro optimization.
Single holder is the most common result, so it is good to handle it in a special way to avoid `sort_by_key` invocation.
https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/group.rs#L147

lostl1ght:
Closes #595

### DIFF
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -153,7 +153,9 @@ impl Group {
             .collect();
 
         if !holders_for_time.is_empty() {
-            holders_for_time.sort_by_key(|h| h.1.start_timestamp());
+            if holders_for_time.len() > 1 {
+                holders_for_time.sort_by_key(|h| h.1.start_timestamp());
+            }
             let (id, holder) = holders_for_time.pop().expect("not empty");
             Ok((id, holder.clone()))
         } else {
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -146,18 +146,18 @@ impl Group {
     // find in all pearls actual pearl
     async fn find_actual_holder(&self, data: &BobData) -> BackendResult<(ChildId, Holder)> {
         let holders = self.holders.read().await;
-        let mut holders_for_time: Vec<(usize, &Holder)> = holders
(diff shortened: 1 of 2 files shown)


bob-backend/src/pearl/group.rs
```diff
@@ -153,7 +153,9 @@ impl Group {
             .collect();
 
         if !holders_for_time.is_empty() {
-            holders_for_time.sort_by_key(|h| h.1.start_timestamp());
+            if holders_for_time.len() > 1 {
+                holders_for_time.sort_by_key(|h| h.1.start_timestamp());
+            }
             let (id, holder) = holders_for_time.pop().expect("not empty");
```
ikopylov, line 159: `pop` modifies vector. I suggest to directly access the last element instead. This will avoid unnecessary modification

bob-backend/src/pearl/group.rs
```diff
@@ -153,7 +153,9 @@ impl Group {
             .collect();
```
ikopylov, line 153: I think, we should go further and optimize away this vector creation. `holders` can be iterated with regular `for` loop and during this iteration we can find easily the one that fits into interval (`gets_into_interval(...)` is true) and has maximum start timestamp (`start_timestamp()`)

bob-backend/src/pearl/group.rs
```diff
@@ -146,16 +146,18 @@ impl Group {
...
             .enumerate()
-            .filter(|h| h.1.gets_into_interval(data.meta().timestamp()))
-            .collect();
+            .filter(|h| h.1.gets_into_interval(data.meta().timestamp()));
 
-        if !holders_for_time.is_empty() {
-            holders_for_time.sort_by_key(|h| h.1.start_timestamp());
-            let (id, holder) = holders_for_time.pop().expect("not empty");
-            Ok((id, holder.clone()))
+        if let Some(mut max) = holders_for_time.next() {
```
ikopylov, line 154: Iteration performed two times: first time here when calling `next` and the second time below in for loop. It is better to do it once with just a for loop

PR #622, opened 2022-09-26
529 internode auth in swarm fix

### LINKED ISSUE #529: Authorization: inter-node authentication does not work in Docker Swarm

Boneyan:
Hostname resolving in background tasks
Closes #529


bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,154 @@
...
+
+    fn check_node_request(&self, node_name: &String, ip: Option<SocketAddr>) -> bool {
+        let mut result = false;
+        let mut unresolved = false;
+        {
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            if nodes.is_empty() {
+                warn!("nodes credentials not set");
+            }
+            if ip.is_none() {
```
idruzhitskiy, line 124: I think you can check this outside lock

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,154 @@
...
+        cred.map(|cred| self.spawn_resolver(cred));
+    }
+
+    fn check_node_request(&self, node_name: &String, ip: Option<SocketAddr>) -> bool {
+        let mut result = false;
+        let mut unresolved = false;
+        {
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            if nodes.is_empty() {
+                warn!("nodes credentials not set");
```
idruzhitskiy, line 122: If it's empty most of the function will not be invoked, I think it's better to return here

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,154 @@
...
+            }
+            let ip = ip.unwrap().ip();
+            nodes
+                .get(node_name)
+                .map(|cred| {
                     if let Some(CredentialsKind::InterNode(other_name)) = cred.kind() {
                         if node_name == other_name {
-                            return true;
+                            cred.ip().as_ref().map(|ips| {
+                                for cred_ip in ips {
```
idruzhitskiy, line 134: I think you can use `result = cred.ip().as_ref().and_then(|ips| ips.iter().find(...)).is_some()`

idruzhitskiy: And then, I guess, you don't need two flags

idruzhitskiy (approved):
LGTM

bob-access/Cargo.toml
```diff
@@ -17,4 +17,9 @@ serde = {version = "1.0", features = ["derive"] }
 jsonwebtoken = "7.2"
 serde_yaml = "0.8"
 serde_json = "1.0"
-futures = "0.3"
\ No newline at end of file
+futures = "0.3"
+
+[dependencies.tokio]
+version = "1.14"
```
ikopylov, line 23: 1.21
In this PR create versions are updated in other places: https://github.com/qoollo/bob/pull/626

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,148 @@
...
+                Ok(address) => Some(address.collect()),
+                _ => None
+            };
+            if let Some(addr) = addr.as_ref() {
+                if addr.len() > 0 {
+                    break;
+                }
+            }
+
+            sleep(Duration::from_secs(1));
```
ikopylov, line 88: You should use `async` variant of `sleep` from tokio

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,148 @@
...
         }
-        self.nodes
-            .get(ip.as_ref()?)
-            .map(|creds| {
-                for cred in creds {
-                    if let Some(CredentialsKind::InterNode(other_name)) = cred.kind() {
-                        if node_name == other_name {
+
+        while addr.is_none() || (addr.is_some() && addr.as_ref().unwrap().len() == 0) {
+            sleep(Duration::from_millis(sleep_dur_ms));
```
ikopylov, line 92: Same here

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,148 @@
...
+            .all(|cred|
+                cred.hostname().is_some() &&
+                cred.kind().map(|k| k.is_internode()) == Some(true))
+    }
+
     pub fn set_nodes_credentials(
         &mut self,
-        nodes: HashMap<IpAddr, Vec<Credentials>>,
+        nodes: HashMap<String, Credentials>,
+        unresolved: Vec<Credentials>,
```
ikopylov, line 52: Why do you split nodes with `ip` and with host name in arguments. IMHO, this function should received them without distinction. Inside this function you can split them if it is required

bob-access/src/credentials.rs
```diff
@@ -64,8 +65,20 @@ impl Credentials {
...
-        Some(self.address?.ip())
+    pub fn ip(&self) -> &Option<Vec<SocketAddr>> {
+        &self.address
+    }
+
+    pub fn set_addresses(&mut self, addresses: Vec<SocketAddr>) {
+        self.address = Some(addresses);
+    }
+
+    pub fn single_ip(&self) -> Option<SocketAddr> {
```
ikopylov, line 76: You must either use all IP addresses or store only one inside this struct. Storing all IPs and use only one is not consistent

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,141 @@
...
-            .get(ip.as_ref()?)
-            .map(|creds| {
-                for cred in creds {
-                    if let Some(CredentialsKind::InterNode(other_name)) = cred.kind() {
-                        if node_name == other_name {
+    }
+
+    fn mark_unresolved(&self, nodename: &str) {
+        let mut nodes = self.nodes.write().expect("nodes credentials lock");
+        let cred = nodes.remove(nodename);
```
ikopylov, line 105: From the common logic perspective it is safer not to remove the node from the list. It is better just to update the value and mark it as having unresolved hostname

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,141 @@
...
+            if let Some(addr) = addr.as_ref() {
+                if addr.len() > 0 {
+                    break;
+                }
+            }
+
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+
+        while addr.is_none() || (addr.is_some() && addr.as_ref().unwrap().len() == 0) {
```
ikopylov, line 84: You have 2 loops that do the same thing with different sleep interval. You should keep only one.
Sleep duration update strategy can be improved: you can start with 100ms duration and on every iteration multiply it by 2, but check that it will not become larger than `sleep_dur_ms` (`sleep_dur_ms.min(cur_sleep_dur * 2)`)

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,141 @@
...
+        {
+            if ip.is_none() {
+                return false;
+            }
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            if nodes.is_empty() {
+                warn!("nodes credentials not set");
+                return false;
+            }
+            let ip = ip.unwrap().ip();
```
ikopylov, line 120: You can do this right after you checked that ip is not None (`if ip.is_none() {`)

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,141 @@
...
+            if let Some(cred) = nodes.get(node_name) {
+                if let Some(CredentialsKind::InterNode(other_name)) = cred.kind() {
+                    if node_name == other_name {
```
ikopylov, line 123: Is it possible at all? Looks like this can be checked in `debug_assert`

bob-access/src/authenticator/basic.rs
ikopylov: Also, deletion is not recommended because requests may come from a fake node with the wrong address. If we delete immediately, then other requests from the real node will stop coming. This opens the door for DDoS

bob-access/src/authenticator/basic.rs
```diff
@@ -1,62 +1,141 @@
...
-            })
+            }
+        }
+        if unresolved {
+            self.mark_unresolved(node_name);
+        }
+        false
     }
 
     fn check_credentials_common(&self, credentials: Credentials) -> Result<Permissions, Error> {
```
ikopylov, line 141: Maybe it is better to pass `Credentials` by reference everywhere? Because, `Credentials` is pretty big struct

bob-access/src/authenticator/basic.rs
ikopylov: Also, to prevent DDoS it would be good to start re-resolving not on the first request with unknown ip, but on the sequence of e.g. 3 such requests (without any requests from correct ip)

bob-access/src/credentials.rs
```diff
@@ -11,7 +11,8 @@ use crate::{error::Error, extractor::ExtractorExt, Authenticator};
 
 #[derive(Debug, Default, Clone)]
 pub struct Credentials {
```
ikopylov, line 13: This struct solving 2 separate tasks: 
1. It stores Credentials (`CredentialsKind`) and associated info (`hostname` and `Vec<SocketAddr>`) from config to compare with
2. It stores Credentials (`CredentialsKind`) and associated info (`SocketAddr`) received from request that compares with the first one

Previously, the stored data was the same, and we used the same struct for 2 tasks. Currently, the stored data are different, so the struct should also be split into 2 different ones.

The first one can be named `DeclaredCredentials`, the second one - `RequestCredentials`

ikopylov (changes requested):
Changes are required

bob-access/src/credentials.rs
```diff
@@ -77,44 +153,96 @@ impl Credentials {
...
     }
 
-    pub fn with_token(&mut self, token: impl Into<String>) -> &mut Self {
+    pub fn with_token(mut self, token: impl Into<String>) -> Self {
         self.kind = Some(CredentialsKind::Token(token.into()));
         self
     }
 
-    pub fn with_address(&mut self, address: Option<SocketAddr>) -> &mut Self {
+    pub fn with_address(mut self, address: Option<Vec<SocketAddr>>) -> Self {
```
ikopylov, line 224: `with_addresses`
Also, you can add a separate function `with_address` which receives a single `SocketAddr` and wraps it with vector in its body

bob-access/src/credentials.rs
```diff
@@ -9,18 +9,71 @@ use tonic::Request;
...
+            ResolveState::Pending => false,
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct DeclaredCredentials {
+    address: Option<Vec<SocketAddr>>,
+    hostname: Option<String>,
+    kind: Option<CredentialsKind>,
```
ikopylov, line 67: Why the `kind` is Optional? In what case it can not be set for `DeclaredCredentials`?

bob-access/src/credentials.rs
```diff
@@ -9,18 +9,71 @@ use tonic::Request;
...
+                }                
+            },
+            ResolveState::Pending => false,
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct DeclaredCredentials {
+    address: Option<Vec<SocketAddr>>,
```
ikopylov, line 65: Maybe it is better to drop `Option<>` here. No addresses case can be represented with empty vector

bob-access/src/credentials.rs
```diff
@@ -9,18 +9,71 @@ use tonic::Request;
...
+    hostname: Option<String>,
+    kind: Option<CredentialsKind>,
+    resolve_state: ResolveState,
```
ikopylov, line 68: `DeclaredCredentials` is not a good place for `ResolveState`. It should contain only fields, required for authentication. It is better to create a wrapper struct, that will contain both  `DeclaredCredentials` and  `ResolveState`

bob-access/src/credentials.rs
```diff
@@ -59,13 +112,36 @@ impl CredentialsKind {
...
+impl DeclaredCredentials {
+    pub fn builder() -> DeclaredCredentialsBuilder {
+        DeclaredCredentialsBuilder::default()
+    }
+
+    pub fn ip(&self) -> &Option<Vec<SocketAddr>> {
+        &self.address
+    }
+
+    pub fn add_addresses(&mut self, mut addresses: Vec<SocketAddr>) {
```
ikopylov, line 124: 1. It is a bit strange to pass `addresses` vector with `mut` modifier.
2. Why it is `add_addresses`? If resolved addresses changed, you should replace old ones with new (whole collection). Otherwise it is possible to get duplicates.

bob-access/src/credentials.rs
```diff
@@ -9,18 +9,71 @@ use tonic::Request;
...
+        ResolveState::Waiting(0)
+    }
+}
+
+impl ResolveState {
+    fn needs_resolve(&mut self, unresolved: bool) -> bool {
+        match self {
+            ResolveState::Waiting(n) => {
+                if unresolved {
+                    *n += 1;
```
ikopylov, line 51: In this form it has a risk of integer overflow with the following panic

bob-access/src/credentials.rs
```diff
@@ -9,18 +9,71 @@ use tonic::Request;
...
+}
+
+impl Default for ResolveState {
+    fn default() -> Self {
+        ResolveState::Waiting(0)
+    }
+}
+
+impl ResolveState {
+    fn needs_resolve(&mut self, unresolved: bool) -> bool {
```
ikopylov, line 47: This name suggests that the function will not modify anything, but it does. You should rename it. Possible name: `process_resolving_result`

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,136 @@
...
+
+    fn check_node_request(&self, node_name: &String, ip: Option<SocketAddr>) -> bool {
+        let mut unresolved = false;
+        {
+            if ip.is_none() {
+                return false;
+            }
+            let ip = ip.unwrap().ip();
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            if nodes.is_empty() {
```
ikopylov, line 112: This check is not needed. Below you call `nodes.get(node_name)` that will return `None` for the empty list

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,136 @@
...
+                        return true;
                     }
+                    
+                    unresolved = true;
                 }
-                false
-            })
+            }
+        }
+        self.check_resolve(node_name, unresolved);
```
ikopylov, line 129: If there is no node with the specified name in `nodes`, you do not need to call check_resolve`

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,136 @@
...
+    fn spawn_resolver(&self, cred: DeclaredCredentials) {
+        tokio::spawn(Self::resolve_worker(self.nodes.clone(), cred, self.resolve_sleep_dur_ms));
+    }
+
+    async fn resolve_worker(nodes: Arc<RwLock<NodesCredentials>>, cred: DeclaredCredentials, sleep_dur_ms: u64) {
+        let hostname = cred.hostname().as_ref().expect("resolve worker without hostname");
+        let mut addr: Option<Vec<SocketAddr>> = None;
+        let mut cur_sleep_dur_ms = 100;
+        while addr.is_none() || (addr.is_some() && addr.as_ref().unwrap().len() == 0) {
+            tokio::time::sleep(Duration::from_millis(cur_sleep_dur_ms)).await;
```
ikopylov, line 72: With this approach the first resolve will happen only after 100ms. I think, you should run `lookup_host` first time before the loop

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,136 @@
...
-            .map(|creds| {
-                for cred in creds {
-                    if let Some(CredentialsKind::InterNode(other_name)) = cred.kind() {
-                        if node_name == other_name {
-                            return true;
-                        }
+
+        let addr = addr.expect("somehow addr is none");
+        let mut nodes = nodes.write().expect("nodes credentials lock");
+        if let Some(CredentialsKind::InterNode(nodename)) = cred.kind() {
```
ikopylov, line 84: `        let mut nodes = nodes.write().expect("nodes credentials lock");` should be placed inside this `if`

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,136 @@
...
-                        cred.kind().map(|k| k.is_internode()) == Some(true)))
-        {
-            self.nodes = nodes;
+        if Self::node_creds_ok(&nodes) {
+            let mut resolved = HashMap::new();
+            for (nodename, cred) in nodes.drain() {
+                if cred.ip().is_none() && cred.hostname().is_some() {
+                    self.spawn_resolver(cred);
+                } else {
+                    resolved.insert(nodename, cred);
```
ikopylov, line 51: Withe the current `resolve_worker` implementation, all nodes should be in `nodes_creds`

bob-access/src/credentials.rs
```diff
@@ -59,64 +73,186 @@ impl CredentialsKind {
...
+            credentials,
+            resolve_state: ResolveState::Waiting(0),
         }
     }
+
+    pub fn update_resolve_state(&mut self, unresolved: bool) -> bool {
+        self.resolve_state.update(unresolved)
+    }
+
+    pub fn set_resolved(&mut self) {
```
ikopylov, line 241: You can pass a new addresses here as the argument and update them inside function. In this way you don't need `creds_mut` function

bob-access/src/credentials.rs
```diff
@@ -59,64 +73,186 @@ impl CredentialsKind {
...
+        }
+    }
+}
+
+const RESOLVE_THRESHOLD: u32 = 3;
+
+#[derive(Debug, Clone)]
+enum ResolveState {
+    Waiting(u32),
+    Pending,
```
ikopylov, line 123: Names are little bit confusing. I think `Waiting` is better to rename to `Resolved` and `Pending` to `InProgress`

bob-access/src/credentials.rs
```diff
@@ -59,64 +73,186 @@ impl CredentialsKind {
...
+    fn new(kind: CredentialsKind) -> Self {
+        Self {
+            kind,
+            address: Vec::new(),
+            hostname: None,
+        }
+    }
+
+    pub fn with_address(mut self, address: SocketAddr) -> Self {
+        self.address.push(address);
```
ikopylov, line 205: This is not what I meant. I meant, that `with_address` would init a vector with single value, not update existing one by pushing a new item

bob-access/src/credentials.rs
ikopylov: You still should use `wrapping_add` to avoid panics in rare corner case (when even u32 overflows)

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
+                        return true;
                     }
+                    
+                    unresolved = Some(true);
                 }
-                false
-            })
+            }
+        }
+        unresolved.map(|unresolved|
```
ikopylov, line 124: IMHO, `if let Some(unresolved) = unresolved {` is better from readability perspective

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
+            }
+        } else {
+            error!("resolved credentials are not internode");
+        }
+    }
+
+    fn process_auth_result(&self, nodename: &str, unresolved: bool) {
+        let mut nodes = self.nodes.write().expect("nodes credentials lock");
+        if let Some(node) = nodes.get_mut(nodename) {
+            if node.update_resolve_state(unresolved) {
```
ikopylov, line 97: With intensive access from a malicious node, just counting the consecutive number of failed authorizations is not enough to prevent DDoS. We should additionally check that the time interval passed from the previous successful resolution is larger than `resolve_sleep_dur_ms`. Otherwise, the malicious node will trigger re-resolving indefinitely without any pause

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
+    async fn resolve_worker(nodes: Arc<RwLock<NodesCredentials>>, cred: DeclaredCredentials, sleep_dur_ms: u64) {
+        let hostname = cred.hostname().as_ref().expect("resolve worker without hostname");
+        let mut addr: Option<Vec<SocketAddr>> = lookup(hostname).await;
+        let mut cur_sleep_dur_ms = 100;
+        while addr.is_none() || (addr.is_some() && addr.as_ref().unwrap().len() == 0) {
+            tokio::time::sleep(Duration::from_millis(cur_sleep_dur_ms)).await;
+
+            addr = lookup(hostname).await;
+
+            cur_sleep_dur_ms = sleep_dur_ms.min(cur_sleep_dur_ms * 2);
```
idruzhitskiy, line 79: I think that there should be max sleep time. And also all plain numbers should be in constants

Boneyan: https://github.com/qoollo/bob/pull/622#discussion_r1032687303
min is intended in this place

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+    pub fn build(self) -> DeclaredCredentials {
+        DeclaredCredentials {
+            address: self.address,
+            kind: self.kind,
+            hostname: self.hostname,
+        }
+    }
+}
+
+const RESOLVE_SLEEP_DUR_MS: u64 = 1000;
```
ikopylov, line 224: This should not be a constant. I suggest to use `hostname_resolve_dur_ms` also for protection period.

bob-common/src/configs/node.rs
```diff
@@ -644,6 +647,14 @@ impl NodeConfig {
         AuthenticationType::None
     }
 
+    fn default_hostname_resolve_dur_ms() -> u64 {
+        5000
+    }
+
+    pub fn hostname_resolve_dur_ms(&self) -> u64 {
```
ikopylov, line 654: `dur` (duration) is not a good name, as it is actually a period. `hostname_resolve_period_ms` is better

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+impl DCredentialsResolveGuard {
+    pub fn new(credentials: DeclaredCredentials) -> Self {
+        Self {
+            credentials,
+            resolve_state: ResolveState::Resolved(0),
+            resolve_threshold: Instant::now() + Duration::from_millis(RESOLVE_SLEEP_DUR_MS),
+        }
+    }
+
+    pub fn update_resolve_state(&mut self, unresolved: bool) -> bool {
```
ikopylov, line 242: In this function, you should also check if the credential was created with an IP address or with a hostname

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+        let now = Instant::now();
+        if !unresolved {
+            if now >= self.resolve_threshold {
```
ikopylov, line 245: I don't understand, why do you check time when the authentication was successful. You don't need that, you can just call `self.resolve_state.update(unresolved)`

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+        Self {
+            credentials,
+            resolve_state: ResolveState::Resolved(0),
+            resolve_threshold: Instant::now() + Duration::from_millis(RESOLVE_SLEEP_DUR_MS),
+        }
+    }
+
+    pub fn update_resolve_state(&mut self, unresolved: bool) -> bool {
+        let now = Instant::now();
+        if !unresolved {
```
ikopylov, line 244: I think, `unresolved` is a bad name for this parameter. It is actually indicates the authentication status, so can it be named `unauthenticated` or, better, be reversed and named `authenticated`

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+            if now >= self.resolve_threshold {
+                self.resolve_state.update(unresolved)
+            } else {
+                false
+            }
+        } else {
+            self.resolve_threshold = now + Duration::from_millis(RESOLVE_SLEEP_DUR_MS);
```
ikopylov, line 251: Only here you should check elapsed time

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,199 @@ impl CredentialsKind {
...
+            self.resolve_state.update(unresolved)
+        }
+    }
+
+    pub fn set_resolved(&mut self, addresses: Vec<SocketAddr>) {
+        self.credentials.replace_addresses(addresses);
+        self.resolve_state = ResolveState::Resolved(0);
```
ikopylov, line 258: `resolve_threshold` should be updated here

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
+        if let Some(node) = nodes.get_mut(nodename) {
+            if node.update_resolve_state(unresolved) {
+                node.set_in_progress();
+                self.spawn_resolver(node.creds().clone());
+            }
+        }
+    }
+
+    fn check_node_request(&self, node_name: &String, ip: Option<SocketAddr>) -> bool {
+        let mut unresolved = None;
```
ikopylov, line 104: `unresolved` is not a good name, because this variable actually indicates the authentication status

bob-access/src/credentials.rs
Boneyan: I accidentally confused if branches, so this check is in unauthenticated case

bob-access/src/credentials.rs
```diff
@@ -59,64 +74,201 @@ impl CredentialsKind {
...
+        Self {
+            credentials,
+            resolve_state: ResolveState::Resolved(0),
+            resolve_threshold: Instant::now() + Duration::from_millis(resolve_protection_period_ms),
+            resolve_protection_period_ms,
         }
     }
+
+    pub fn update_resolve_state(&mut self, authenticated: bool) -> bool {
+        if self.credentials.hostname().is_none() {
```
ikopylov, line 244: I suggest to declare a function in `DeclaredCredentials` that returns true if the address resolved from hostname (`fn is_address_resolved_from_hostname(..) -> bool`)

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
+                    } else {
+                        authenticated = Some(false);
                     }
                 }
-                false
-            })
+            }
+        }
+        if let Some(authenticated) = authenticated {
+            self.process_auth_result(node_name, authenticated);
```
ikopylov, line 123: `process_auth_result` acquires write lock inside, so it is better to avoid calling it in most common scenarios. I suggest to add a function to `DCredentialsResolveGuard` that checks whether the status update needed (it is not needed if  `resolve_state == Resolved(0) && authenticated` or `resolve_state == InProgress`). Basically the logic is the same as in `update_resolve_state`, but without the actual update. Then you can call it inside read lock above and if it returned `false` then do not call `process_auth_result`.

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,134 @@
...
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            nodes.get(node_name).map(|guard| {
+                let cred = guard.creds();
+                if let CredentialsKind::InterNode(other_name) = cred.kind() {
+                    debug_assert!(node_name == other_name);
+                    if cred.ip().iter().find(|cred_ip| cred_ip.ip() == ip).is_some() {
+                        authenticated = true;
+                        needs_update = guard.needs_update(true);
+                    } else {
+                        needs_update = guard.needs_update(false);
```
ikopylov, line 120: Estimate `needs_update` value after `if` statement. The code will be shorter:
```
if cred.ip().iter().find(|cred_ip| cred_ip.ip() == ip).is_some() {
    authenticated = true;
}
needs_update = guard.needs_update(authenticated);
```

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,134 @@
...
+    fn check_node_request(&self, node_name: &String, ip: Option<SocketAddr>) -> bool {
+        let mut authenticated = false;
+        let mut needs_update = false;
+        {
+            if ip.is_none() {
+                return false;
+            }
+            let ip = ip.unwrap().ip();
+            let nodes = self.nodes.read().expect("nodes credentials lock");
+            nodes.get(node_name).map(|guard| {
```
ikopylov, line 112: Looks like `if` statement here will be clearer:
```
if Some(guard) = nodes.get(node_name) {
...
}
```

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,223 @@ impl CredentialsKind {
...
+        }
+        if authenticated {
+            match self.resolve_state {
+                ResolveState::Resolved(n) => {
+                    n > 0
+                },
+                ResolveState::InProgress => false,
+            }
+        } else {
+            Instant::now() >= self.resolve_threshold
```
ikopylov, line 275: First you should check for `ResolveState::InProgress`. When resolving is in progress we don't need to update the state.
I think, you can check `if self.resolve_state == ResolveState::InProgress` right after `self.credentials.is_created_with_address()` check

bob-access/src/credentials.rs
```diff
@@ -59,66 +74,223 @@ impl CredentialsKind {
...
+        } else {
+            if Instant::now() >= self.resolve_threshold {
+                self.resolve_state.update(authenticated)
+            } else {
+                false
+            }
+        }
+    }
+
+    pub fn needs_update(&self, authenticated: bool) -> bool {
```
ikopylov, line 263: Add comment that the logic of this function is tightly coupled with `update_resolve_state`

bob-access/src/authenticator/basic.rs
```diff
@@ -1,65 +1,132 @@
...
-                        cred.kind().map(|k| k.is_internode()) == Some(true)))
-        {
-            self.nodes = nodes;
+        if Self::node_creds_ok(&nodes) {
+            let mut nodes_creds = self.nodes.write().expect("nodes credentials lock");
+            for (nodename, cred) in nodes.drain() {
+                let guard = DCredentialsResolveGuard::new(cred.clone(), self.resolve_sleep_period_ms);
+                nodes_creds.insert(nodename, guard);
+                if cred.ip().is_empty() && cred.hostname().is_some() {
+                    self.spawn_resolver(cred);
```
ikopylov, line 56: You should set resolving status to `InProgress` before spawning the resolver thread

ikopylov:
Please also run a test to make sure bob stops properly when resolving the hostname task is in progress. To do so, create a cluster configuration with 2 nodes, where the address of the second node is the hostname that is not available (just a random string), and start only the first node. That node will try to resolve the hostname indefinitely in background. After about 30 seconds, stop the bob process with `Ctrl+C` and see how long it takes.

Boneyan:
> Please also run a test to make sure bob stops properly when resolving the hostname task is in progress. To do so, create a cluster configuration with 2 nodes, where the address of the second node is the hostname that is not available (just a random string), and start only the first node. That node will try to resolve the hostname indefinitely in background. After about 30 seconds, stop the bob process with `Ctrl+C` and see how long it takes.

Bob stops pretty immediately

ikopylov (approved):
LGTM

PR #627, opened 2022-10-02
616 proper basic auth

### LINKED ISSUE #616: Authorization: implement proper basic auth
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

Boneyan:
Using standard Authorization header for basic auth 
Closes #616


bob-access/src/extractor.rs
```diff
@@ -40,14 +40,25 @@ impl<T: Extractor> ExtractorExt for T {
...
-                Ok(creds)
-            },
-            (None, None) => {
+        let auth_header = self.get_header("authorization")?;
+        let mut err = "missing authorization header";
+        if let Some(auth_header) = auth_header {
+            let mut parts = auth_header.split_whitespace();
+            if let (Some("Basic"), Some(credentials)) = (parts.next(), parts.next()) {
+                let credentials = base64::decode(credentials)
+                    .map_err(|_| Error::ConversionError("bad base64 credentials".into()))?;
```
idruzhitskiy, line 49: Original error should be included in result error, at least part of it

bob-access/src/extractor.rs
```diff
@@ -40,14 +40,25 @@ impl<T: Extractor> ExtractorExt for T {
...
-                    .with_username_password(username, password)
-                    .build();
-                Ok(creds)
-            },
-            (None, None) => {
+        let auth_header = self.get_header("authorization")?;
+        let mut err = "missing authorization header";
+        if let Some(auth_header) = auth_header {
+            let mut parts = auth_header.split_whitespace();
+            if let (Some("Basic"), Some(credentials)) = (parts.next(), parts.next()) {
```
idruzhitskiy, line 47: I think this should be extracted into separate function, "find_username_password" or something similar

bob-apps/bin/bobt.rs
```diff
@@ -374,13 +374,13 @@ impl Settings {
...
-            if self.password.is_some() {
-                b = b.header("password", self.password.as_ref().unwrap());
-            }
+    fn append_request_headers(&self, mut b: RequestBuilder) -> RequestBuilder {
+        if let (Some(username), Some(password)) = (self.username.as_ref(), self.password.as_ref()) {
+            let credentials = format!("{}:{}", username, password);
+            let credentials = base64::encode(credentials);
+            let authorization = format!("Basic {}", credentials);
+
+            b = b.header("authorization", authorization);
```
ikopylov, line 383: Here is a function for basic auth: https://docs.rs/reqwest/0.11.13/reqwest/struct.RequestBuilder.html#method.basic_auth

bob-access/src/extractor.rs
```diff
@@ -40,14 +53,16 @@ impl<T: Extractor> ExtractorExt for T {
...
+                let (username, password) = username_password_from_credentials(credentials)?;
                 let creds = builder
                     .with_username_password(username, password)
                     .build();
-                Ok(creds)
-            },
-            (None, None) => {
+                return Ok(creds);
+            } else {
```
ikopylov, line 65: Wrong nesting level: inter-node auth should be checked when no `authorization` header is provided

bob-access/src/extractor.rs
```diff
@@ -23,6 +23,19 @@ fn prepare_builder<T: Extractor>(slf: &T) -> Result<CredentialsBuilder, Error> {
     Ok(builder)
 }
 
+fn username_password_from_credentials(credentials: &str) -> Result<(String, String), Error> {
+    let credentials = base64::decode(credentials)
+        .map_err(|e| Error::ConversionError(format!("bad base64 credentials: {}", e)))?;
+    let str_cred = String::from_utf8(credentials)
+        .map_err(|e| Error::ConversionError(format!("invalid utf8 credentials characters: {}", e)))?;
+    let mut user_pass = str_cred.split_terminator(":");
+    if let (Some(username), Some(password)) = (user_pass.next(), user_pass.next()) {
```
ikopylov, line 32: `user_pass` can be empty. You should support this case here

bob-access/src/extractor.rs
```diff
@@ -40,14 +53,16 @@ impl<T: Extractor> ExtractorExt for T {
...
+                let (username, password) = username_password_from_credentials(credentials)?;
                 let creds = builder
                     .with_username_password(username, password)
                     .build();
-                Ok(creds)
-            },
-            (None, None) => {
+                return Ok(creds);
+            } else {
                 if let Some(node_name) = self.get_header("node_name")? {
```
ikopylov, line 66: I think it would be good to perform inter-node auth also through `authorization` header. We can use similar format: `Authorization: InterNode <base64 node_name>`

bob-access/src/extractor.rs
```diff
@@ -40,14 +53,16 @@ impl<T: Extractor> ExtractorExt for T {
 
     fn extract_basic(&self) -> Result<Credentials, Error> {
         let mut builder = prepare_builder(self)?;
-        match (self.get_header("username")?, self.get_header("password")?) {
-            (Some(username), Some(password)) => {
+        let auth_header = self.get_header("authorization")?;
+        if let Some(auth_header) = auth_header {
+            let mut parts = auth_header.split_whitespace();
+            if let (Some("Basic"), Some(credentials)) = (parts.next(), parts.next()) {
```
ikopylov, line 59: Unfortunately, authorization scheme should be case-insensitive according to RFC (https://www.rfc-editor.org/rfc/rfc2617). So you should compare with `Basic` in case insensitive way. Please, don't use `to_lowercase` conversion in comparison, because this logic is on the hot path. It is better to find (or write) a function, that do case-insensitive comparison in letter-by-letter way without string allocation

bob-common/src/bob_client.rs
```diff
@@ -204,9 +207,10 @@ pub mod b_client {
         }
 
         fn set_credentials<T>(&self, req: &mut Request<T>) {
-            let val = MetadataValue::from_str(&self.local_node_name)
+            let val = MetadataValue::from_str(&self.auth_header)
                 .expect("failed to create metadata value from node name");
-            req.metadata_mut().insert("node_name", val);
+            req.metadata_mut().insert("authorization", val);
```
ikopylov, line 212: It is better to start with capital letter: `Authorization`

bob-apps/bin/bobp.rs
```diff
@@ -194,19 +194,17 @@ impl TaskConfig {
...
-            .expect("can not parse password into header");
+            .expect("can not parse authorization value");
 
         move |a| {
             let mut request = Request::new(a);
             if do_basic_auth {
                 let req_md = request.metadata_mut();
-                req_md.insert("username", basic_username.clone());
-                req_md.insert("password", basic_password.clone());
+                req_md.insert("authorization", authorization.clone());
```
ikopylov, line 207: It is better to start with capital letter: `Authorization`

bob-access/src/extractor.rs
```diff
@@ -40,27 +67,32 @@ impl<T: Extractor> ExtractorExt for T {
...
-                    .build();
-                Ok(creds)
-            },
-            (None, None) => {
-                if let Some(node_name) = self.get_header("node_name")? {
+        let auth_header = self.get_header("authorization")?;
+        if let Some(auth_header) = auth_header {
+            let mut parts = auth_header.split_whitespace();
+            match (parts.next(), parts.next()) {
+                (Some("Basic"), Some(credentials)) => {
```
ikopylov, line 74: As I wrote before, the comparison should be case insensitive. You can write simple wrapper struct that support case insensitive comparison with String, so it will be possible to use `match` in the same way

ikopylov: Looks like there is a crate, that do exactly what I wrote: https://docs.rs/unicase/latest/unicase/struct.Ascii.html

ikopylov (approved):
LGTM

PR #629, opened 2022-10-03
570 uniform alien distribution

### LINKED ISSUE #570: Choose supported nodes for aliens evenly
In the current implementation first nodes in the list are chosen for every request. As a result, these nodes contain most of the aliens. 
We should achieve an uniform distribution of aliens. Possible solutions:
1. Introduce number that represent the offset in the list and increment that number for every call to `get_support_nodes`. In this way every next request will give the slice of nodes shifted by 1 from the previous one
2. Calculate some simple hash function from the key and use its result as an offset in the list of nodes. In this way we will achieve uniform distribution and will have stable results for the same key

https://github.com/qoollo/bob/blob/ab2b97c32fed59fe31aad50c1610ab3891777a5c/bob-common/src/mapper.rs#L118

idruzhitskiy:
Closes #570


bob-common/src/mapper.rs
```diff
@@ -116,22 +122,24 @@ impl Virtual {
...
 
     pub fn get_support_nodes(&self, key: BobKey, count: usize) -> Vec<&Node> {
+        debug_assert!(count < self.nodes.len());
         trace!("get target nodes for given key");
         let target_nodes = self.get_target_nodes_for_key(key);
         trace!("extract indexes of target nodes");
-        let mut target_indexes = target_nodes.iter().map(Node::index);
-        let len = target_indexes.size_hint().0;
-        debug!("iterator size lower bound: {}", len);
+        let target_indexes: Vec<_> = target_nodes.iter().map(Node::index).collect();
```
ikopylov, line 129: `collect` will most likely only slow things down. `target_nodes` is a slice so its iteration is super fast. If you want to optimize, then it is better to remove `target_indexes` variable and use `target_nodes` directly below

bob-common/src/mapper.rs
```diff
@@ -116,22 +122,24 @@ impl Virtual {
...
             .iter()
             .filter_map(|(id, node)| {
-                if target_indexes.all(|i| &i != id) {
+                if !target_indexes.contains(id) {
                     Some(node)
                 } else {
                     None
                 }
             })
+            .cycle()
```
ikopylov, line 141: Unfortunately, there are edge cases where a `cycle()` can result in a duplicate nodes in the result. 
Example: 3 nodes, 2 replicas (replicas are for vdisks and therefore they can be easily located on 3 nodes), quorum is 2. If data written and both replicas are not available, then this function will be called with `count = 2` and 2 nodes out of 3 will be filtered out (because they are target for specific key). This results in a single node to be repeated 2 times in the result

ikopylov: This can be avoided with the following approach:
1. Create vector with capacity = `count`
2. Iterate and fill vector: `for node in self.nodes.iter().skip(offset % self.nodes.len()) { .. }`
3. If vector length is less than `count` and `offset % self.nodes.len() > 0` then additionally iterate from the beginning: `for node in self.nodes.iter().take(offset % self.nodes.len()) { ... }`

lostl1ght:
Fixed.

bob-common/src/mapper.rs
```diff
@@ -116,24 +122,26 @@ impl Virtual {
...
-            .filter_map(|(id, node)| {
-                if target_indexes.all(|i| &i != id) {
-                    Some(node)
-                } else {
-                    None
+        let offset = self.support_nodes_offset.fetch_add(1, Ordering::Relaxed);
+        let mut support_nodes = Vec::with_capacity(count);
+        for (id, node) in self.nodes.iter().skip(offset % self.nodes.len()) {
+            if target_nodes.iter().all(|i| i.index() != *id) {
+                support_nodes.push(node)
```
ikopylov, line 134: You should `break` if `support_nodes.len() >= count`

bob-common/src/mapper.rs
```diff
@@ -116,24 +122,26 @@ impl Virtual {
...
+            }
+        }
+        if support_nodes.len() < count && offset % self.nodes.len() > 0 {
+            for (id, node) in self.nodes.iter().take(offset % self.nodes.len()) {
+                if target_nodes.iter().all(|i| i.index() != *id) {
+                    support_nodes.push(node)
```
ikopylov, line 140: Same here: You should `break` if `support_nodes.len() >= count`

bob-common/src/mapper.rs
```diff
@@ -116,24 +122,26 @@ impl Virtual {
     }
 
     pub fn get_support_nodes(&self, key: BobKey, count: usize) -> Vec<&Node> {
```
ikopylov, line 124: Please, add shortcut for situation when `count == 0`:
```
if count == 0 {
    return vec![];
}

File: bob-common/src/mapper.rs [hunk 6]
```diff
@@ -116,24 +122,26 @@ impl Virtual {
...
+        debug_assert!(count < self.nodes.len());
```
ikopylov, line 125: In theory it is possible, that `count == self.nodes.len()`
```suggestion
        debug_assert!(count <= self.nodes.len());
```

bob-common/src/mapper.rs
lostl1ght: This `if` will not be entered due to `support_nodes.len() < count` condition

bob-common/src/mapper.rs
lostl1ght: Done.

bob-common/src/mapper.rs
lostl1ght: Done.

bob-common/src/mapper.rs
lostl1ght: Done.

bob-common/src/mapper.rs
```diff
@@ -116,24 +122,35 @@ impl Virtual {
...
+            for (id, node) in self.nodes.iter().take(offset % self.nodes.len()) {
+                if target_nodes.iter().all(|i| i.index() != *id) {
+                    support_nodes.push(node);
+                    if support_nodes.len() >= count {
+                        break;
+                    }
+                }
+            }
+        }
+        support_nodes
```
ikopylov, line 153:
```suggestion
        debug_assert!(support_nodes.len() <= count);
        support_nodes
```

ikopylov (approved):
LGTM

PR #631, opened 2022-10-09
592 limit metric log

### LINKED ISSUE #592: Limit log about metrics queue overflow
When graphite is not available the following log are printed for every attempt to update counters:
```
2022-08-29 10:20:16  bob_common::metrics::collector: 17 ERROR Can't send metric to thread, which processing metrics: no available capacity
```

We should accumulate this logs and write it once in 5s

Boneyan:
Using interval logger in metric exporter to reduce error log density
Closes #592


idruzhitskiy (approved):

bob-common/Cargo.toml
```diff
@@ -23,6 +24,7 @@ metrics = { version = "0.17", features = ["std"] }
 metrics-exporter-prometheus = { version = "0.6", features = ["tokio-exporter"]}
 metrics-util = "0.10"
 mockall = "0.10"
+parking_lot = "0.12.0"
```
ikopylov, line 27: In recent versions of Rust, the standard Mutex has become much faster. It is even slightly faster than in the parking_lot one. So we decided to use standard implementation everywhere and you should drop this dependency

bob-common/src/interval_logger.rs
```diff
@@ -48,17 +48,19 @@ impl<E: Hash + Eq + Display> IntervalLogger<E> {
...
     pub fn new(interval_ms: u64, level: Level) -> Self {
         Self {
-            inner: PLMutex::new(IntervalLogger::new(interval_ms, level))
+            inner: Mutex::new(IntervalLogger::new(interval_ms, level))
         }
     }
 
     pub fn report_error(&self, action: E) {
-        self.inner.lock().report_error(action);
+        if let Ok(mut lock) = self.inner.lock() {
```
ikopylov, line 62: It is better to assume, that mutex will be acquired successfully every time. Thus I suggest to use `expect()` instead of condition

bob-common/src/metrics/collector/recorder.rs
```diff
@@ -1,23 +1,44 @@
...
+        }
     }
 
     fn push_metric(&self, m: Metric) {
         if let Err(e) = self.tx.try_send(m) {
-            error!(
-                "Can't send metric to thread, which processing metrics: {}",
-                e
-            );
+            let action = PushMetricError {err: e.to_string()};
```
ikopylov, line 40: I don't like this `to_string()` conversion that will happen on every metrics, when Graphite is not available. According to the documentation there are 2 types of error possible: when closed and when no capacity (https://docs.rs/tokio/0.1.22/tokio/sync/mpsc/error/struct.TrySendError.html). I suggest to convert `PushMetricError` into enum with 3 states: `Unknown`, `Closed`, `NoCapacity`. Here you can convert to that enum with functions `is_closed()`, `is_full()`

ikopylov (approved):
LGTM

PR #633, opened 2022-10-18
573 parallel local put

### LINKED ISSUE #573: Run local put in parallel with remote puts
In the current implementation, if the local node contains target vdisk, then it first executes local put request (https://github.com/qoollo/bob/blob/master/bob/src/cluster/quorum.rs#L38). Only after its completion it executes put requests to remote nodes (https://github.com/qoollo/bob/blob/master/bob/src/cluster/quorum.rs#L53). This approach is not optimal. It is better to start local put and remote put simultaniously and only after that awaits for result.
Possible algorithm:
1. Start local put, but do not await for result
2. Start remote puts, but do not await for results
3. Await for local put result
4. Await for remote put results with `at_least` value incorporating local put success

idruzhitskiy:
Closes #573

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Bob versions changelog
 
 
 #### Changed
-
+- Make local put parallel to remote (#573)
 
 #### Fixed
 
@@ -23,7 +23,6 @@ Bob versions changelog
 #### Changed
 - Removed `open_blobs_soft_limit` and `open_blobs_hard_limit` from node config example (#703)
 - Add condition to background_put (#589)
-- Make local put parallel to remote (#573)
 
 #### Fixed
 - Fix incorrect timestamp used in `Group::delete()` (#741)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Bob versions changelog
 - Remove unnecessary data clone (#506)
 - Compare vdiskid first (#594)
 - Optimize finding actual holders (#595)
+- Make local put parallel to remote (#573)
 
 #### Fixed
 - Print full error text received from Pearl in exist function (#581)
--- a/bob/src/cluster/quorum.rs
+++ b/bob/src/cluster/quorum.rs
@@ -2,9 +2,9 @@ use crate::prelude::*;
 
 use super::{
     operations::{
-        delete_at_nodes, delete_at_local_node, group_keys_by_nodes, lookup_local_alien, lookup_local_node,
-        lookup_remote_aliens, lookup_remote_nodes, put_at_least, put_local_all, put_local_node,
-        put_sup_nodes, Tasks,
+        delete_at_local_node, delete_at_nodes, finish_at_least_handles, group_keys_by_nodes,
+        lookup_local_alien, lookup_local_node, lookup_remote_aliens, lookup_remote_nodes,
+        put_at_least, put_local_all, put_local_node, put_sup_nodes, Tasks,
     },
     Cluster,
 };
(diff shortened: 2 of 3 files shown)


bob/src/cluster/quorum.rs
```diff
@@ -30,27 +30,34 @@ impl Quorum {
...
+        let local_put_task = disk_path.map(|path| {
             debug!("disk path is present, try put local");
-            let res = put_local_node(&self.backend, key, data, vdisk_id, path).await;
-            if let Err(e) = res {
+            put_local_node(&self.backend, key, data, vdisk_id, path).boxed()
+        });
+        debug!(
+            "PUT[{}] need at least {} additional puts",
+            key,
+            at_least - 1
```
ikopylov, line 43: This value depends on whether local put started or not. 
I suggest to use `if let Some(path) = disk_path` as it was before and update `at_least` value inside

bob/src/cluster/quorum.rs
```diff
@@ -30,27 +30,34 @@ impl Quorum {
...
+        );
+        debug!("PUT[{}] ~~~PUT TO REMOTE NODES~~~", key);
+        let (mut tasks, mut errors) = self.put_remote_nodes(key, data, at_least - 1).await;
```
ikopylov, line 46: Again: `at_least - 1` is incorrect if local put was not started

bob/src/cluster/quorum.rs
```diff
@@ -30,27 +30,34 @@ impl Quorum {
...
+        if let Some(task) = local_put_task {
+            let local_put = task.await;
```
ikopylov, line 48: There is a possibility, that local put task would not start until this `await`. Please, check how it works in Tokio. 
Generally, it is better to use equivalent of `Task.WhenAll`

bob/src/cluster/quorum.rs
```diff
@@ -30,27 +30,34 @@ impl Quorum {
...
+                debug!("PUT[{}] local failed, put another remote", key);
             } else {
                 local_put_ok += 1;
-                at_least -= 1;
                 debug!("PUT[{}] local node put successful", key);
             }
-        } else {
-            debug!("skip local put");
+        };
+        if local_put_ok == 0 {
```
ikopylov, line 58: This should be run only if local put finished with error. It should not if local put wasn't started at all (because there is no need to put to the local node)

bob/src/cluster/quorum.rs
ikopylov: I ran tests. I am right. To run both tasks you should use `tokio::join!` macro

bob/src/cluster/quorum.rs
```diff
@@ -30,32 +30,34 @@ impl Quorum {
...
-        debug!("PUT[{}] ~~~PUT TO REMOTE NODES~~~", key);
-        let (tasks, errors) = self.put_remote_nodes(key, data, at_least).await;
+                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES~~~", key, at_least);
+                self.put_remote_nodes(key, data, at_least).await
+            };
         let all_count = self.mapper.get_target_nodes_for_key(key).len();
-        remote_ok_count += all_count - errors.len() - tasks.len() - local_put_ok;
+        let remote_ok_count = all_count - errors.len() - tasks.len() - local_put_ok;
         failed_nodes.extend(errors.iter().map(|e| e.node_name().to_string()));
         if remote_ok_count + local_put_ok >= self.quorum {
```
ikopylov, line 62: Please, add `assert(tasks.is_empty(), "All target nodes put are expected to be completed before alien put begins");` on the `else` branch

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Bob versions changelog
 #### Changed
 - Removed `open_blobs_soft_limit` and `open_blobs_hard_limit` from node config example (#703)
 - Add condition to background_put (#589)
+- Make local put parallel to remote (#573)
```
ikopylov, line 26: Move this line to `[Unreleased]` section

ikopylov (approved):
LGTM

PR #634, opened 2022-10-18
576, 585 lookup existence in aliens, separate remote & local lookup

### LINKED ISSUE #576: `exists` operation does not check aliens
https://github.com/qoollo/bob/blob/master/bob/src/cluster/quorum.rs#L212

### LINKED ISSUE #585: Low `exists` performance on single key requests
Roughly ~1.5 times slower than `get`

lostl1ght:
Closes #576.
Closes #585.


bob/src/cluster/quorum.rs
```diff
@@ -272,21 +272,130 @@ impl Cluster for Quorum {
...
-                for (&r, &ind) in result.inner().iter().zip(&indexes) {
-                    exist[ind] |= r;
+
+        // filter local keys
+        let mut indices = Vec::new();
+        let local_keys = keys
+            .iter()
+            .enumerate()
+            .filter_map(|(idx, &key)| {
+                if self
```
idruzhitskiy, line 284: Maybe it's better to use `find(...).map(|idx| {indices.push(idx); idx})` to remove `if` and explicit `Some` / `None`

bob/src/cluster/quorum.rs
```diff
@@ -272,21 +272,130 @@ impl Cluster for Quorum {
...
-        );
         let len = keys.len();
         let mut exist = vec![false; len];
-        for (nodes, (keys, indexes)) in keys_by_nodes {
-            let cluster_results = LinkManager::exist_on_nodes(&nodes, &keys).await;
-            for result in cluster_results.into_iter().flatten() {
-                for (&r, &ind) in result.inner().iter().zip(&indexes) {
-                    exist[ind] |= r;
+
+        // filter local keys
```
idruzhitskiy, line 278: This comments looks like function names :) Maybe it's better to move this code to separate functions

bob/src/cluster/quorum.rs
```diff
@@ -272,21 +272,130 @@ impl Cluster for Quorum {
     }
 
     async fn exist(&self, keys: &[BobKey]) -> Result<Vec<bool>, Error> {
```
idruzhitskiy, line 274: Function is too large and should be decomposed

idruzhitskiy (changes requested):

bob/src/cluster/quorum.rs
lostl1ght: Yes, the question is: in which file should I put them?

idruzhitskiy: Why not in the same file, since they are related to this functionality

bob/src/cluster/quorum.rs
lostl1ght: Done: [2bd9373](https://github.com/qoollo/bob/pull/634/commits/2bd9373699069170d8982387dc7806df38e3c957)

bob/src/cluster/quorum.rs
lostl1ght: I was able to eliminate `is_some` call using `find_map` in [2695039](https://github.com/qoollo/bob/pull/634/commits/2695039d6d39b4b0353ae0b3f62bfe70f8c33091).

Using `find(...).map(...)` on `keys` will not work as `find` will return the first key and it is needed to iterate over every key.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Bob versions changelog
 
 ## [Unreleased]
 #### Added
+- Lookup existence in aliens, separate local and remote lookup (#576, #585)
```
ikopylov, line 7: Please, split this line into 2: first for issue #576 and second for #585

bob-apps/Cargo.toml
```diff
@@ -51,7 +51,8 @@ default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-version = "0.14.0"
+# version = "0.14.0"
+git = "https://github.com/qoollo/pearl"
```
ikopylov, line 55: New pearl is already available, so changes in this file should be reverted

bob-backend/Cargo.toml
```diff
@@ -37,7 +37,8 @@ tower = "0.4"
 tower-service = "0.3"
 
 [dependencies.pearl]
-version = "0.14.0"
+# version = "0.14.0"
+git = "https://github.com/qoollo/pearl"
```
ikopylov, line 41: Same here

bob/Cargo.toml
```diff
@@ -45,7 +45,8 @@ tower-service = "0.3"
 uuid = "0.8.2"
 
 [dependencies.pearl]
-version = "0.14.0"
+# version = "0.14.0"
+git = "https://github.com/qoollo/pearl"
```
ikopylov, line 49: Same here

lostl1ght: Done.

bob-backend/Cargo.toml
lostl1ght: Done.

bob-apps/Cargo.toml
lostl1ght: Done.

CHANGELOG.md
lostl1ght: Done.

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
-            let cluster_results = LinkManager::exist_on_nodes(&nodes, &keys).await;
-            for result in cluster_results.into_iter().flatten() {
-                for (&r, &ind) in result.inner().iter().zip(&indexes) {
-                    exist[ind] |= r;
+
+        // filter local keys
+        let (indices, local_keys) = filter_local_keys(keys, &self.mapper);
+        // end
+        debug!("local keys {:?}", local_keys);
+        debug!("local indices {:?}", indices);
```
ikopylov, line 282: 2 logs should be merged into one. Also they should be `trace!`, because they print the raw request data

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        if local_keys.len() > 0 {
+            let result = exist_on_local_node(&self.backend, &local_keys).await?;
+            for (idx, r) in indices.into_iter().zip(result.into_iter()) {
+                exist[idx] |= r;
+            }
+            debug!("exist after local node {:?}", exist);
```
ikopylov, line 288: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        }
+
+        // filter keys that were not found
+        let local_alien = filter_not_found(&exist, keys);
+        // end
+        debug!("local alien keys {:?}", local_alien);
```
ikopylov, line 294: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        if local_alien.len() > 0 {
+            let result = exist_on_local_alien(&self.backend, &local_alien).await?;
+            update_exist(&mut exist, &result);
+            debug!("exist after local alien {:?}", exist);
```
ikopylov, line 298: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        }
+
+        // filter remote not found keys by nodes
+        let remote_keys = filter_remote_not_found_by_nodes(&exist, keys, &self.mapper);
+        // end
+
+        debug!("remote keys by nodes {:?}", remote_keys);
```
ikopylov, line 305: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        if remote_keys.len() > 0 {
+            for (nodes, (keys, indices)) in remote_keys {
+                let result = exist_on_remote_nodes(&nodes, &keys).await;
+                for res in result.into_iter().flatten() {
+                    for (&r, &ind) in res.inner().iter().zip(&indices) {
+                        exist[ind] |= r;
+                    }
+                }
+            }
+            debug!("exist after remote nodes {:?}", exist);
```
ikopylov, line 315: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+        }
+
+        // filter remote not found keys
+        let remote_alien = filter_not_found(&exist, keys);
+        // end
+        if remote_alien.len() > 0 {
+            // filter remote nodes
+            let remote_nodes = filter_remote_nodes(&self.mapper);
+            // end
+            debug!("remote nodes {:?}", remote_nodes);
```
ikopylov, line 325: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+
+            let result = exist_on_remote_aliens(&remote_nodes, &remote_alien).await;
+            debug!("alien result {:?}", result);
```
ikopylov, line 328: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+            for res in result.into_iter() {
+                if let Ok(inner) = res {
+                    debug!("inner {:?}", inner);
```
ikopylov, line 331: `trace!`

bob/src/cluster/quorum.rs
```diff
@@ -294,3 +340,73 @@ impl Cluster for Quorum {
...
+    }
+    not_found_keys
+}
+
+fn filter_remote_not_found_by_nodes(exist: &[bool], keys: &[BobKey], mapper: &Virtual) -> HashMap<Vec<Node>, (Vec<BobKey>, Vec<usize>)> {
+    let mut remote_keys: HashMap<_, (Vec<_>, Vec<_>)> = HashMap::new();
+    for (idx, &r) in exist.iter().enumerate() {
+        if r == false {
+            remote_keys
+                .entry(mapper.get_target_nodes_for_key(keys[idx]).to_vec())
```
ikopylov, line 378: Local node should be skipped. Otherwise the request will go to the local node through network like it was before

ikopylov, line 378: `mapper.get_target_nodes_for_key` already called in `filter_local_keys`. It seems that it is possible to do the grouping once

bob/src/cluster/quorum.rs
```diff
@@ -294,3 +340,73 @@ impl Cluster for Quorum {
...
+                Some(node.clone())
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<Node>>();
+    remote_nodes
+}
+
+fn update_exist(exist: &mut [bool], result: &[bool]) {
```
ikopylov, line 404: It seems to me, that it is possible to add a `struct` that will be responsible for keys mapping by indices

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+            }
+            debug!("exist after local node {:?}", exist);
+        }
+
+        // filter keys that were not found
+        let local_alien = filter_not_found(&exist, keys);
+        // end
+        debug!("local alien keys {:?}", local_alien);
+        if local_alien.len() > 0 {
+            let result = exist_on_local_alien(&self.backend, &local_alien).await?;
```
ikopylov, line 296: This will go to local aliens for keys, that belongs to local node only.
Also it is better to perform aliens lookup only after the normal lookup passed

bob/src/cluster/quorum.rs
```diff
@@ -272,18 +272,64 @@ impl Cluster for Quorum {
...
+            }
+            debug!("exist after remote nodes {:?}", exist);
+        }
+
+        // filter remote not found keys
+        let remote_alien = filter_not_found(&exist, keys);
+        // end
+        if remote_alien.len() > 0 {
+            // filter remote nodes
+            let remote_nodes = filter_remote_nodes(&self.mapper);
```
ikopylov, line 323: This will lead to skipping local aliens for keys, that don't belong to the local node

bob/src/cluster/quorum.rs
```diff
@@ -294,3 +340,73 @@ impl Cluster for Quorum {
...
+    let mut not_found_keys = Vec::new();
+    for (idx, &r) in exist.iter().enumerate() {
+        if r == false {
+            not_found_keys.push(keys[idx]);
+        }
+    }
+    not_found_keys
+}
+
+fn filter_remote_not_found_by_nodes(exist: &[bool], keys: &[BobKey], mapper: &Virtual) -> HashMap<Vec<Node>, (Vec<BobKey>, Vec<usize>)> {
```
ikopylov, line 373: I don't like grouping by `Vec<Node>`: this is a strong indicator of poor program logic

ikopylov:
Additional changes required on backend level:
`exist` function in (`disk_controller.rs`)[https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/disk_controller.rs#L478] should be split into 2: one for normal `exist` and the other for aliens (`exist_alien`). This should be done in the same way as it is already done for `get`. Crucial difference is the error processing: for alien version, if the group is not found, then it is not an error and result must be a vector of `false` values. Also for normal version the error should be `vdisk_not_found` (not `internal`)

ikopylov:
You can introduce a struct responsible for indexes mapping. This will simplify operating with keys. Possible interface:
```Rust
struct IndexMap {
   indexes: Vec<usize>
}

impl IndexMap {
    fn len(&self) -> usize { indexes.len() }
    fn is_empty(&self) -> bool { /* true if empty */ }
    fn push(&mut self, item: usize) { /* adds new index */ }
    fn get<T>(&self, data: &[T], offset: usize) -> &T { data[indexes[offset]] }
    fn collect<T>(&self, data: &[T]>) -> Vec<T> { /* collect vector consists of data, present in indexes only */ }
    fn update_presence(&self, original: &mut [bool], mapped: &[bool]) { /* move values from mapped to original with |=. (original[indexes[i]] |= mapped[i])  */ }
    fn retain_not_presented(&mut self, original: &[bool]) { /* removes indexes for which original[indexes[i]] == true */ }
}
```

ikopylov:
Grouping logic should be updated. First, grouping can be performed only once. Second, grouping should be performed by `Node` (not by `Vec<Node>`). Backend can properly separate keys between vdisks on node.

To optimize keys lookup we can split all nodes into 3 categories:
1. Local node
2. Primary remote nodes (minimal set of nodes, that can respond for all the keys from request)
3. Secondary remote nodes (all other nodes, that can respond for the keys from request)

Separation into primary and secondary nodes is necessary due to replication. Primary set is smaller, so this allows to reduce the number of nodes to lookup.
Example: `Node1 (vdisk0, vdisk1), Node2 (vdisk2, vdisk3), Node3 (vdisk0, vdisk2), Node4 (vdisk1, vdisk3)`. 
1. Here for the key `1` requested from Node1 the split can be: `(local: Node1, primary: [], secondary: [Node4])`.
2. For the key `3` requested from Node1 the split can be: `(local: None, primary: [Node2], secondary: [Node4])`.
3. For the keys `1, 3` requested from Node1 the split can be: `(local: Node1, primary: [Node2], secondary: [Node4])`.
4. For the keys `1, 2, 3` requested from Node1 the split can be: `(local: Node1, primary: [Node2], secondary: [Node3, Node4])`.
5. For the keys `2, 3` requested from Node1 the split can be: `(local: None, primary: [Node2], secondary: [Node3, Node4])`.

Grouping logic pseudo-code:
```Rust
fn group_by_nodes(keys: &[BobKey], mapper: &Virtual) -> (Option<IndexMap>, HashSet<Node, IndexMap>, HashSet<Node, IndexMap>) {
   let local = IndexMap::new();
   let primary = HashSet::<Node, IndexMap>::new();
   let secondary = HashSet::<Node, IndexMap>::new();
   
   for (key, index) in keys {
       let target_nodes = mapper.get_target_nodes_for_key(key);
 
       // Add new primary node if required
       if !target_nodes.contains(local_node) && !target_nodes.contains_any(primary) {
           if let Some(not_present_node) = target_nodes.find_not_presented(secondary) {
               primary.add(not_present_node);
           }
       }
      
       // Fill keys for all nodes
       for target_node in target_nodes {
           if target_node == local_node {
               local.push(index);
           } else if let Some(primary_node_map) = primary.get(target_node) {
              primary_node_map.push(index);
           } else {
               let secondary_node_map = secondary.get_or_add_entry(target_node);
               secondary_node_map.push(index);
           }
       }
   }
   
   if local.is_empty() {
       return (None, primary, secondary); 
   }
   return (local, primary, secondary);
}
```

ikopylov:
General exists logic now can be expressed in this pseudo-code:
```Rust
fn exist(&self, keys: &[BobKey]) -> Vec<bool> {
    let mut result = vec![false; len];
    
    let (local, primary, secondary) = group_by_nodes(keys, mapper);

    // Local node lookup
    if let Some(local) = local && !local.is_empty() {
         let keys = local.collect(keys);
         let local_exist = exist_on_local_node(keys);
         local.update_presence(result, local_exist);
    } 

    // Primary nodes lookup
    if !primary.is_empty() {
         // Fill requests
         let nodes_and_keys = vec![];
         for (primary_node, primary_node_map) in primary {
              primary_node_map.retain_not_presented(result);
              if !primary_node_map.is_empty() {
                   nodes_and_keys.push((primary_node, primary_node_map.collect(keys));
              }
         }

         // Run requests simultaneously and update result
         if !nodes_and_keys.is_empty() {
              let remote_exist = exist_on_remote_nodes(nodes_and_keys);
              for (remote_node, remote_node_exist) in remote_exist {
                    primary.get(remote_node).update_presence(result, remote_node_exist);
              }
         }
    }

    // Secondary nodes lookup
    if !secondary.is_empty() {
          // Identical to primary and should be refactored into a separate function
    }

    IndexMap alien_index_map = IndexMap::from_not_presented_indexes(result);
    
    // Local node aliens lookup
    if !alien_index_map.is_empty() {
          let local_alien_exist = exist_on_local_alien(alien_index_map.collect(keys));
          alien_index_map.update_presence(result, local_alien_exist);
          alien_index_map.retain_not_presented(result);
    }

    // Remote nodes aliens lookup
    if !alien_index_map.is_empty() {
         let all_remote_nodes = mapper.get_all_remote_nodes();
         let remote_nodes_aliens_exist = exist_on_remote_aliens(all_remote_nodes, alien_index_map.collect(keys));
         for (remote_node, remote_node_aliens_exist) in remote_nodes_aliens_exist {
               alien_index_map.update_presence(result, remote_node_aliens_exist);
         }
    }
   
    return result;
}
```

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Bob versions changelog
 - Exist test on doubled range of keys for integration tests (#764)
 - Used swap and bob virtual memory metrics added to hardware metrics (#771)
 - Added validate_data_checksum_in_index_regen field to pearl config (#607)
+- 'Exists' method through HEAD request in REST API (#691)
+- Support for 'exists' method in bobt (#691)
```
ikopylov, line 15: Those 2 lines for #691 are duplicate and should be removed. These changes was merged within `v2.1.0-alpha.8`

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+            } else {
+                secondary
+                    .entry(node.clone())
+                    .or_insert(IndexMap::new())
+                    .push(index);
+            }
+        }
+    }
+
+    (
```
ikopylov, line 485: Don't like this syntax. Please, add `return` keyword to emphasis, that this is a result of a function

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        self.indexes
+            .iter()
+            .zip(mapped.iter())
+            .filter(|(i, _)| **i < max)
+            .for_each(|(i, f)| original[*i] |= f)
+    }
+
+    pub(crate) fn retain_not_existed(&mut self, original: &[bool]) {
+        self.indexes
+            .retain(|&i| i >= original.len() || original[i] == false);
```
ikopylov, line 543: It is expected, that `IndexMap` was build on top of passed `original`, thus filtering by `i >= original.len()` is not needed. Add `debug_assert!(i < original.len())` to enforce the expectation

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        data.into_iter()
+            .enumerate()
+            .filter(|(i, _)| self.indexes.contains(i))
+            .map(|(_, i)| i.clone())
+            .collect()
+    }
+
+    pub(crate) fn update_existence(&self, original: &mut [bool], mapped: &[bool]) {
+        let max = original.len();
+        self.indexes
```
ikopylov, line 534: Please, do this with regular `for` loop. That will guarantee the best possible performance

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+    pub(crate) fn push(&mut self, item: usize) {
+        debug_assert!(!self.indexes.contains(&item));
+        self.indexes.push(item);
+    }
+
+    pub(crate) fn collect<'a, T: Clone + 'a>(
+        &'a self,
+        data: impl IntoIterator<Item = &'a T>,
+    ) -> Vec<T> {
+        data.into_iter()
```
ikopylov, line 525: The complexity of implemented approach is `O(n*m)`. It is better to iterate over `indexes`, which will give the complexity `O(m)`. Also, please implement it with regular `for` loop to achieve the best possible performance. Additionally, the result vector can be pre-allocated with `Vec::with_capacity(self.indexes.len())`

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+pub(crate) struct IndexMap {
+    indexes: Vec<usize>,
+}
+
+impl IndexMap {
+    pub(crate) fn new() -> Self {
+        Self { indexes: vec![] }
+    }
+
+    pub(crate) fn where_not_exists(data: &[bool]) -> Self {
```
ikopylov, line 501: Please, add comments for all functions inside `IndexMap`

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+    }
+
+    pub(crate) fn push(&mut self, item: usize) {
+        debug_assert!(!self.indexes.contains(&item));
+        self.indexes.push(item);
+    }
+
+    pub(crate) fn collect<'a, T: Clone + 'a>(
+        &'a self,
+        data: impl IntoIterator<Item = &'a T>,
```
ikopylov, line 523: From performance point of view, it is better to pass a slice here

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+    }
+
+    (
+        if local.is_empty() { None } else { Some(local) },
+        primary,
+        secondary,
+    )
+}
+
+pub(crate) struct IndexMap {
```
ikopylov, line 492: Move `IndexMap` to `support_types` module

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+
+        Ok(result)
     }
 
     async fn delete(&self, key: BobKey, meta: &BobMeta) -> Result<(), Error> {
         self.delete_on_nodes(key, meta).await
     }
 }
+
+async fn collect_remote_exists(
```
ikopylov, line 411: For unification, I would prefer to move the function inside `impl Quorum {` above

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        debug!("EXIST {} keys", len);
+
+        let mut result = vec![false; len];
+
+        let (local, mut primary, mut secondary) = group_by_nodes(keys, &self.mapper);
+
+        if let Some(local) = local {
+            if !local.is_empty() {
+                let local_keys = local.collect(keys);
+                let local_exist = exist_on_local_node(&self.backend, &local_keys).await?;
```
ikopylov, line 365: It is better to use the same error processing approach as it is implemented in `get`: do not stop processing on the error here. If error happened, then it should be written to the log on `warn!` level and processing should continue

ikopylov: Please add logs so that the progress of the function can be tracked. 
Add log describing `exist_on_local_node`: on error log that error, on success log on `trace` level the number of keys passed and the number of keys presented locally

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+                local.update_existence(&mut result, &local_exist);
+            }
+        }
+
+        collect_remote_exists(&mut result, keys, &mut primary).await?;
```
ikopylov, line 370: 1. Write log before that function on trace level, that `primary` nodes processing started.
2. On error, log that error and continue the execution

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        collect_remote_exists(&mut result, keys, &mut secondary).await?;
```
ikopylov, line 371: Same here

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+
+        let alien_index_map = IndexMap::where_not_exists(&result);
+
+        if !alien_index_map.is_empty() {
+            let local_alien_result =
+                exist_on_local_alien(&self.backend, &alien_index_map.collect(keys)).await?;
```
ikopylov, line 377: Same here:
1. Write log on the result of this function call
2. Do not stop processing on error

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+            alien_index_map.update_existence(&mut result, &local_alien_result);
```
ikopylov, line 378: Call `alien_index_map.retain_not_existed()` to remove the values, that passed the check in local aliens. This will prevent unneeded remote aliens check later

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        }
+
+        if !alien_index_map.is_empty() {
+            let all_remote_nodes: Vec<_> = self
+                .mapper
+                .nodes()
+                .iter()
+                .filter(|n| n.name() != self.mapper.local_node_name())
+                .cloned()
```
ikopylov, line 387: Are you sure, that this `clone` is really needed? I think it is ok to pass reference to the `Node` struct to `exist_on_remote_aliens()`

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+                .collect();
+            let remote_nodes_aliens_exist =
+                exist_on_remote_aliens(&all_remote_nodes, &alien_index_map.collect(keys)).await;
```
ikopylov, line 390: Add log on the result of this call

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+            for remote_alien_result in remote_nodes_aliens_exist {
+                match remote_alien_result {
+                    Ok(remote_alien_result) => {
+                        alien_index_map.update_existence(&mut result, remote_alien_result.inner());
+                    }
+                    Err(e) => {
+                        debug!("Failed to check existence in aliens: {:?}", e);
```
ikopylov, line 397: Add node name to this log

bob/src/cluster/operations.rs
```diff
@@ -295,7 +276,48 @@ pub(crate) async fn put_local_node(
...
+    Ok(backend
+        .exist(keys, &BobGetOptions::new_get(Some(GetOptions::new_alien())))
+        .await?)
+}
+
+pub(crate) async fn exist_on_remote_nodes(
+    nodes: &[Node],
+    keys_by_node: HashMap<Node, Vec<BobKey>>,
+) -> Vec<Result<NodeOutput<Vec<bool>>, NodeOutput<Error>>> {
+    LinkManager::call_nodes(nodes.iter(), |client| {
```
ikopylov, line 303: Looks like you can use `keys_by_node.keys()` as the first parameter. In this way you don't need to pass `nodes` slice as a separate parameter

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+        let mut keys_by_node = HashMap::new();
+        for (node, node_map) in indexes_by_node.iter_mut() {
+            node_map.retain_not_existed(&result);
+            if !node_map.is_empty() {
+                keys_by_node.insert(node.clone(), node_map.collect(keys));
+            }
+        }
+
+        if !keys_by_node.is_empty() {
+            let nodes: Vec<_> = keys_by_node.keys().cloned().collect();
```
ikopylov, line 426: There is no need to `clone` nodes. You can pass reference to the `Node` struct

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+            let remote_results = exist_on_remote_nodes(&nodes, keys_by_node).await;
+            for (remote_result, node) in remote_results.into_iter().zip(nodes) {
+                match remote_result {
+                    Ok(remote_result) => {
+                        indexes_by_node
+                            .get(&node)
```
ikopylov, line 432: The node should be presented. This should be enforced by algorithm, thus it is better to use `expect()` to access the `IndexMap` for node

ikopylov: Additionally, add `debug_assert!(node.name() == remote_result.node_name())`

bob/src/cluster/quorum.rs
```diff
@@ -351,25 +352,194 @@ impl Cluster for Quorum {
...
+                            .map(|idx| idx.update_existence(result, remote_result.inner()));
```
ikopylov, line 433: Also, write summary of the operation to the log on `trace` level

bob/src/cluster/support_types.rs
```diff
@@ -57,3 +57,76 @@ impl From<RemoteDeleteError> for Error {
...
+    }
+
+    /// Collect data with allowed indexes
+    pub(crate) fn collect<T: Clone>(
+        &self,
+        data: &[T]
+    ) -> Vec<T> {
+        let mut result = Vec::with_capacity(self.indexes.len());
+        for &i in &self.indexes {
+            if i < data.len() {
```
ikopylov, line 106: Here it is expected, that `data` is the original data on top of which this `IndexMap` was built. Thus it should be impossible for `i` to larger or equal than `data.len()`. To enforce this I suggest to use `debug_assert!(i <data.len());` instead of condition

bob/src/cluster/support_types.rs
```diff
@@ -57,3 +57,76 @@ impl From<RemoteDeleteError> for Error {
...
+                result.push(data[i].clone());
+            }
+        }
+        result
+    }
+
+    /// Update elements in original according to internal indexes mapping
+    pub(crate) fn update_existence(&self, original: &mut [bool], mapped: &[bool]) {
+        let max = original.len();
+        let len = std::cmp::max(self.indexes.len(), mapped.len());
```
ikopylov, line 116: Same here. We expect, that `self.indexes.len() == mapped.len()`. This should be enforced with `assert!(self.indexes.len() == mapped.len());`. Here `assert` is better than `debug_assert`, because it allow to detect problem in the production build before entering the loop (in previous case in production the code would still fail on index-of-range)

bob/src/cluster/support_types.rs
```diff
@@ -57,3 +57,76 @@ impl From<RemoteDeleteError> for Error {
...
+        for i in 0..len {
+            let ind = self.indexes[i];
+            if ind < max {
```
ikopylov, line 119: `debug_assert!(ind < max)` here

bob/src/cluster/quorum.rs
```diff
@@ -351,22 +437,78 @@ impl Cluster for Quorum {
...
+        let (local, mut primary, mut secondary) = Self::group_by_nodes(keys, &self.mapper);
+
+        if let Some(local) = local {
+            if !local.is_empty() {
+                trace!("EXIST {} keys check local node", len);
+                let local_keys = local.collect(keys);
+                match exist_on_local_node(&self.backend, &local_keys).await {
+                    Ok(local_exist) => {
+                        trace!("EXIST {} keys check local node: found {}/{} keys",
+                               len, local_exist.len(), local.len());
```
ikopylov, line 454: 2nd argument should be `local_exist.filter(|v| v).count()`

bob/src/cluster/quorum.rs
```diff
@@ -318,6 +319,91 @@ impl Quorum {
...
+                node_map.retain_not_existed(&result);
+                if !node_map.is_empty() {
+                    keys_by_node.insert(node.clone(), node_map.collect(keys));
+                }
+            }
+
+            if !keys_by_node.is_empty() {
+                let nodes: Vec<_> = keys_by_node.keys().collect();
+                let remote_results = exist_on_remote_nodes(&nodes, &keys_by_node).await;
+                for (remote_result, node) in remote_results.into_iter().zip(nodes) {
```
ikopylov, line 340: Unfortunately, this `zip` is incorrect, because `exist_on_remote_nodes` calls `LinkManager::call_nodes` which uses `FuturesUnordered<_>` inside. Thus the order of the results is not preserved. 
The mapping should use node name stored within `remote_result`.

Possible solution: add `nodes_by_name: HashMap<String, Node>` to `Virtual` inside `mapper.rs` and function `fn node_by_name(&self, name: &str) -> Option<&Node>` to get node by its name. Then here you can pass only `keys_by_node` to `exist_on_remote_nodes` (`nodes` vector is not needed anymore). And to get the `IndexMap` you can use double indirection `keys_by_node.get(mapper.node_by_name(remote_result.node_name()))`

bob/src/cluster/quorum.rs
```diff
@@ -318,6 +319,91 @@ impl Quorum {
...
+                    match remote_result {
+                        Ok(remote_result) => {
+                            debug_assert!(node.name() == remote_result.node_name());
+                            indexes_by_node
+                                .get(node)
+                                .expect("node should exist")
+                                .update_existence(result, remote_result.inner());
+                            trace!("Check existence on node {}: found {} keys",
```
ikopylov, line 348: Add total number of keys to output, that was sent to the node: `found {}/{} keys`

bob/src/cluster/quorum.rs
```diff
@@ -318,6 +319,91 @@ impl Quorum {
         }
         Ok(())
     }
+
+    async fn collect_remote_exists(
+        result: &mut [bool],
+        keys: &[BobKey],
+        indexes_by_node: &mut HashMap<Node, IndexMap>,
+    ) -> Result<(), Error> {
```
ikopylov, line 327: Looks like this function does not return any error. It might be better to remove `Result<(), Error>` as return value

bob/src/cluster/quorum.rs
```diff
@@ -351,22 +437,78 @@ impl Cluster for Quorum {
...
+        }
+
+        let mut alien_index_map = IndexMap::where_not_exists(&result);
+
+        if !alien_index_map.is_empty() {
+            trace!("EXIST {} keys check local alien", len);
+            match exist_on_local_alien(&self.backend, &alien_index_map.collect(keys)).await {
+                Ok(local_alien_result) => {
+                    trace!("EXIST {} keys check local alien finished: found {}/{} keys",
+                           len, local_alien_result.len(), alien_index_map.len());
```
ikopylov, line 481: 2nd argument should be `local_alien_result.filter(|v| v).count()`

bob/src/cluster/quorum.rs
```diff
@@ -351,22 +437,78 @@ impl Cluster for Quorum {
...
+                .collect();
+            let remote_nodes_aliens_exist =
+                exist_on_remote_aliens(&all_remote_nodes, &alien_index_map.collect(keys)).await;
+            trace!("EXIST {} keys check remote alien finished", len);
+            for remote_alien_result in remote_nodes_aliens_exist {
+                match remote_alien_result {
+                    Ok(remote_alien_result) => {
+                        alien_index_map.update_existence(&mut result, remote_alien_result.inner());
+                    }
+                    Err(e) => warn!("EXIST {} keys check remote alien failed on node {}: {:?}",
```
ikopylov, line 505: Maybe it is better to log at the `debug` level, since if one node is unavailable, every request will write a log at `wran` level and overflow the log files

bob/src/cluster/operations.rs
```diff
@@ -295,7 +276,48 @@ pub(crate) async fn put_local_node(
...
+    backend: &Backend,
+    keys: &[BobKey],
+) -> Result<Vec<bool>, Error> {
+    Ok(backend
+        .exist(keys, &BobGetOptions::new_get(Some(GetOptions::new_alien())))
+        .await?)
+}
+
+pub(crate) async fn exist_on_remote_nodes(
+    nodes: &[&Node],
```
ikopylov, line 300: Looks like `nodes` is not needed. You can take `Node` instance from `keys_by_node`: `keys_by_node.values().map(|n| &n.0)`

bob/src/cluster/quorum.rs
```diff
@@ -318,6 +319,93 @@ impl Quorum {
...
+        let mut node_keys_by_node_name = HashMap::new();
+        for (node, node_map) in indexes_by_node.iter_mut() {
+            node_map.retain_not_existed(&result);
+            if !node_map.is_empty() {
+                node_keys_by_node_name.insert(node.name().to_owned(), (node.clone(), node_map.collect(keys)));
+            }
+        }
+
+        if !node_keys_by_node_name.is_empty() {
+            let nodes: Vec<_> = node_keys_by_node_name.values().map(|(n, _)| n).collect();
```
ikopylov, line 341: No need to collect nodes. `exist_on_remote_nodes` can use `Node` instances from `node_keys_by_node_name`

bob/src/cluster/quorum.rs
```diff
@@ -318,6 +319,93 @@ impl Quorum {
...
+            let remote_results = exist_on_remote_nodes(&nodes, &node_keys_by_node_name).await;
+            for remote_result in remote_results.into_iter() {
+                match remote_result {
+                    Ok(remote_result) => {
+                        let node = &node_keys_by_node_name.get(remote_result.node_name()).unwrap().0;
```
ikopylov, line 346: Use `expect()` instead of `unwrap()`. `expect` allows to pass a message, describing the problem

bob/src/cluster/operations.rs
```diff
@@ -295,7 +276,48 @@ pub(crate) async fn put_local_node(
...
+    keys_by_node: &HashMap<String, (Node, Vec<BobKey>)>,
+) -> Vec<Result<NodeOutput<Vec<bool>>, NodeOutput<Error>>> {
+    LinkManager::call_nodes(nodes.iter().map(|n| *n), |client| {
+        Box::pin(client.exist(
+            keys_by_node.get(client.node().name()).unwrap().1.clone(),
```
ikopylov, line 305: Use `expect` instead of `unwrap`

bob/src/cluster/quorum.rs
```diff
@@ -351,22 +439,74 @@ impl Cluster for Quorum {
...
+                .iter()
+                .filter(|n| n.name() != self.mapper.local_node_name())
+                .collect();
+            let remote_nodes_aliens_exist =
+                exist_on_remote_aliens(&all_remote_nodes, &alien_index_map.collect(keys)).await;
+            trace!("EXIST {} keys check remote alien finished", len);
+            for remote_alien_result in remote_nodes_aliens_exist {
+                match remote_alien_result {
+                    Ok(remote_alien_result) => {
+                        alien_index_map.update_existence(&mut result, remote_alien_result.inner());
```
ikopylov, line 501: Please, add `trace` log here, printing the number of keys found on node (`EXIST in aliens on node {}: found {}/{} keys`)

ikopylov (approved):
LGTM

ikopylov:
@idruzhitskiy please, backport this request to `release_v2.0.0`

idruzhitskiy (approved):

ikopylov (approved):
Trigger integration tests

PR #641, opened 2022-10-25
575 sequential errors counting

### LINKED ISSUE #575: Mark node as unavailable if several sequential requests failed due to network error
Currently, node is marked as unavailable by clearing its connection object by `ping` operation, that runs periodically (https://github.com/qoollo/bob/blob/master/bob-common/src/node.rs#L100). We can additionally control request errors and if there are more than 3 (possibly can be configured from `node.yaml`) sequential request errors with an error code related to a network problem, then we also can mark node as unavailable

idruzhitskiy:
Closes #575


bob-common/src/configs/cluster.rs
```diff
@@ -270,6 +270,12 @@ pub struct Cluster {
     racks: Vec<Rack>,
     #[serde(default)]
     distribution_func: DistributionFunc,
+    #[serde(default = "default_max_sequential_errors")]
+    max_sequential_errors: usize,
```
ikopylov, line 274: It should be in the `node.yaml`. This setting has nothing to do with the cluster configuration

bob-common/src/node.rs
```diff
@@ -68,6 +80,7 @@ impl Node {
 
     pub async fn set_connection(&self, client: BobClient) {
         *self.conn.write().await = Some(client);
+        self.reset_error_count();
```
ikopylov, line 83: This update should be under `write` lock

bob-common/src/node.rs
```diff
@@ -78,6 +91,17 @@ impl Node {
...
     }
 
+    pub fn reset_error_count(&self) {
+        self.err_count.store(0, Ordering::Relaxed);
+    }
+
+    pub async fn increase_error_and_clear_conn_if_needed(&self) {
+        let count = self.err_count.fetch_add(1, Ordering::Relaxed);
+        if count >= self.max_sequential_errors {
+            self.clear_connection().await;
```
ikopylov, line 101: This is incorrect because another thread may reset the counter. You must acquire the `write` lock and check the value, and only if it is still greater than `max_sequential_errors`, clear the connection.

bob-common/src/error.rs
```diff
@@ -38,6 +38,10 @@ impl Error {
         self.ctx == Kind::PossibleDiskDisconnection
     }
 
+    pub fn is_network_error(&self) -> bool {
+        self.ctx == Kind::Timeout || matches!(self.ctx, Kind::Failed(_))
```
ikopylov, line 42: `Timeout` is not a network error! It is specific for particular request (one can timeout, the following not). 
You should check actual error code from gRPC request. Not the high-level Bob error

bob/src/link_manager.rs
```diff
@@ -70,7 +70,16 @@ impl LinkManager {
...
     {
         match node.get_connection().await {
-            Some(conn) => f(&conn).await,
+            Some(conn) => match f(&conn).await {
+                Err(e) if e.inner().is_network_error() => {
+                    node.increase_error_and_clear_conn_if_needed().await;
+                    Err(e)
+                }
+                o => {
+                    node.reset_error_count();
```
ikopylov, line 79: The counter should only be reset if the operation is successfully completed

bob-common/src/node.rs
```diff
@@ -20,6 +23,8 @@ pub struct Node {
     address: String,
     index: Id,
     conn: Arc<RwLock<Option<BobClient>>>,
+    err_count: Arc<AtomicUsize>,
```
ikopylov, line 26: `Node` is bad type for this counter. It is better to count errors inside `BobClient`, because it is actually executes all the operations

bob-common/src/configs/node.rs
```diff
@@ -543,6 +543,8 @@ pub struct Node {
     bind_to_ip_address: Option<SocketAddr>,
     #[serde(default = "NodeConfig::default_holder_group_size")]
     holder_group_size: usize,
+    #[serde(default = "NodeConfig::default_max_sequential_errors")]
+    max_sequential_errors: usize,
```
ikopylov, line 547: Please, place this parameter closer to other network parameters: after `check_interval`

config-examples/cluster.yaml
```diff
@@ -28,4 +28,4 @@ vdisks:
           #[str] node name, to find target node in nodes list 
         - node: local_node
           #[str] disk name, to find location of the replica on the node in disks list
-          disk: disk1
\ No newline at end of file
+          disk: disk1
```
ikopylov, line 31: Please, revert this change

bob-common/src/configs/node.rs
ikopylov: Also, add this parameter to `node.yaml` in config-examples

bob-common/src/error.rs
```diff
@@ -168,6 +178,7 @@ impl From<Status> for Error {
                 "PearlChangeState" => Some(Self::pearl_change_state(rest_words(words, length))),
                 "Unauthorized" => Some(Self::unauthorized()),
                 "HolderTemporaryUnavailable" => Some(Self::holder_temporary_unavailable()),
+                _ if code == Code::Unavailable => Some(Self::network()),
```
ikopylov, line 181: This will never be chosen, because in the case of `Unavailable` error the `name` will be None` on the first `match`

bob-common/src/bob_client.rs
```diff
@@ -209,6 +212,14 @@ pub mod b_client {
                 })
         }
 
+        pub fn reset_error_count(&self) {
+            self.err_count.store(0, Ordering::Relaxed);
+        }
+
+        pub fn increase_error(&self) -> usize {
```
ikopylov, line 219: These 2 functions should not be public. The should be used within the functions of this struct: in `put`, `get`, `exist` and `delete`

ikopylov: You can add a public function to read `self.err_count` value

bob-common/src/error.rs
```diff
@@ -142,15 +150,17 @@ impl From<Error> for Status {
             )),
             Kind::DisksEventsLogger(msg) => {
                 Self::internal(format!("disk events logger error: {}", msg))
-            },
+            }
             Kind::Unauthorized => Self::unauthenticated("Unauthorized"),
             Kind::HolderTemporaryUnavailable => Self::unavailable("HolderTemporaryUnavailable"),
```
ikopylov, line 155: This should be converted to `internal`. `unavailable` is a bad status code

bob-common/src/node.rs
```diff
@@ -81,6 +91,19 @@ impl Node {
         self.conn.read().expect("rwlock").clone()
     }
 
+    pub async fn increase_error_and_clear_conn_if_needed(&self) {
```
ikopylov, line 94: This should is not needed if the counter is tracked by BobClient. 
Instead you need a function, that checks that counter to be greater than `max_sequential_errors` and resets the connection in this case. Check should be done in 2 phases:
```
if <check> {
    <write lock>
     if <check> {
          <reset connection>
     }
}
```

bob-common/src/error.rs
```diff
@@ -142,15 +150,17 @@ impl From<Error> for Status {
...
+            Kind::Network => Self::unavailable("network error"),
```
ikopylov, line 156: Even `Network` should not be converted to `unavailable`. `internal` is better

ikopylov:
Recent changes in ping logic should improve disconnect detection. Therefore, this request loses its significance

PR #645, opened 2022-10-30
464 corrupted blobs metric

### LINKED ISSUE #464: Add a metric to track the number of corrupted BLOBs

Boneyan:
Corrupted blobs count metric
Closes #464


bob-apps/Cargo.toml
```diff
@@ -51,7 +51,8 @@ default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-version = "0.13.0"
+git = "https://github.com/qoollo/pearl"
```
idruzhitskiy, line 54: Should be removed before merge

bob-backend/Cargo.toml
```diff
@@ -37,7 +37,8 @@ tower = "0.4"
 tower-service = "0.3"
 
 [dependencies.pearl]
-version = "0.13.0"
+git = "https://github.com/qoollo/pearl"
```
idruzhitskiy, line 40: Should be removed before merge

bob/Cargo.toml
```diff
@@ -46,7 +46,8 @@ tower-service = "0.3"
 uuid = "0.8.2"
 
 [dependencies.pearl]
-version = "0.13.0"
+git = "https://github.com/qoollo/pearl"
```
idruzhitskiy, line 49: Should be removed before merge

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -538,6 +538,24 @@ impl DiskController {
...
+                let holders = holders_guard.read().await;
+                for holder in holders.iter() {
+                    cnt += holder.corrupted_blobs_count().await;
+                }
+            }
+            cnt
+        } else {
+            0
+        };
+        self.blobs_count_cached.store(cnt as u64, Ordering::Release);
```
idruzhitskiy, line 555: Either variable name is a bit off, or this line should not be here

idruzhitskiy (changes requested):

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -538,6 +538,22 @@ impl DiskController {
         cnt
     }
 
+    pub(crate) async fn corrupted_blobs_count(&self) -> usize {
+        if *self.state.read().await == GroupsState::Ready {
+            let mut cnt: usize = 0;
+            for group in self.groups.read().await.iter() {
+                let holders_guard = group.holders();
+                let holders = holders_guard.read().await;
+                for holder in holders.iter() {
```
ikopylov, line 547: It is not ok to access `holders` on this level. It breaks SOLID principles. You should add `corrupted_blobs_count` function to the `Group` struct and iterate holders on the group level.

I also see that you have copied the code structure from the `blobs_count` function, so I ask you to also update `blobs_count` and `index_memory` functions

bob-backend/src/pearl/group.rs
```diff
@@ -498,6 +498,36 @@ impl Group {
         result
     }
 
+    pub(crate) async fn corrupted_blobs_count(&self) -> usize {
+        let mut corrupted_blobs = 0;
+        let holders_guard = self.holders();
```
ikopylov, line 503: This variable is not needed. You can get `holders` in one line: 
```
let holders = self.holders().read().await;
```

Boneyan: It will cause "temporary value dropped while borrowed" error, cause self.holders() value is dropped at the end of the statement

ikopylov: Ok. Thank you for explanation!

ikopylov (approved):
LGTM

bob-backend/src/pearl/group.rs
```diff
@@ -613,6 +613,36 @@ impl Group {
...
```
ikopylov, line 618: `self.holders` can be used directly here. In this way there is no need in `holders_guard` variable

ikopylov (approved):
LGTM

PR #658, opened 2022-11-11
651 fix bobt output

### LINKED ISSUE #651: `bobt` output should be directed to stdout
Currently, it prints text to stderr. Probably this is due to env_logger behavior and we should switch to simple `println`

Boneyan:
Logger output directed to stdout instead of stderr in bobt
Closes #651


ikopylov (approved):
LGTM

PR #661, opened 2022-11-15
660 remove deprecated chrono functions

### LINKED ISSUE #660: Fix usage of deprecated `chrono` functions
bob-backend/src/pearl/utils.rs

```
use of deprecated associated function `chrono::NaiveDateTime::from_timestamp`: use `from_timestamp_opt()` instead
```

```
use of deprecated associated function `chrono::DateTime::<Tz>::date`: Use `date_naive()` instead
```

```
use of deprecated associated function `chrono::Date::<Tz>::and_hms`: Use and_hms_opt() instead
```

Boneyan:
Replaced deprecated chrono functions
Closes #660

### DIFF
--- a/bob-backend/src/pearl/utils.rs
+++ b/bob-backend/src/pearl/utils.rs
@@ -111,16 +111,16 @@ impl Utils {
         if !config.round {
             return time.timestamp().try_into().unwrap();
         }
+        let time = time.naive_local();
         let mut start_time = match period {
-            period if period <= ChronoDuration::days(1) => time.date_naive().and_hms_opt(0, 0, 0).unwrap(),
+            period if period <= ChronoDuration::days(1) => time.date().and_hms_opt(0, 0, 0).unwrap(),
             period if period <= ChronoDuration::weeks(1) => {
-                let time = time.date_naive().and_hms_opt(0, 0, 0).unwrap();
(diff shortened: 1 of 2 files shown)


bob-backend/src/pearl/utils.rs
```diff
@@ -112,14 +112,15 @@ impl Utils {
...
+            period if period <= ChronoDuration::days(1) => time.date_naive().and_hms_opt(0, 0, 0).unwrap(),
             period if period <= ChronoDuration::weeks(1) => {
-                let time = time.date().and_hms(0, 0, 0);
+                let time = time.date_naive().and_hms_opt(0, 0, 0).unwrap();
                 time - ChronoDuration::days(i64::from(time.weekday().num_days_from_monday() - 1))
             }
             _ => panic!("pearid: {} is too large", period),
         };
 
+        let time = time.naive_utc();
```
ikopylov, line 123: You should convert `DateTime<Utc>` to `NaiveDateTime` earlier (right before `start_time` assignment). Currently it is not safe, because `date_naive` use local time according to the source code (which is still `UTC`, but depends on function argument), while `naive_utc` explicitly use UTC time. Furthermore, to preserve original behavior all calculations should be performed in local time, so `naive_local()` function should be used

ikopylov (approved):
LGTM

PR #670, opened 2022-11-24
572 join error

### LINKED ISSUE #572: Question: best JoinError handling
May be it is better to panic on JoinError here: https://github.com/qoollo/bob/blob/master/bob/src/cluster/operations.rs#L47

idruzhitskiy:
Closes #572

### DIFF
--- a/bob/src/cluster/operations.rs
+++ b/bob/src/cluster/operations.rs
@@ -57,6 +57,10 @@ fn is_result_successful(
         },
         Err(e) => {
             error!("{:?}", e);
+            errors.push(NodeOutput::new(
+                "unknown".to_string(),
+                Error::failed(e.to_string()),
+            ))
         }
     }
     0
@@ -278,10 +282,7 @@ pub(crate) async fn put_local_node(
     backend.put_local(key, data, op).await
 }
 
-pub(crate) async fn delete_at_local_node(
-    backend: &Backend,
-    key: BobKey,
-) -> Result<(), Error> {
+pub(crate) async fn delete_at_local_node(backend: &Backend, key: BobKey) -> Result<(), Error> {
     debug!("local node has vdisk replica, put local");
     backend.delete(key).await?;
     Ok(())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
+- Error propagation for JoinHandle errors (#572)
 
 
 #### Updated


bob/src/cluster/operations.rs
```diff
@@ -278,10 +282,7 @@ pub(crate) async fn put_local_node(
     backend.put_local(key, data, op).await
 }
 
-pub(crate) async fn delete_at_local_node(
-    backend: &Backend,
-    key: BobKey,
-) -> Result<(), Error> {
+pub(crate) async fn delete_at_local_node(backend: &Backend, key: BobKey) -> Result<(), Error> {
```
ikopylov, line 285: Please, revert this change

bob/src/cluster/operations.rs
```diff
@@ -57,6 +57,10 @@ fn is_result_successful(
         },
         Err(e) => {
             error!("{:?}", e);
+            errors.push(NodeOutput::new(
+                "unknown".to_string(),
```
ikopylov, line 61: This is incorrect. The actual node name should be here. To achieve that the `JoinError` should be processed within the function that calls `tokio::spawn` and should be converted to proper `NodeOutput<Error>` there

ikopylov: `JoinError` will be returned only on panic. Right now we are using `abort` on panic, so it is better to use `panic!` here too

idruzhitskiy: Where exactly we should panic, on specified error type?

ikopylov:
I searched for information about the `cancelled` error. There are comments in several places where it is indicated that there is a possibility that there will be a cancellation error when the runtime is stopped. Comments are here: 
1. https://github.com/tokio-rs/tokio/issues/2077#issuecomment-572671950
2. https://users.rust-lang.org/t/can-this-tokio-task-get-canceled-if-i-await-its-handle/88203/4

So, unfortunately, we should handle `cancelled` error in a special way

bob/src/cluster/support_types.rs
```diff
@@ -128,3 +128,27 @@ impl IndexMap {
...
+pub(crate) struct NodeOutputJoinHandle<Err>(JoinHandle<Result<NodeOutput<()>, NodeOutput<Err>>>, NodeName);
+
+impl<Err> NodeOutputJoinHandle<Err> {
+    pub(super) fn new(handle: JoinHandle<Result<NodeOutput<()>, NodeOutput<Err>>>, node_name: NodeName) -> Self {
+        Self(handle, node_name)
+    }
+}
+
+impl<Err> Future for NodeOutputJoinHandle<Err> {
+    type Output = Result<Result<NodeOutput<()>, NodeOutput<Err>>, JoinError>;
```
ikopylov, line 141: But the idea was to propagate the node name in case of `JoinError`. Right now this future basically do nothing: it replaces node name, but the original node name inside is already the correct one.

The proper output type should be `Result<NodeOutput<()>, NodeOutput<Err>>`

bob/src/cluster/operations.rs
```diff
@@ -188,13 +191,14 @@ fn call_node_put(
...
-) -> JoinHandle<Result<NodeOutput<()>, NodeOutput<Error>>> {
+) -> NodeOutputJoinHandle<Error> {
     debug!("PUT[{}] put to {}", key, node.name());
+    let name = node.name().clone();
     let task = async move {
         let grpc_options = options.to_grpc();
         LinkManager::call_node(&node, |conn| conn.put(key, data, grpc_options).boxed()).await
     };
-    tokio::spawn(task)
+    NodeOutputJoinHandle::new(tokio::spawn(task), name)
```
ikopylov, line 201: Same should be done for delete as well

PR #671, opened 2022-11-24
654 ensure correct working with multiple replicas

### LINKED ISSUE #654: Ensure correct working when node contains multiple replicas of single vdisk
It is perfectly valid to place vdisk replicas on a single node, but on different physical disks. Currently, `Virtual::get_operation` (https://github.com/qoollo/bob/blob/master/bob-common/src/mapper.rs#L185) returns only the first physical disk containing vdisk replica. Thus `quorum` (https://github.com/qoollo/bob/blob/master/bob/src/cluster/quorum.rs#L35) and `backend` (https://github.com/qoollo/bob/blob/master/bob-backend/src/core.rs#L243) writes data only to that one replica and always skips the other ones

lostl1ght:
Closes #654

### DIFF
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -51,7 +51,8 @@ default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-version = "0.14.0"
+# version = "0.14.0"
+git = "https://github.com/qoollo/pearl"
 
 [build-dependencies]
 tonic-build = "0.6"
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -47,6 +47,7 @@ lazy_static = "1.4"
 log = "0.4"
 log4rs = "1.2"
 metrics = { version = "0.17", features = ["std"] }
+network-interface = "0.1.2"
 mockall = "0.11"
 prost = "0.11"
 regex = "1.6.0"
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bob-apps"
-version = "2.1.0-alpha.5"
+version = "2.1.0-alpha.6"
 authors = [
     "Kirill Bushminkin <kb@qoollo.com>",
     "Pavel Iakushin <pyakushin@qoollo.com>",
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -1,25 +1,43 @@
 [package]
 name = "bob-apps"
-version = "2.1.0-alpha.6"
+version = "2.1.0-alpha.10"
 authors = [
     "Kirill Bushminkin <kb@qoollo.com>",
     "Pavel Iakushin <pyakushin@qoollo.com>",
+    "Nikita Menshov",
+    "Ilia Kopylov",
+    "Ivan Druzhitskiy",
+    "Vladimir Stepanenko",
+    "Pavel Perestoronin",
+    "Konstantin Konnov",
+    "Konstantin Bulany"
 ]
 edition = "2021"
+
+[features]
+default = ["mimalloc"]
+mimalloc = ["dep:mimalloc"]
+mimalloc-secure = ["dep:mimalloc", "mimalloc/secure"]
+# Important: can't be used with panic = "abort"
+# Please use profile "jemallocator-profile"
+jemallocator = ["dep:jemallocator"]
+jemallocator-profile = ["dep:jemallocator", "jemallocator/profiling"]
 
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
 axum = "0.4"
+base64 = "0.13.0"
 bincode = "1.3"
 bitflags = "1.3"
 bob = { path = "../bob" }
 bob-access = { path = "../bob-access" }
-bytes = "1.1"
+bob-common = { path = "../bob-common" }
+bytes = "1.2"
 cfg-if = "1.0"
 chrono = "0.4"
 clap = "2.34"
-crc = "1.8"
+crc = "3.0"
 env_logger = "0.9"
 futures = { version = "0.3", features = ["thread-pool"] }
 http = "0.2"
@@ -27,31 +45,38 @@ humantime = "2.1"
 hyper = { version = "0.14", default-features = false }
 lazy_static = "1.4"
 log = "0.4"
-log4rs = "1.0"
+log4rs = "1.2"
 metrics = { version = "0.17", features = ["std"] }
-mockall = "0.10"
-prost = "0.9"
+mockall = "0.11"
+prost = "0.11"
+regex = "1.6.0"
 ring = "0.16"
 serde = "1.0"
 serde_derive = "1.0"
 serde_yaml = "0.8"
-stopwatch = "0.0.7"
-termion = "1.5"
+serde_json = "1.0"
 thiserror = "1.0"
 tonic = { version = "0.6", features = ["prost", "tls"] }
 tower = "0.4"
 rand = "0.8"
 tower-service = "0.3"
-log4rs-logstash = { version = "0.1.0", features = ["rustls"] }
+hex = "0.4.3"
+qoollo-log4rs-logstash = { version = "0.2", features = ["rustls"] }
 reqwest = { version = "0.11", default-features = false }
 
 [dependencies.tokio]
-version = "1.14"
+version = "1.21"
 default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-version = "0.15.0"
+version = "0.19.0"
+
+[target.'cfg(all(target_env = "musl", target_arch = "x86_64", target_pointer_width = "64"))'.dependencies]
+mimalloc = { version = "0.1", default-features = false, optional = true }
+
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
+jemallocator = { version = "0.5.0", optional = true }
 
 [build-dependencies]
 tonic-build = "0.6"
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -51,8 +51,7 @@ default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-# version = "0.14.0"
-git = "https://github.com/qoollo/pearl"
+version = "0.15.0"
 
 [build-dependencies]
 tonic-build = "0.6"
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bob-backend"
-version = "2.1.0-alpha.5"
+version = "2.1.0-alpha.6"
 authors = [
     "Kirill Bushminkin <kb@qoollo.com>",
     "Pavel Iakushin <pyakushin@qoollo.com>",
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -37,8 +37,7 @@ tower = "0.4"
 tower-service = "0.3"
 
 [dependencies.pearl]
-# version = "0.14.0"
-git = "https://github.com/qoollo/pearl"
+version = "0.15.0"
 
 [dependencies.tokio]
 version = "1.14"
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -37,7 +37,8 @@ tower = "0.4"
 tower-service = "0.3"
 
 [dependencies.pearl]
-version = "0.14.0"
+# version = "0.14.0"
+git = "https://github.com/qoollo/pearl"
 
 [dependencies.tokio]
 version = "1.14"
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -1,21 +1,29 @@
 [package]
 name = "bob-backend"
-version = "2.1.0-alpha.6"
+version = "2.1.0-alpha.10"
 authors = [
     "Kirill Bushminkin <kb@qoollo.com>",
     "Pavel Iakushin <pyakushin@qoollo.com>",
+    "Nikita Menshov",
+    "Ilia Kopylov",
+    "Ivan Druzhitskiy",
+    "Vladimir Stepanenko",
+    "Pavel Perestoronin",
+    "Konstantin Konnov",
+    "Konstantin Bulany"
 ]
 edition = "2021"
 
 [dependencies]
 anyhow = "1.0"
+async-std = "1.12"
 async-trait = "0.1"
 bitflags = "1.3"
 bob-common = { path = "../bob-common", features = [] }
 bytes = "1.1"
 cfg-if = "1.0"
 chrono = "0.4"
-coarsetime = "0.1.21"
+coarsetime = "0.1.22"
 futures = { version = "0.3", features = ["thread-pool"] }
 http = "0.2"
 humantime = "2.1"
@@ -23,29 +31,31 @@ hyper = { version = "0.14", default-features = false }
 lazy_static = "1.4"
 log = "0.4"
 metrics = { version = "0.17", features = ["std"] }
-mockall = "0.10"
-parking_lot = "0.12.0"
-prost = "0.9"
+mockall = "0.11"
+prost = "0.11"
 ring = "0.16"
 serde = "1.0"
 serde_derive = "1.0"
 serde_yaml = "0.8"
-stopwatch = "0.0.7"
 thiserror = "1.0"
 tonic = { version = "0.6", features = ["prost"] }
 tower = "0.4"
 tower-service = "0.3"
+smallvec = { version = "1.10", features = ["union"] }
 
 [dependencies.pearl]
-version = "0.15.0"
+version = "0.19.0"
 
 [dependencies.tokio]
-version = "1.14"
+version = "1.21"
 default-features = false
 features = []
 
 [dev-dependencies]
-criterion = "0.3"
+criterion = "0.4"
+
+[features]
+async-io = ["pearl/async-io-rio"]
 
 [[bench]]
 name = "key_cmp_benchmark"
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -369,16 +369,14 @@ impl Backend {
             // Lookup local data
             if let Some(paths) = disk_paths {
                 trace!("GET[{}] try read normal", key);
-                let mut futures: FuturesUnordered<_> = paths.into_iter().map(|path| {
-                    self.get_local(key, Operation::new_local(vdisk_id, path))
-                }).collect();
-                while let Some(result) = futures.next().await {
-                    match result {
+                let mut error = None;
+                for path in paths {
+                    match self.get_local(key, Operation::new_local(vdisk_id, path)).await {
                         Ok(data) => return Ok(data),
-                        Err(_) => continue,
+                        Err(e) => error = Some(e),
                     }
                 }
-                Err(Error::internal())
+                Err(error.unwrap_or(Error::key_not_found(key)))
             } else {
                 error!("GET[{}] we read data but can't find path in config", key);
                 Err(Error::internal())
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -454,10 +454,12 @@ impl Backend {
     }
 
     pub async fn delete(&self, key: BobKey) -> Result<u64, Error> {
-        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
+        let (vdisk_id, disk_paths) = self.mapper.get_operation(key);
         let mut ops = vec![];
-        if let Some(path) = disk_path {
-            ops.push(Operation::new_local(vdisk_id, path.clone()));
+        if let Some(paths) = disk_paths {
+            for p in paths {
+                ops.push(Operation::new_local(vdisk_id, p))
+            }
         }
         ops.push(Operation::new_alien(vdisk_id));
         let total_count = futures::future::join_all(ops.into_iter().map(|op| {
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -267,13 +267,15 @@ impl Backend {
                 "remote nodes is empty, /{:.3}ms/",
                 sw.elapsed().as_secs_f64() * 1000.0
             );
-            for p in paths {
-                if let Err(e) = self.put_single(key, data, Operation::new_local(vdisk_id, p.clone())).await {
-                    warn!("PUT[{}] error put to {:?}: {:?}", key, p, e);
+            let mut result = Ok(());
+            for path in paths {
+                if let Err(e) = self.put_single(key, data, Operation::new_local(vdisk_id, path.clone())).await {
+                    warn!("PUT[{}] error put to {:?}: {:?}", key, path, e);
+                    result = Err(e);
                 }
             }
             trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
-            Ok(())
+            result
         } else {
             error!(
                 "PUT[{}] dont now what to do with data: op: {:?}. Data is not local and alien",
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -349,10 +349,8 @@ impl Backend {
             let mut all_result = Err(Error::key_not_found(key));
             if let Some(paths) = disk_paths {
                 trace!("GET[{}] try read normal", key);
-                let mut futures: FuturesUnordered<_> = paths.into_iter().map(|path| {
-                    self.get_local(key, Operation::new_local(vdisk_id, path))
-                }).collect();
-                while let Some(result) = futures.next().await {
+                for path in paths {
+                    let result = self.get_local(key, Operation::new_local(vdisk_id, path)).await;
                     match result {
                         Ok(data) => return Ok(data),
                         Err(_) => continue,
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -430,12 +430,12 @@ impl Backend {
         keys_by_operations
     }
 
-    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Operation> {
-        let (vdisk_id, path) = self.mapper.get_operation(key);
+    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Vec<Operation>> {
+        let (vdisk_id, paths) = self.mapper.get_operation(key);
         if options.get_normal() {
-            path.map(|path| Operation::new_local(vdisk_id, path))
+            paths.map(|paths| paths.into_iter().map(|path| Operation::new_local(vdisk_id, path)).collect())
         } else if options.get_alien() {
-            Some(Operation::new_alien(vdisk_id))
+            Some(vec![Operation::new_alien(vdisk_id)])
         } else {
             None
         }
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -240,10 +240,10 @@ impl Backend {
     pub async fn put(&self, key: BobKey, data: &BobData, options: BobOptions) -> Result<(), Error> {
         trace!(">>>>>>- - - - - BACKEND PUT START - - - - -");
         let sw = Stopwatch::start_new();
-        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
+        let (vdisk_id, disk_paths) = self.mapper.get_operation(key);
         trace!(
             "get operation {:?}, /{:.3}ms/",
-            disk_path,
+            disk_paths,
             sw.elapsed().as_secs_f64() * 1000.0
         );
         let res = if !options.remote_nodes().is_empty() {
@@ -257,16 +257,16 @@ impl Backend {
                 self.put_single(key, data, op).await?;
             }
             Ok(())
-        } else if let Some(path) = disk_path {
+        } else if let Some(paths) = disk_paths {
             debug!(
                 "remote nodes is empty, /{:.3}ms/",
                 sw.elapsed().as_secs_f64() * 1000.0
             );
-            let res = self
-                .put_single(key, data, Operation::new_local(vdisk_id, path))
-                .await;
-            trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
-            res
+            for p in paths {
+                self.put_single(key, data, Operation::new_local(vdisk_id, p)).await?;
+                trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
+            }
+            Ok(())
         } else {
             error!(
                 "PUT[{}] dont now what to do with data: op: {:?}. Data is not local and alien",
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -456,8 +456,8 @@ impl Backend {
         
         if options.get_normal() {
             if let Some(paths) = paths {
-                for p in paths {
-                    result.push(Operation::new_local(vdisk_id, p));
+                for path in paths {
+                    result.push(Operation::new_local(vdisk_id, path));
                 }
             }
         } 
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -15,8 +15,6 @@ use log::Level;
 
 use bob_common::{interval_logger::IntervalLoggerSafe, metrics::BLOOM_FILTERS_RAM};
 
-use futures::StreamExt;
-
 pub const BACKEND_STARTING: f64 = 0f64;
 pub const BACKEND_STARTED: f64 = 1f64;
 
--- a/bob-backend/src/core.rs
+++ b/bob-backend/src/core.rs
@@ -1,18 +1,19 @@
 use crate::prelude::*;
 use std::{
     collections::HashMap,
-    fmt::{Display, Formatter, Result as FMTResult}, hash::Hash,
+    fmt::{Display, Formatter, Result as FMTResult},
+    hash::Hash,
 };
+use smallvec::SmallVec;
 
 use crate::{
-    interval_logger::IntervalLoggerSafe,
     mem_backend::MemBackend,
     pearl::{DiskController, Pearl},
     stub_backend::StubBackend,
 };
 use log::Level;
 
-use bob_common::metrics::BLOOM_FILTERS_RAM;
+use bob_common::{interval_logger::IntervalLoggerSafe, metrics::BLOOM_FILTERS_RAM};
 
 use futures::StreamExt;
 
@@ -23,7 +24,7 @@ pub const BACKEND_STARTED: f64 = 1f64;
 pub struct Operation {
     vdisk_id: VDiskId,
     disk_path: Option<DiskPath>,
-    remote_node_name: Option<String>, // save data to alien/<remote_node_name>
+    remote_node_name: Option<NodeName>, // save data to alien/<remote_node_name>
 }
 
 impl Operation {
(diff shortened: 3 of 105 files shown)


bob-apps/Cargo.toml
```diff
@@ -51,7 +51,7 @@ default-features = false
 features = ["signal"]
 
 [dependencies.pearl]
-version = "0.14.0"
+version = "0.15.0"
```
ikopylov, line 54: Revert this change and use version from master

bob-backend/Cargo.toml
```diff
@@ -37,7 +37,7 @@ tower = "0.4"
 tower-service = "0.3"
 
 [dependencies.pearl]
-version = "0.14.0"
+version = "0.15.0"
```
ikopylov, line 40: Same here

bob-backend/src/core.rs
```diff
@@ -257,16 +259,16 @@ impl Backend {
...
                 "remote nodes is empty, /{:.3}ms/",
                 sw.elapsed().as_secs_f64() * 1000.0
             );
-            let res = self
-                .put_single(key, data, Operation::new_local(vdisk_id, path))
-                .await;
-            trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
-            res
+            for p in paths {
+                self.put_single(key, data, Operation::new_local(vdisk_id, p)).await?;
```
ikopylov, line 268: If there was an error on one disk, the operations on the others still need to be processed. So don't use `?` at the end

bob-backend/src/core.rs
```diff
@@ -257,16 +259,16 @@ impl Backend {
...
+                trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
```
ikopylov, line 269: This log is for the entire operation, not per drive, so you should move out it from the loop

bob-backend/src/core.rs
```diff
@@ -335,15 +337,23 @@ impl Backend {
...
                 trace!("GET[{}] try read normal", key);
-                self.get_local(key, Operation::new_local(vdisk_id, path.clone()))
-                    .await
+                let mut futures: FuturesUnordered<_> = paths.into_iter().map(|path| {
+                    self.get_local(key, Operation::new_local(vdisk_id, path))
+                }).collect();
+                while let Some(result) = futures.next().await {
+                    match result {
+                        Ok(data) => return Ok(data),
+                        Err(_) => continue,
```
ikopylov, line 353: You lost particular error here and silently convert it to `internal` at the end. This is not correct

bob-backend/src/core.rs
```diff
@@ -335,15 +337,23 @@ impl Backend {
...
 
         // we cannot get data from alien if it belong this node
 
         if options.get_normal() {
-            if let Some(path) = disk_path {
+            if let Some(paths) = disk_paths {
                 trace!("GET[{}] try read normal", key);
-                self.get_local(key, Operation::new_local(vdisk_id, path.clone()))
-                    .await
+                let mut futures: FuturesUnordered<_> = paths.into_iter().map(|path| {
```
ikopylov, line 347: Most of the time we will have one element in the `paths` collection. So using `FuturesUnordered<_>` will slow down 99% of practical use cases. I suggest using a normal for loop to simplify the code and avoid performance degradation

ikopylov:
Here is TODO in VDisk struct, that should be fixed: https://github.com/qoollo/bob/blob/master/bob-common/src/data.rs#L268
Otherwise we will get single node multiple times in that collection

bob-backend/src/core.rs
```diff
@@ -405,25 +415,27 @@ impl Backend {
...
-    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Operation> {
-        let (vdisk_id, path) = self.mapper.get_operation(key);
+    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Vec<Operation>> {
+        let (vdisk_id, paths) = self.mapper.get_operation(key);
         if options.get_normal() {
-            path.map(|path| Operation::new_local(vdisk_id, path))
+            paths.map(|paths| paths.into_iter().map(|path| Operation::new_local(vdisk_id, path)).collect())
         } else if options.get_alien() {
-            Some(Operation::new_alien(vdisk_id))
+            Some(vec![Operation::new_alien(vdisk_id)])
```
ikopylov, line 438: Call `get_vdisk_for_key` in this branch

bob-backend/src/core.rs
```diff
@@ -405,25 +415,27 @@ impl Backend {
...
+                }
             }
         }
         keys_by_operations
     }
 
-    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Operation> {
-        let (vdisk_id, path) = self.mapper.get_operation(key);
+    fn find_operation(&self, key: BobKey, options: &BobOptions) -> Option<Vec<Operation>> {
+        let (vdisk_id, paths) = self.mapper.get_operation(key);
```
ikopylov, line 434: Move this call into `if options.get_normal()` branch, because `paths` are only needed there

bob-backend/src/core.rs
```diff
@@ -442,10 +454,12 @@ impl Backend {
     }
 
     pub async fn delete(&self, key: BobKey) -> Result<u64, Error> {
```
ikopylov, line 456: This probably should be updated after https://github.com/qoollo/bob/pull/621 will be merged

bob-common/src/mapper.rs
```diff
@@ -179,23 +179,27 @@ impl Virtual {
...
-        if disk.is_none() {
+        }).collect();
+        let return_disks;
+        if disks.len() == 0 {
             debug!(
                 "cannot find node: {} for vdisk: {}",
                 self.local_node_name,
                 virt_disk.id()
             );
+            return_disks = None;
```
ikopylov, line 198: You don't need to introduce additional variable `return_disks`. You can just return value here: `(virt_disk.id(), None)`.
The same thing is in the branch below

bob/Cargo.toml
```diff
@@ -45,7 +45,7 @@ tower-service = "0.3"
 uuid = "0.8.2"
 
 [dependencies.pearl]
-version = "0.14.0"
+version = "0.15.0"
```
ikopylov, line 48: This should be reverted

bob-common/src/mapper.rs
```diff
@@ -179,23 +179,27 @@ impl Virtual {
             .collect()
     }
 
-    pub fn get_operation(&self, key: BobKey) -> (VDiskId, Option<DiskPath>) {
+    pub fn get_operation(&self, key: BobKey) -> (VDiskId, Option<Vec<DiskPath>>) {
         let virt_disk = self.get_vdisk_for_key(key).expect("vdisk not found");
-        let disk = virt_disk.replicas().iter().find_map(|disk| {
+        let disks: Vec<DiskPath> = virt_disk.replicas().iter().filter_map(|disk| {
```
ikopylov, line 184: Most of the time we will have 0 or 1 paths in the result. That's why I suggest to use regular for-loop. Initially, result should be `None` and on the first match it should be converted into `Some(Vec::with_capacity(1))`

bob/src/cluster/quorum.rs
```diff
@@ -32,17 +32,19 @@ impl Quorum {
...
-                failed_nodes.push(self.mapper.local_node_name().to_owned());
-            } else {
-                local_put_ok += 1;
-                at_least -= 1;
-                debug!("PUT[{}] local node put successful", key);
+            for p in paths {
+                let res = put_local_node(&self.backend, key, data, vdisk_id, p).await;
+                if let Err(e) = res {
+                    error!("{}", e);
+                    failed_nodes.push(self.mapper.local_node_name().to_owned());
```
ikopylov, line 42: You should push local node only once, if error happened. Otherwise `failed_nodes` will contain local node name multiple times

bob-backend/src/core.rs
idruzhitskiy: We do need vdisk_id for alien though

bob-backend/src/core.rs
ikopylov: As I wrote, `for` loop would be better here:
```
for path in paths {
    self.get_local(...).await
}
```

bob-backend/src/core.rs
```diff
@@ -338,14 +342,22 @@ impl Backend {
...
-            if let Some(path) = disk_path {
+            if let Some(paths) = disk_paths {
                 trace!("GET[{}] try read normal", key);
-                all_result = self.get_single(key, Operation::new_local(vdisk_id, path.clone())).await;
+                let mut futures: FuturesUnordered<_> = paths.into_iter().map(|path| {
+                    self.get_local(key, Operation::new_local(vdisk_id, path))
+                }).collect();
+                while let Some(result) = futures.next().await {
+                    match result {
+                        Ok(data) => return Ok(data),
```
ikopylov, line 357: All operations should compete. `return` here is incorrect

bob-backend/src/core.rs
```diff
@@ -355,9 +367,16 @@ impl Backend {
...
         else if options.get_normal() {
             // Lookup local data
-            if let Some(path) = disk_path {
+            if let Some(paths) = disk_paths {
                 trace!("GET[{}] try read normal", key);
-                self.get_single(key, Operation::new_local(vdisk_id, path.clone())).await
+                let mut error = None;
+                for path in paths {
+                    match self.get_local(key, Operation::new_local(vdisk_id, path)).await {
+                        Ok(data) => return Ok(data),
```
ikopylov, line 375: All operations should complete. `return` here is incorrect

bob-backend/src/core.rs
```diff
@@ -260,16 +262,18 @@ impl Backend {
...
-                .put_single(key, data, Operation::new_local(vdisk_id, path))
-                .await;
+            for p in paths {
+                if let Err(e) = self.put_single(key, data, Operation::new_local(vdisk_id, p.clone())).await {
+                    warn!("PUT[{}] error put to {:?}: {:?}", key, p, e);
+                }
+            }
             trace!("put single, /{:.3}ms/", sw.elapsed().as_secs_f64() * 1000.0);
-            res
+            Ok(())
```
ikopylov, line 276: Errors are lost here. We must return an error if the write to at least one replica ended with error

bob-backend/src/core.rs
```diff
@@ -427,17 +446,19 @@ impl Backend {
...
+        let capacity = if options.get_normal() { paths.as_ref().map(|v| v.len()).unwrap_or_default() } else { 0 } +
                               if options.get_alien() { 1 } else { 0 };
 
         let mut result = SmallVec::with_capacity(capacity);
         
         if options.get_normal() {
-            if let Some(path) = path {
-                result.push(Operation::new_local(vdisk_id, path));
+            if let Some(paths) = paths {
+                for p in paths {
```
ikopylov, line 459: `p` -> `path`

bob-backend/src/core.rs
```diff
@@ -260,16 +262,18 @@ impl Backend {
...
-        } else if let Some(path) = disk_path {
+        } else if let Some(paths) = disk_paths {
             debug!(
                 "remote nodes is empty, /{:.3}ms/",
                 sw.elapsed().as_secs_f64() * 1000.0
             );
-            let res = self
-                .put_single(key, data, Operation::new_local(vdisk_id, path))
-                .await;
+            for p in paths {
```
ikopylov, line 270: `p` -> `path`

bob-backend/src/core.rs
ikopylov: In `delete` this is implemented correctly

bob-common/src/mapper.rs
```diff
@@ -291,23 +291,26 @@ impl Virtual {
             .collect()
     }
 
-    pub fn get_operation(&self, key: BobKey) -> (VDiskId, Option<DiskPath>) {
+    pub fn get_operation(&self, key: BobKey) -> (VDiskId, Option<Vec<DiskPath>>) {
```
ikopylov, line 294: Please, use `SmallVec<[DiskPath; 1]>` here instead of `Vec<DiskPath>`. `SmallVec` stores up to 1 element on the stack, and in 99% of cases we will have just one element

bob/src/cluster/quorum.rs
```diff
@@ -419,9 +421,13 @@ impl Cluster for Quorum {
     //todo check no data (no error)
     async fn get(&self, key: BobKey) -> Result<BobData, Error> {
         debug!("GET[{}] ~~~LOOKUP LOCAL NODE~~~", key);
-        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
-        if let Some(data) = lookup_local_node(&self.backend, key, vdisk_id, disk_path).await {
-            return Ok(data);
+        let (vdisk_id, disk_paths) = self.mapper.get_operation(key);
+        if let Some(paths) = disk_paths {
+            for p in paths {
```
ikopylov, line 426: `p` -> `path`

bob/src/cluster/quorum.rs
```diff
@@ -32,31 +32,31 @@ impl Quorum {
...
     async fn put_at_least(&self, key: BobKey, data: &BobData) -> Result<(), Error> {
         let mut local_put_ok = 0_usize;
-        let at_least = self.quorum;
+        let mut at_least = self.quorum;
         let mut failed_nodes = Vec::new();
-        let (vdisk_id, disk_path) = self.mapper.get_operation(key);
-        let (tasks, errors) =
-            if let Some(disk_path) = disk_path {
-                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES AND LOCAL NODE~~~", key, at_least - 1);
-                let (remote_put, local_put) = tokio::join!(
```
ikopylov, line 41: Parallelization should be returned. This was an important optimization.
I think this can be done by adding a new function to `operations.rs` that puts in multiple paths. In this case, here the `tokio::join!`  with `put_remote_nodes` and this new function can be called

bob/src/cluster/quorum.rs
```diff
@@ -32,31 +32,31 @@ impl Quorum {
...
+        } else {
+            debug!("skip local put");
+        }
+        debug!("PUT[{}] need at least {} additional puts", key, at_least);
+
+        debug!("PUT[{}] ~~~PUT TO REMOTE NODES~~~", key);
+        let (tasks, errors) = self.put_remote_nodes(key, data, at_least).await;
         let all_count = self.mapper.get_target_nodes_for_key(key).len();
         let remote_ok_count = all_count - errors.len() - tasks.len() - local_put_ok;
         failed_nodes.extend(errors.iter().map(|e| e.node_name().clone()));
```
ikopylov, line 62: There is a big problem in the line below: `remote_ok_count + local_put_ok >= self.quorum`.
`get_target_nodes_for_key` returns unique remote nodes, and if one of these nodes contains multiple vdisk replicas, then that node should be counted multiple times. 
Example: `quorum = 2`, `number of replicas = 2`. If both replicas of vdisk are on the one node, then `remote_ok_count = 1`, `local_put_ok = 0`. In this case the condition `remote_ok_count + local_put_ok >= self.quorum` will be `false` and this leads to internal error returned from `put_aliens`.

Right now I don't know what is the best way to resolve this problem

ikopylov: `VDisk` struct contains both nodes list and replicas list: https://github.com/qoollo/bob/blob/master/bob-common/src/core_types.rs#L45 Therefore, we can check if their lengths are equal or not. If not, then for each request we call a separate function that calculates the number of vdisk replicas for specific node. If equal, obviously, every requests adds 1 to `remote_ok_count`

idruzhitskiy: So put to node should return count of written replicas?

ikopylov: Adding info to gRPC is the best way to solve the problem, but I think that changing the gRPC protocol within the current version is not worth doing. Right now it is better to treat error as if no one replica was written, and success as all replicas was written. 

I've created an issue for future to fix the problem properly: https://github.com/qoollo/bob/issues/818

idruzhitskiy: I've added passing affected replicas to results so we can count actual number of replicas written

bob-common/src/mapper.rs
```diff
@@ -291,23 +292,26 @@ impl Virtual {
...
-                Some(DiskPath::from(disk))
-            } else {
-                None
+        let mut disks = None;
+        for replica in virt_disk.replicas() {
+            if replica.node_name() == &self.local_node_name {
+                if disks.is_none() {
+                    disks = Some(SmallVec::<[DiskPath; 1]>::new());
+                }
+                disks.as_mut().unwrap().push(DiskPath::from(replica));
```
ikopylov, line 303: Use `expect()` instead of unwrap, because it is expected, that `disks` is not `None` here

bob-common/src/mapper.rs
```diff
@@ -317,4 +321,24 @@ impl Virtual {
             .iter()
             .any(|node| node.name() == node_name)
     }
+
+    pub fn get_replicas_count(&self, node_name: &str, key: BobKey) -> usize {
```
ikopylov, line 325: `node_name` should have `NodeName` type

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,34 @@ impl Quorum {
...
-                    debug!("PUT[{}] local node put successful", key);
+        let (vdisk_id, disk_paths) = self.mapper.get_operation(key);
+        let affected_replicas_by_node = self.mapper.get_replicas_count_by_node(key);
+        let (tasks, errors) = 
+            if let Some(paths) = disk_paths {
+                let paths_len = paths.len();
+                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES AND LOCAL NODE~~~", key, at_least - paths_len);
+                let (local_puts, (tasks, errors)) = tokio::join!(
+                    put_local_node_all(&self.backend, key, data, vdisk_id, paths),
+                    self.put_remote_nodes(key, data, at_least, &affected_replicas_by_node));
```
ikopylov, line 45: Please, make `put_remote_nodes(...)` first in the join. It is very likely that `join!` processes futures sequentially. Local put can synchronously goes right to the disk, while network put is guaranteed to start in the background (we use `spawn`). Therefore, it is better to start background task first and only after that write data locally

bob-common/src/node.rs
```diff
@@ -46,6 +46,7 @@ struct NodeInner {
 pub struct Output<T> {
     node_name: NodeName,
     inner: T,
+    affected_replicas: usize
```
ikopylov, line 49: I don't like that this field is now in the results of all operations, although it is only needed for the put operation.
There are 2 possible ways to avoid that:
1. Change `PutResult` to something like `Result<NodeOutput<PutResponse>, NodeOutput<PutError>>`, where `PutResponse` will contain `affected_replicas` and `PutError` will contain `affected_replicas` + `error`.
2. Perform similar changes, but on the `operations.rs` level. Declare `RemotePutError` and `RemotePutResponse` in `support_types.rs`, add `affected_replicas` to these structures, and convert put result from `BobClient` to these types (something similar is done right now for `delete` where regular error is converted to `RemoteDeleteError`)

Personally, I'd prefer the second way, because in this way the network code is not affected

bob/src/cluster/quorum.rs
```diff
@@ -143,6 +146,7 @@ impl Quorum {
         mut failed_nodes: Vec<NodeName>,
         key: BobKey,
         data: &BobData,
+        affected_replicas_by_node: &HashMap<NodeName, usize>
```
ikopylov, line 149: `affected_replicas_by_node` is not needed for aliens, because right now there is no way to save multiple replicas in same alien folder. That's because we have `node_name` -> `vdisk` folder structure

bob/src/cluster/simple.rs
```diff
@@ -90,6 +90,7 @@ impl Cluster for Quorum {
                     force_node: true,
                     overwrite: false,
                 },
+                self.mapper.get_replicas_count(c.target_node_name().as_str(), key)
```
ikopylov, line 93:
```suggestion
                self.mapper.get_replicas_count(c.target_node_name(), key)
```

bob-common/src/mapper.rs
```diff
@@ -317,4 +321,24 @@ impl Virtual {
             .iter()
             .any(|node| node.name() == node_name)
     }
+
+    pub fn get_replicas_count(&self, node_name: &NodeName, key: BobKey) -> usize {
+        let id = self.vdisk_id_from_key(key);
+        if let Some(vdisk) = self.vdisks.get(&id) {
+            vdisk.replicas().iter().filter(|r| r.node_name().as_str() == node_name).count()
```
ikopylov, line 328:
```suggestion
            vdisk.replicas().iter().filter(|r| r.node_name() == node_name).count()
```

bob/src/cluster/support_types.rs
```diff
@@ -25,6 +25,38 @@ impl<T: Eq + core::hash::Hash> HashSetExt<T> for HashSet<T> {
...
+    pub(super) fn affected_replicas(&self) -> usize {
+        self.affected_replicas
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct RemotePutError {
+    affected_replicas: usize,
+    #[allow(dead_code)] // We need error for debug print
+    error: Error
```
ikopylov, line 47: Add also a function, that returns this error. It would be better to make it visible, that error can be accessed outside

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,34 @@ impl Quorum {
...
+                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES AND LOCAL NODE~~~", key, at_least - paths_len);
+                let ((tasks, errors), local_puts) = tokio::join!(
+                    self.put_remote_nodes(key, data, at_least, &affected_replicas_by_node),
+                    put_local_node_all(&self.backend, key, data, vdisk_id, paths));
+                if local_puts != paths_len {
+                    failed_nodes.push(self.mapper.local_node_name().to_owned());
                 }
-                (remote_tasks, errors)
+                local_put_ok += local_puts;
+                at_least -= local_puts;
```
ikopylov, line 50: Don't see a reason to update `at_least`, because later it only used once id `debug` log, where the same value can be calculated on the fly (`at_least - local_puts`)

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,34 @@ impl Quorum {
...
-                self.put_remote_nodes(key, data, at_least).await
+                debug!("PUT[{}] ~~~PUT TO REMOTE NODES~~~", key);
+                self.put_remote_nodes(key, data, at_least, &affected_replicas_by_node).await
             };
-        let all_count = self.mapper.get_target_nodes_for_key(key).len();
-        let remote_ok_count = all_count - errors.len() - tasks.len() - local_put_ok;
+        debug!("PUT[{}] need at least {} additional puts", key, at_least);
+        let target_nodes = self.mapper.get_target_nodes_for_key(key);
+        let all_count: usize = target_nodes.iter()
+            .map(|n| *affected_replicas_by_node.get(n.name()).unwrap_or(&1))
```
ikopylov, line 59: All node names returned by `get_target_nodes_for_key(key)` should be presented inside `affected_replicas_by_node`, so it is better to use `expect("...")` instead of `unwrap_or(..)`

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,34 @@ impl Quorum {
...
+            .sum();
+        let errored: usize = errors.iter().map(|o| o.inner().affected_replicas()).sum();
+        let remote_ok_count = all_count - errored - tasks.len() - local_put_ok;
```
ikopylov, line 62: This calculation has an error: `tasks.len()` can't be used here, because every task created for separate node, but that node can still have multiple replicas. It is better to return the vector of successful results from `put_remote_nodes` like it is done for errors right now. In this way we can count the successfully written replicas number

bob/src/cluster/quorum.rs
```diff
@@ -125,7 +126,8 @@ impl Quorum {
         key: BobKey,
         data: &BobData,
         at_least: usize,
-    ) -> (Tasks<Error>, Vec<NodeOutput<Error>>) {
+        affected_replicas_by_node: &HashMap<NodeName, usize>
+    ) -> (Tasks<RemotePutResponse, RemotePutError>, Vec<NodeOutput<RemotePutError>>) {
```
ikopylov, line 130: It is better to add successful results as well here:
```suggestion
    ) -> (Tasks<RemotePutResponse, RemotePutError>, Vec<NodeOutput<RemotePutResponse>>, Vec<NodeOutput<RemotePutError>>) {
```

bob/src/cluster/operations.rs
```diff
@@ -204,10 +209,12 @@ pub(crate) async fn put_at_least(
     target_nodes: impl Iterator<Item = &Node>,
     at_least: usize,
     options: BobPutOptions,
-) -> (Tasks<Error>, Vec<NodeOutput<Error>>) {
+    affected_replicas_by_node: &HashMap<NodeName, usize>
+) -> (Tasks<RemotePutResponse, RemotePutError>, Vec<NodeOutput<RemotePutError>>) {
```
ikopylov, line 213:
```suggestion
) -> (Tasks<RemotePutResponse, RemotePutError>, Vec<NodeOutput<RemotePutResponse>>, Vec<NodeOutput<RemotePutError>>) {
```

bob/src/cluster/operations.rs
```diff
@@ -43,14 +45,15 @@ pub(crate) async fn finish_at_least_handles<TErr: Debug>(
...
 
-async fn call_at_least<TOp, TErr: Debug>(
+async fn call_at_least<TOp, TRes, TErr: Debug>(
     target_nodes: impl Iterator<Item = TOp>,
     at_least: usize,
-    f: impl Fn(TOp) -> JoinHandle<Result<NodeOutput<()>, NodeOutput<TErr>>>,
-) -> (FuturesUnordered<JoinHandle<Result<NodeOutput<()>, NodeOutput<TErr>>>>, Vec<NodeOutput<TErr>>) {
+    f: impl Fn(TOp) -> JoinHandle<Result<NodeOutput<TRes>, NodeOutput<TErr>>>,
+    get_affected_count: impl Fn(TRes) -> usize
+) -> (FuturesUnordered<JoinHandle<Result<NodeOutput<TRes>, NodeOutput<TErr>>>>, Vec<NodeOutput<TErr>>) {
```
ikopylov, line 53: Here it is better to start return also a vector of successful results:
```suggestion
) -> (FuturesUnordered<JoinHandle<Result<NodeOutput<TRes>, NodeOutput<TErr>>>>, Vec<NodeOutput<TRes>>, Vec<NodeOutput<TErr>>) {
```

bob/src/cluster/operations.rs
```diff
@@ -43,14 +45,15 @@ pub(crate) async fn finish_at_least_handles<TErr: Debug>(
...
 }
 
-async fn call_at_least<TOp, TErr: Debug>(
+async fn call_at_least<TOp, TRes, TErr: Debug>(
     target_nodes: impl Iterator<Item = TOp>,
     at_least: usize,
-    f: impl Fn(TOp) -> JoinHandle<Result<NodeOutput<()>, NodeOutput<TErr>>>,
-) -> (FuturesUnordered<JoinHandle<Result<NodeOutput<()>, NodeOutput<TErr>>>>, Vec<NodeOutput<TErr>>) {
+    f: impl Fn(TOp) -> JoinHandle<Result<NodeOutput<TRes>, NodeOutput<TErr>>>,
+    get_affected_count: impl Fn(TRes) -> usize
```
ikopylov, line 52: `call_at_least` should not know anything about replicas count. It is a general function. It is better to propagate the number of affected replicas through the vector with successful results

bob/src/cluster/operations.rs
```diff
@@ -26,15 +27,16 @@ fn is_result_successful<TErr: Debug>(
     0
 }
 
-pub(crate) async fn finish_at_least_handles<TErr: Debug>(
-    handles: &mut Tasks<TErr>,
+pub(crate) async fn finish_at_least_handles<TRes, TErr: Debug>(
+    handles: &mut Tasks<TRes, TErr>,
     at_least: usize,
+    get_affected_count: impl Fn(TRes) -> usize
```
ikopylov, line 33: `finish_at_least` should not know anything about affected replicas count. Instead it should return 2 vectors: 
1. With successful results
2. With errors

bob/src/cluster/operations.rs
```diff
@@ -1,19 +1,20 @@
...
+pub(crate) type Tasks<Res, Err> = FuturesUnordered<JoinHandle<Result<NodeOutput<Res>, NodeOutput<Err>>>>;
 
 // ======================= Helpers =================
 
-fn is_result_successful<TErr: Debug>(
-    join_res: Result<Result<NodeOutput<()>, NodeOutput<TErr>>, JoinError>,
+fn is_result_successful<TRes, TErr: Debug>(
+    join_res: Result<Result<NodeOutput<TRes>, NodeOutput<TErr>>, JoinError>,
     errors: &mut Vec<NodeOutput<TErr>>,
+    get_affected_count: impl Fn(TRes) -> usize
```
ikopylov, line 12: Instead of passing a function as parameter, it would be better to declare a trait, that has function returning the number of affected replicas, and add restriction that `TRes` should support that trait. 
After that you can implement that trait on `NodeOutput<()>` by returning `1` and on `NodeOutput<RemotePutResponse>` returning the actual affected replicas count

bob/src/cluster/quorum.rs
idruzhitskiy: We don't have `local_puts` later, it's inside if scope, I think it's better to leave modification here

bob/src/cluster/support_types.rs
idruzhitskiy: Another statement with allowed dead code? :) Why not defer it until we actually need this error?

bob/src/cluster/operations.rs
```diff
@@ -26,33 +48,34 @@ fn is_result_successful<TErr: Debug>(
...
-pub(crate) async fn finish_at_least_handles<TErr: Debug>(
-    handles: &mut Tasks<TErr>,
+pub(crate) async fn finish_at_least_handles<TRes: AffectedReplicasProvider, TErr: Debug>(
+    handles: &mut Tasks<TRes, TErr>,
     at_least: usize,
-) -> Vec<NodeOutput<TErr>> {
+) -> (Vec<NodeOutput<TRes>>, Vec<NodeOutput<TErr>>) {
     let mut ok_count = 0;
     let mut errors = Vec::new();
+    let mut oks = Vec::new();
```
ikopylov, line 57: I suggest to create this vector with initial capacity = `at_least.min(handles.len())`

bob/src/cluster/operations.rs
```diff
@@ -61,10 +84,11 @@ async fn finish_all_handles<TErr: Debug>(
 ) -> Vec<NodeOutput<TErr>> {
     let mut ok_count = 0;
     let mut total_count = 0;
+    let mut oks = Vec::new();
```
ikopylov, line 87: Most of the time all `handles` will complete with success, so I suggest to create this vector with initial capacity = `handles.len()`

bob/src/cluster/quorum.rs
ikopylov: But you have `local_put_ok`, that equals to `local_puts`

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,29 @@ impl Quorum {
...
-                    errors.extend(finish_at_least_handles(&mut remote_tasks, 1).await.into_iter());
-                } else {
-                    local_put_ok += 1;
-                    debug!("PUT[{}] local node put successful", key);
+        let (vdisk_id, disk_paths) = self.mapper.get_operation(key);
+        let affected_replicas_by_node = self.mapper.get_replicas_count_by_node(key);
+        let (tasks, oks, errors) = 
+            if let Some(paths) = disk_paths {
+                let paths_len = paths.len();
+                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES AND LOCAL NODE~~~", key, at_least - paths_len);
```
ikopylov, line 42:
```suggestion
                debug!("PUT[{}] ~~~PUT {} REPLICAS TO REMOTE NODES AND {} REPLICAS TO LOCAL NODE~~~", key, at_least - paths_len, paths_len);
```

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,29 @@ impl Quorum {
...
+                    failed_nodes.push(self.mapper.local_node_name().to_owned());
                 }
-                (remote_tasks, errors)
+                local_put_ok += local_puts;
+                at_least -= local_puts;
+                (tasks, oks, errors)
             } else {
-                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES~~~", key, at_least);
-                self.put_remote_nodes(key, data, at_least).await
+                debug!("PUT[{}] ~~~PUT TO REMOTE NODES~~~", key);
```
ikopylov, line 53:
```suggestion
                debug!("PUT[{}] ~~~PUT {} REPLICAS TO REMOTE NODES~~~", key, at_least);
```

bob/src/cluster/quorum.rs
```diff
@@ -32,33 +32,29 @@ impl Quorum {
...
+                self.put_remote_nodes(key, data, at_least, &affected_replicas_by_node).await
             };
-        let all_count = self.mapper.get_target_nodes_for_key(key).len();
-        let remote_ok_count = all_count - errors.len() - tasks.len() - local_put_ok;
+        debug!("PUT[{}] need at least {} additional puts", key, at_least);
```
ikopylov, line 56: This log is outdated, because at this point all remote puts has already done. I suggest to remove that log and add another, that contains the number of successful puts and number of errors.
```
debug!("PUT[{}] LOCAL PUT OK: {}, REMOTE PUT OK: {}, REMOTE PUT ERRORS: {}", key, local_put_ok, remote_ok_count, remote_error_count);
```

bob/src/cluster/support_types.rs
ikopylov: Only function will have `#[allow(dead_code)]`. The field in this case is not needed to be marked with this attribute

ikopylov: Actually it makes no difference here. So you can do it according to your own preference

bob-common/src/mapper.rs
```diff
@@ -317,4 +321,24 @@ impl Virtual {
             .iter()
             .any(|node| node.name() == node_name)
     }
+
+    pub fn get_replicas_count(&self, node_name: &NodeName, key: BobKey) -> usize {
```
ikopylov, line 325: Looks like this function is not used anywhere

bob/src/cluster/quorum.rs
```diff
@@ -34,32 +34,29 @@ impl Quorum {
...
+                (tasks, oks, errors)
             } else {
-                debug!("PUT[{}] ~~~PUT TO {} REMOTE NODES~~~", key, at_least);
-                self.put_remote_nodes(key, data, at_least).await
+                debug!("PUT[{}] ~~~PUT {} REPLICAS TO REMOTE NODES~~~", key, at_least);
+                self.put_remote_nodes(key, data, at_least, &affected_replicas_by_node).await
             };
-        let all_count = self.mapper.get_target_nodes_for_key(key).len();
-        let remote_ok_count = all_count - errors.len() - tasks.len() - local_put_ok;
+        let remote_ok_count = oks.iter().map(|f| affected_replicas_by_node[f.node_name()]).sum::<usize>();
```
ikopylov, line 55: Just noticed, that you have `affected_replicas` in the result data and can use that value instead of access to HashMap

bob/src/cluster/quorum.rs
```diff
@@ -34,32 +34,29 @@ impl Quorum {
...
         failed_nodes.extend(errors.iter().map(|e| e.node_name().clone()));
+        debug!("PUT[{}] LOCAL PUT OK: {}, REMOTE PUT OK: {}, REMOTE PUT ERRORS: {}", 
+               key, local_put_ok, remote_ok_count, 
+               errors.iter().map(|e| affected_replicas_by_node[e.node_name()]).sum::<usize>());
```
ikopylov, line 59: Same here: you can use `affected_replicas` from result for calculation

ikopylov (approved):
LGTM

PR #677, opened 2022-11-28
Ping response

### LINKED ISSUE #625: Reconnect to a node when a ping was received from it
If ping received then it would be good to immediately run reconnection logic if the node is not in connected state

idruzhitskiy:
Closes #625

### DIFF
--- a/bob-common/src/node.rs
+++ b/bob-common/src/node.rs
@@ -6,7 +6,10 @@ use http::Uri;
 use std::{
     fmt::{Debug, Formatter, Result as FmtResult},
     hash::{Hash, Hasher},
-    sync::{Arc, RwLock},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, RwLock,
+    },
 };
 
 pub type Id = u16;
@@ -19,6 +22,7 @@ pub struct Node {
     address: String,
     index: Id,
     conn: Arc<RwLock<Option<BobClient>>>,
+    conn_available: Arc<AtomicBool>,
 }
 
 #[derive(Debug)]
@@ -35,12 +39,13 @@ pub struct Disk {
 }
 
 impl Node {
-    pub async fn new(name: String, address: &str, index: u16) -> Self {
+    pub fn new(name: String, address: &str, index: u16) -> Self {
         Self {
             name,
             address: address.to_string(),
             index,
             conn: Arc::default(),
+            conn_available: Arc::default(),
         }
     }
 
@@ -70,15 +75,23 @@ impl Node {
     }
 
     pub fn set_connection(&self, client: BobClient) {
-        *self.conn.write().expect("rwlock") = Some(client);
+        let mut conn = self.conn.write().expect("rwlock write error");
+        *conn = Some(client);
+        self.conn_available.store(true, Ordering::Release);
     }
 
     pub fn clear_connection(&self) {
-        *self.conn.write().expect("rwlock") = None;
+        let mut conn = self.conn.write().expect("rwlock write error");
+        *conn = None;
+        self.conn_available.store(false, Ordering::Release);
     }
 
     pub fn get_connection(&self) -> Option<BobClient> {
         self.conn.read().expect("rwlock").clone()
+    }
+
+    pub fn connection_available(&self) -> bool {
+        self.conn_available.load(Ordering::Acquire)
     }
 
     pub async fn check(&self, client_factory: &Factory) -> Result<(), String> {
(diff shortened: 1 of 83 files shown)


bob/src/link_manager.rs
```diff
@@ -6,6 +11,7 @@ use termion::color;
 pub(crate) struct LinkManager {
     nodes: Arc<[Node]>,
     check_interval: Duration,
+    node_check_queue: Arc<RwLock<Option<Sender<String>>>>,
```
ikopylov, line 14: Looks like this can be a sync `RwLock`

bob/src/link_manager.rs
```diff
@@ -46,9 +53,40 @@ impl LinkManager {
...
+                    );
+                } else {
+                    debug!("Create connection in response to ping from {}", node.name());
+                }
+            }
+        }
+    }
+
+    pub(crate) async fn spawn_checker(&self, factory: Factory) {
+        let (sender, receiver) = channel(self.nodes.len() * 10);
```
ikopylov, line 78:
```suggestion
        let (sender, receiver) = channel(self.nodes.len() * 2);
```

bob/src/server.rs
```diff
@@ -217,8 +217,13 @@ where
         }
     }
 
-    async fn ping(&self, _: Request<Null>) -> ApiResult<Null> {
+    async fn ping(&self, r: Request<Null>) -> ApiResult<Null> {
         debug!("PING");
+        if let Some(node_name) = r.metadata().get("node_name") {
```
ikopylov, line 222: `node_name` was used for authentication and the behavior has changed. You should add `set_node_name` function to [bob_client.rs](https://github.com/qoollo/bob/blob/master/bob-common/src/bob_client.rs), that will set the name to metadata

bob-common/src/node.rs
```diff
@@ -93,6 +93,10 @@ impl Node {
         }
     }
 
+    pub fn connection_is_set(&self) -> bool {
```
ikopylov, line 96: Should be replaced with `connection_available` when PR https://github.com/qoollo/bob/pull/630 will be merged

bob/src/link_manager.rs
```diff
@@ -85,4 +123,16 @@ impl LinkManager {
         })
         .await
     }
+
+    pub(crate) fn update_node_connection(&self, node_name: &str) {
+        if let Some(node) = self.nodes.iter().find(|n| n.name() == node_name) {
+            if let Some(queue) = self.node_check_queue.read().expect("rwlock").as_ref() {
+                if !node.connection_is_set() {
+                    if let Err(e) = queue.try_send(node_name.to_string()) {
+                        warn!("error while updating node {} status: {}", node_name, e);
```
ikopylov, line 132: Check the error. The queue overflow is not really an error, it's just an indicator that there are too many ping received from disconnected nodes. Change the message accordingly

bob/src/link_manager.rs
```diff
@@ -44,9 +49,42 @@ impl LinkManager {
...
+                        );
+                    } else {
+                        debug!("Create connection in response to ping from {}", node.name());
+                    }
+                }
+            }
+        }
+    }
+
+    pub(crate) async fn spawn_checker(&self, factory: Factory) {
```
ikopylov, line 75: `async` can be removed

ikopylov: And all added `async` all the way up too

bob/src/link_manager.rs
```diff
@@ -78,9 +83,42 @@ impl LinkManager {
...
+        nodes: Arc<[Node]>,
+        factory: Factory,
+        mut node_check_queue: Receiver<String>,
+    ) {
+        while let Some(name) = node_check_queue.recv().await {
+            if let Some(node) = nodes.iter().find(|n| n.name() == name) {
+                if !node.connection_is_set() {
+                    if let Err(e) = node.check(&factory).await {
+                        error!(
+                            "No connection to {}:[{}] - {}",
```
ikopylov, line 96: Need to add information, that this check happen as a result of incoming ping

bob/src/link_manager.rs
```diff
@@ -119,4 +157,19 @@ impl LinkManager {
...
     }
+
+    pub(crate) fn update_node_connection(&self, node_name: &str) {
+        if let Some(node) = self.nodes.iter().find(|n| n.name() == node_name) {
+            if let Some(queue) = self.node_check_queue.read().expect("rwlock").as_ref() {
+                if !node.connection_is_set() {
+                    if let Err(e) = queue.try_send(node_name.to_string()) {
+                        match e {
+                            TrySendError::Full(_) => warn!("too many pings received from node {}", node_name),
+                            TrySendError::Closed(msg) => error!("reconnection channel closed: {}", msg),
```
ikopylov, line 168: Both errors contains original message, so the handling code should be similar

ikopylov (approved):
LGTM

ikopylov (approved):
LGTM

Issue #683, opened 2022-12-05
Data loss during vdisk remount


ikopylov:
If execute vdisk remount while put requests are in progress, then a small number of records are not read after that. No write errors are reported to the client, no aliens created.

Normally, during vdisk remount we should either write records to aliens (preferred) or return write error to client.

PR #692, opened 2022-12-16
Draft: Not clear indexes on remount

ikopylov:
Not clear indexes on remount. This can lead to creation of the holder with the existed timestamp, when `put` runs at the same time the remount procedure happened


ikopylov:
Will be fixed properly as a part of https://github.com/qoollo/bob/issues/683

PR #693, opened 2022-12-17
672 integrationtests run tests with and without authorization

Ziktorin:


.github/workflows/integration-tests.yml
```diff
@@ -44,6 +44,11 @@ on:
         description: 'Upper boundry for chain test'
         required: true
         default: '100000'
+      bob_min_port:
```
ikopylov, line 47: I think there is no need to ask the user to set this parameter, since it does not affect the test in significant way

.github/workflows/integration-tests.yml
```diff
@@ -52,7 +57,7 @@ jobs:
         key: [8, 16]
     env:
       TARGET: "x86_64-unknown-linux-musl"
-      BUILD_MODE: "release"
+      BUILD_MODE: "debug"
```
ikopylov, line 60: Revert this change in this request. This change is already made in https://github.com/qoollo/bob/pull/524

.github/workflows/integration-tests.yml
```diff
@@ -72,7 +77,7 @@ jobs:
         uses: actions-rs/cargo@v1.0.1
         with:
           command: build
-          args: --release --target=${{ env.TARGET }}
+          args: --target=${{ env.TARGET }}
```
ikopylov, line 80: Same here

.github/workflows/integration-tests.yml
```diff
@@ -182,17 +192,18 @@ jobs:
       - name: List availible images
         run: |
           docker image ls
+      #Also sets BOB_BOBP_ACTIONS_OFFSET as GITHUB_ENV from check_interval parameter (check_interval + 1s)
```
ikopylov, line 195: I added comment for this in https://github.com/qoollo/bob/pull/524. In this request that line should be removed

.github/workflows/integration-tests.yml
```diff
@@ -201,6 +212,7 @@ jobs:
     strategy:
       matrix:
         key: [8, 16]
+        auth: [None, Basic]
```
ikopylov, line 215: There is no need to run aliens test with auth. gRPC auth was already tested within put-get-exist test

ikopylov: But you need to add auth for `bobt` test, because `bobt` works through REST API

CHANGELOG.md
```diff
@@ -4,13 +4,24 @@ Bob versions changelog
...
+
+## [2.1.0-alpha.7] - 2022-11-28
+#### Added
 - Include bobt into the zip archive attached to the release infrastructure (#669)
 - Include bobc into release builds (#569)
 - Added blob-info and index-info features to brt (#356)
 - Support for files, file name patterns, key ranges and a 'exists' subcommand to 'bobc' (#539)
-- Integration tests (#518)
-- Added integration tests for aliens (#642)
-- Added integration tests for bobt (#648)
```
ikopylov, line 13: Something bad happened here. Revert all this changes and add normal description of this particular PR

integration-tests/generate_from_templates.py
```diff
@@ -101,5 +103,5 @@ def pathified(string):
 with open("Templates/users_template.yml.j2") as users:
     template = Template(users.read())
     with open(os.path.join(path, 'users.bobnet.yaml'), 'w') as f:
-        f.write(template.render())
+        f.write(template.render(auth_user=args.auth_user, auth_password=args.auth_password))
```
ikopylov, line 106: You should also enable auth in `node.yaml`

Ziktorin: Auth is enabled by passing `vars(args)` into `template.render()` for node config. It is the same as explicitly listing every parameter that was parsed by parse_args() function, and as long as there is a parameter with the same name in a template it will be automatically substituded with an according value. And among arguments there is an `auth_type` argument that sets the auth type depending on value in the matrix.

.github/workflows/integration-tests.yml
```diff
@@ -143,6 +143,7 @@ jobs:
       matrix:
         key: [8, 16]
         mode: [normal, random]
+        auth: [None, Basic]
```
ikopylov, line 146: With this line we will run 8 tests in parallel. But it is enough to run a single test with auth enabled.
Please, use `include` to add one specific configuration: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations

This line will be `auth: [None]` and include should add `(key: 8, mode: normal, auth: Basic)`

.github/workflows/integration-tests.yml
```diff
@@ -283,6 +289,7 @@ jobs:
     strategy:
       matrix:
         key: [8, 16]
+        auth: [None, Basic]
```
ikopylov, line 292: Same here. Run test with `Basic` auth only on 8 bytes keys

ikopylov:
Merge master to this branch and redirect request to master

.github/workflows/integration-tests.yml
```diff
@@ -32,15 +33,15 @@ on:
         required: true
         default: '0'
       bob_bobt_count:
-        description: 'Amount of iterations for bobt test'
+        description: 'Amount of iterations for chain operation test'
         required: true
         default: '100000'
       bob_bobt_first_index:
-        description: 'Bottom boundary for key in bobt test'
+        description: 'Bottom boundry for chain test'
```
ikopylov, line 40: And this

.github/workflows/integration-tests.yml
```diff
@@ -32,15 +33,15 @@ on:
...
         required: true
         default: '0'
       bob_bobt_last_index:
-        description: 'Upper boundary for key in bobt test'
+        description: 'Upper boundry for chain test'
```
ikopylov, line 44: And this one too

.github/workflows/integration-tests.yml
```diff
@@ -229,7 +239,7 @@ jobs:
           INPUT="${{ github.event.inputs.bob_replicas_amount }}"
           echo "BOB_REPLICAS_AMOUNT=${INPUT:-"2"}" >> $GITHUB_ENV
           INPUT="${{ github.event.inputs.bob_payload_size }}"
-          echo "BOB_PAYLOAD_SIZE=${INPUT:-"4096"}" >> $GITHUB_ENV
+          echo "BOB_PAYLOAD_SIZE=${INPUT:-"10240"}" >> $GITHUB_ENV
```
ikopylov, line 242: Revert this change

.github/workflows/integration-tests.yml
```diff
@@ -263,6 +273,7 @@ jobs:
       - name: List availible images
         run: |
           docker image ls
+      #Also sets BOB_BOBP_ACTIONS_OFFSET as GITHUB_ENV from check_interval parameter (check_interval + 1s)
```
ikopylov, line 276: Revert this change

.github/workflows/integration-tests.yml
```diff
@@ -301,11 +316,15 @@ jobs:
           INPUT="${{ github.event.inputs.bob_bobt_first_index }}"
           echo "BOB_BOBT_FIRST_INDEX=${INPUT:-"0"}" >> $GITHUB_ENV
           INPUT="${{ github.event.inputs.bob_bobt_last_index }}"
-          echo "BOB_BOBT_LAST_INDEX=${INPUT:-"10000"}" >> $GITHUB_ENV
+          echo "BOB_BOBT_LAST_INDEX=${INPUT:-"100000"}" >> $GITHUB_ENV
```
ikopylov, line 319: Revert this change

.github/workflows/integration-tests.yml
```diff
@@ -333,15 +352,16 @@ jobs:
       - name: List availible images
         run: |
           docker image ls
+      #Also sets BOB_BOBP_ACTIONS_OFFSET as GITHUB_ENV from check_interval parameter (check_interval + 1s)
```
ikopylov, line 355: This one too

.github/workflows/integration-tests.yml
```diff
@@ -32,15 +33,15 @@ on:
         required: true
         default: '0'
       bob_bobt_count:
-        description: 'Amount of iterations for bobt test'
+        description: 'Amount of iterations for bobt operation test'
         required: true
         default: '100000'
       bob_bobt_first_index:
-        description: 'Bottom boundary for key in bobt test'
+        description: 'Bottom boundry for key in bobt test'
```
ikopylov, line 40:
```suggestion
        description: 'Bottom boundary for key in bobt test'
```

.github/workflows/integration-tests.yml
```diff
@@ -32,15 +33,15 @@ on:
...
         required: true
         default: '0'
       bob_bobt_last_index:
-        description: 'Upper boundary for key in bobt test'
+        description: 'Upper boundry for key in bobt test'
```
ikopylov, line 44:
```suggestion
        description: 'Upper boundary for key in bobt test'
```

.github/workflows/integration-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: integration-tests
 on:
   push:
+    branches: ['672-integrationtests-run-tests-with-and-without-authorization'] #REMOVE WHEN ON MASTER
```
ikopylov, line 4: Remove this line

ikopylov (approved):
LGTM

PR #696, opened 2022-12-20
683 data loss during remount

### LINKED ISSUE #683: Data loss during vdisk remount

### LINKED ISSUE #701: Add active Holder to holder collection in `Group::run`
~In the current implementation `self.create_current_pearl()` is called, but its result is not added to `holders` vector:~
https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/group.rs#L82

Perhaps, it is better to remove this call from `run`. Holder will be created on the first put request

Boneyan:
Data access operations will be locked during remount
Closes #683
Closes #701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+- Data access operations will be locked during remount (#683)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@ Bob versions changelog
 #### Changed
 - Log message about the lack of connection to graphite became more understandable (#684)
 - Hostname resolving in background tasks (#529)
-- Removed holder creation from post-processing (#701)
+- Removed broken logic of holder creation from group::run function (#701)
 
 #### Fixed
-- Data access operations will be locked during remount (#683)
+- Data access operations will be protected during remount to prevent data loss (#683)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Bob versions changelog
 #### Changed
 - Log message about the lack of connection to graphite became more understandable (#684)
 - Hostname resolving in background tasks (#529)
+- Removed holder creation from post-processing (#701)
 
 #### Fixed
 - Data access operations will be locked during remount (#683)
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0"
-async-std = "1.12.0"
+async-std = "1.12"
 async-trait = "0.1"
 bitflags = "1.3"
 bob-common = { path = "../bob-common", features = [] }
--- a/bob-backend/Cargo.toml
+++ b/bob-backend/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0"
+async-std = "1.12.0"
 async-trait = "0.1"
 bitflags = "1.3"
 bob-common = { path = "../bob-common", features = [] }
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -76,12 +76,6 @@ impl Group {
             .await
             .with_context(|| "backend pearl group read vdisk directory failed")?;
         debug!("{}: count holders: {}", self, new_holders.len());
-        if !new_holders
-            .iter()
-            .any(|holder| holder.gets_into_interval(get_current_timestamp()))
-        {
-            self.create_current_pearl();
-        }
         debug!("{}: save holders to group", self);
         let mut holders = self.holders.write().await;
         holders.extend(new_holders).await;   
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -336,11 +336,12 @@ impl Group {
         timestamp_config: StartTimestampConfig,
         force_delete: bool,
     ) -> Result<u64, Error> {
+        let _reinit_lock = self.reinit_lock.try_read().map_err(|_| Error::holder_temporary_unavailable())?;
         let mut reference_timestamp = meta.timestamp();
         let mut total_deletion_count = 0;
 
         if force_delete {
-            let actual_holder = self.get_actual_holder(get_current_timestamp(), timestamp_config).await?;
+            let actual_holder = self.get_or_create_actual_holder(get_current_timestamp(), timestamp_config).await?;
             reference_timestamp = actual_holder.1.start_timestamp();
             total_deletion_count += self.delete_in_actual_holder(actual_holder, key, meta).await?;
         }
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -309,7 +309,7 @@ impl Group {
         result
     }
 
-    pub async fn exist(&self, keys: &[BobKey]) -> Vec<bool> {
+    pub async fn exist(&self, keys: &[BobKey]) -> Result<Vec<bool>, Error> {
         let _reinit_lock = self.reinit_lock.try_read().map_err(|_| Error::holder_temporary_unavailable())?;
         let mut exist = vec![false; keys.len()];
         let holders = self.holders.read().await;
@@ -321,7 +321,7 @@ impl Group {
                 }
             }
         }
-        exist
+        Ok(exist)
     }
 
     pub fn holders(&self) -> Arc<RwLock<HoldersContainer>> {
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -8,13 +8,14 @@ use crate::{
 use futures::Future;
 use pearl::BloomProvider;
 use ring::digest::{digest, SHA256};
+use async_std::sync::{RwLock as UgradableRwLock, RwLockUpgradableReadGuard};
 
 pub type HoldersContainer =
     HierarchicalFilters<Key, <Holder as BloomProvider<Key>>::Filter, Holder>;
 
 #[derive(Clone, Debug)]
 pub struct Group {
-    holders: Arc<RwLock<HoldersContainer>>,
+    holders: Arc<UgradableRwLock<HoldersContainer>>,
     reinit_lock: Arc<RwLock<()>>,
     settings: Arc<Settings>,
     directory_path: PathBuf,
@@ -22,7 +23,6 @@ pub struct Group {
     node_name: String,
     disk_name: String,
     owner_node_name: String,
-    created_holder_indexes: Arc<RwLock<HashMap<u64, ChildId>>>,
     dump_sem: Arc<Semaphore>,
 }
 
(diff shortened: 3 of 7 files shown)


bob-common/src/error.rs
```diff
@@ -196,4 +202,5 @@ pub enum Kind {
     RequestFailedCompletely(String),
     DisksEventsLogger(String),
     Unauthorized,
+    HolderTemporaryUnavailable,
```
ikopylov, line 205: Please, check all places, where Error conversion happened. Example: https://github.com/qoollo/bob/blob/master/bob/src/api/mod.rs#L1129
This new error should be converted properly

bob-backend/src/pearl/group.rs
```diff
@@ -88,6 +90,7 @@ impl Group {
     }
 
     pub async fn remount(&self, pp: impl Hooks) -> AnyResult<()> {
+        let _reinit_lock = self.reinit_lock.write().await;
```
ikopylov, line 93: `run` should also acquire `reinit_lock`

bob-backend/src/pearl/group.rs
```diff
@@ -261,6 +265,7 @@ impl Group {
             .iter_possible_childs_rev(&Key::from(key))
             .map(|(_, x)| &x.data)
         {
+            let _reinit_lock = self.reinit_lock.try_read().map_err(|_| Error::holder_temporary_unavailable())?;
```
ikopylov, line 268: This should be at the beginning of the function (you should protect `holders` iteration as well)

bob-backend/src/pearl/group.rs
```diff
@@ -15,6 +15,7 @@ pub type HoldersContainer =
 #[derive(Clone, Debug)]
 pub struct Group {
     holders: Arc<RwLock<HoldersContainer>>,
+    reinit_lock: Arc<RwLock<()>>,
```
ikopylov, line 18: I think we can add this lock directly to the `holder`. But to do this, you need to make sure that the functions called inside do not try to re-acquire this lock

bob-backend/src/pearl/group.rs
Boneyan: As I understand, reinit_lock is dropped at the end of the function, so it is still acquired while run() is running. Or is it better to drop reinit_lock before run() and reacquire it inside run()?

ikopylov: You should refactor `run` function. You can move all the code from `run` function into the new private function (you can name it `run_under_lock` or `run_inner`) and call that function from `remount`. The original `run` function in this case will acquire `reinit_lock` and then call the new private function

bob-backend/src/pearl/group.rs
```diff
@@ -63,34 +64,42 @@ impl Group {
         }
     }
 
-    pub async fn run(&self, pp: impl Hooks) -> AnyResult<()> {
+    async fn run_under_lock(&self, pp: impl Hooks) -> AnyResult<()> {
```
ikopylov, line 67: `run_under_lock` -> `run_under_reinit_lock`

bob-backend/src/pearl/group.rs
```diff
@@ -136,17 +135,20 @@ impl Group {
...
+            self.create_write_pearl(data, timestamp_config).await
+        } else {
+            holder_info
+        }
     }
 
     // find in all pearls actual pearl
-    async fn find_actual_holder(&self, data: &BobData) -> BackendResult<(ChildId, Holder)> {
-        let holders = self.holders.read().await;
+    async fn find_actual_holder(holders: &HoldersContainer, data: &BobData) -> BackendResult<(ChildId, Holder)> {
```
ikopylov, line 151: Pass `data_timestamp` instead of `data` to this function

bob-backend/src/pearl/group.rs
```diff
@@ -170,54 +172,37 @@ impl Group {
     // create pearl for current write
     async fn create_write_pearl(
         &self,
-        ts: u64,
+        data: &BobData,
```
ikopylov, line 175: It is better to pass `data_timestamp` instead of `data`

bob-backend/src/pearl/group.rs
```diff
@@ -170,54 +172,37 @@ impl Group {
...
         timestamp_config: StartTimestampConfig,
     ) -> Result<(ChildId, Holder), Error> {
-        let mut indexes = self.created_holder_indexes.write().await;
-        let created_holder_index = indexes.get(&ts).copied();
-        let index = if let Some(exisiting_index) = created_holder_index {
-            exisiting_index
+        let holders = self.holders.upgradable_read().await;
```
ikopylov, line 178: Please add a comment that only one thread can hold the `upgradable_read` lock at a time. This is crucial for proper working

bob-backend/src/pearl/group.rs
```diff
@@ -170,54 +172,37 @@ impl Group {
...
         } else {
-            let new_index = self
+            info!("creating pearl for timestamp {}", data.meta().timestamp());
+            let pearl = self.create_pearl_by_timestamp(data.meta().timestamp(), &timestamp_config);
+            self
                 .settings
                 .config()
                 .try_multiple_times_async(
-                    || self.try_create_write_pearl(ts, &timestamp_config),
+                    || pearl.prepare_storage(),
```
ikopylov, line 190: You should preserve old behavior: run first `create_pearl_by_timestamp` and the `prepare_storage`. 
It is very likely that in the case of a error, the `holder` needs to be recreated.

bob-backend/src/pearl/group.rs
```diff
@@ -170,54 +172,37 @@ impl Group {
...
-            index,
-            self.holders
-                .read()
-                .await
-                .get_child(index)
+            debug!("backend pearl group save pearl storage prepared");    
+            let mut holders = RwLockUpgradableReadGuard::upgrade(holders).await;
+            let new_index = holders.push(pearl).await;
+            debug!("group create write pearl holder inserted, index {:?}", new_index);
+            let holder = holders
```
ikopylov, line 199: You've just added `pearl` and got its index. There is no reason to take it from the collection. You can just return `(new_index, pearl)`

bob-backend/src/pearl/group.rs
```diff
@@ -136,17 +135,20 @@ impl Group {
         data: &BobData,
         timestamp_config: StartTimestampConfig,
     ) -> Result<(ChildId, Holder), Error> {
```
ikopylov, line 137: It is better to name this function `get_or_create_actual_holder`

bob-backend/src/pearl/group.rs
```diff
@@ -170,54 +172,37 @@ impl Group {
     // create pearl for current write
     async fn create_write_pearl(
```
ikopylov, line 173: I think now it is better to name this function `get_or_create_write_pearl`

bob-backend/src/pearl/group.rs
```diff
@@ -335,7 +323,7 @@ impl Group {
     }
 
     pub async fn attach(&self, start_timestamp: u64) -> BackendResult<()> {
-        let holders = self.holders.read().await;
+        let holders = self.holders.upgradable_read().await;
```
ikopylov, line 326: Same as above: Please add a comment that only one thread can hold the `upgradable_read` lock at a time. This is crucial for proper working

ikopylov: Do I understand correctly that there was a bug here before?

ikopylov:
Please, backport part with `reinit_lock` to `release_v2.0.0`

bob-backend/src/pearl/group.rs
Boneyan: Yeah, as I understand there was attempt to get write lock inside save_pearl() (inside add()) before read lock was dropped

bob-backend/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0"
+async-std = "1.12.0"
```
ikopylov, line 12:
```suggestion
async-std = "1.12"
```

bob-backend/src/pearl/group.rs
```diff
@@ -63,34 +64,42 @@ impl Group {
...
             self.create_current_pearl();
         }
         debug!("{}: save holders to group", self);
-        self.add_range(holders).await;
+        {
+            let mut holders = self.holders.write().await;
+            holders.extend(new_holders).await;   
+        }
         debug!("{}: start holders", self);
         self.run_pearls(pp).await
```
ikopylov, line 91: This is the only place, where `run_pearls` is called. I suggest to change `run_pearls` to not acquire the lock and pass `holders` as parameter. In this way the write lock taken here can be acquired up until the end of the function

bob-backend/src/pearl/group.rs
```diff
@@ -63,39 +64,43 @@ impl Group {
...
             .await
             .with_context(|| "backend pearl group read vdisk directory failed")?;
-        debug!("{}: count holders: {}", self, holders.len());
-        if !holders
+        debug!("{}: count holders: {}", self, new_holders.len());
+        if !new_holders
             .iter()
             .any(|holder| holder.gets_into_interval(get_current_timestamp()))
         {
             self.create_current_pearl();
```
ikopylov, line 83: Can you also fix https://github.com/qoollo/bob/issues/701 as a part of this PR?

bob-backend/src/pearl/group.rs
```diff
@@ -63,39 +64,43 @@ impl Group {
...
                 || self.read_vdisk_directory(),
                 "can't create pearl holders",
                 self.settings.config().fail_retry_timeout(),
             )
             .await
             .with_context(|| "backend pearl group read vdisk directory failed")?;
-        debug!("{}: count holders: {}", self, holders.len());
-        if !holders
+        debug!("{}: count holders: {}", self, new_holders.len());
+        if !new_holders
```
ikopylov, line 79: Looks like `find_actual_holder` can be used here

bob-backend/src/pearl/group.rs
ikopylov: Actually, I think it is better to delete this call at all. Actual holder will be created on the first put request

bob-backend/src/pearl/group.rs
```diff
@@ -63,39 +64,37 @@ impl Group {
...
-            .iter()
-            .any(|holder| holder.gets_into_interval(get_current_timestamp()))
-        {
-            self.create_current_pearl();
-        }
+        debug!("{}: count holders: {}", self, new_holders.len());
         debug!("{}: save holders to group", self);
-        self.add_range(holders).await;
+        let mut holders = self.holders.write().await;
+        holders.extend(new_holders).await;
```
ikopylov, line 81: In fact, I don't see any normal scenario where we need to keep old items in the `holders` collection. It might be good to clear `holders` right before calling `extend`

CHANGELOG.md
```diff
@@ -10,9 +10,10 @@ Bob versions changelog
 #### Changed
 - Log message about the lack of connection to graphite became more understandable (#684)
 - Hostname resolving in background tasks (#529)
+- Removed holder creation from post-processing (#701)
```
ikopylov, line 13:
```suggestion
- Removed broken logic of holder creation from group::run function (#701)
```

CHANGELOG.md
```diff
@@ -10,9 +10,10 @@ Bob versions changelog
...
 
 #### Fixed
-
+- Data access operations will be locked during remount (#683)
```
ikopylov, line 16:
```suggestion
- Data access operations will be protected during remount to prevent data loss (#683)
```

ikopylov (approved):
LGTM

ikopylov (approved):
LGTM

PR #699, opened 2022-12-24
683 data loss during remount backport

Boneyan:
Data access operations will be locked during remount

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+- Data access operations will be locked during remount (#683)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-- Data access operations will be locked during remount (#683)
+- Data access operations will be protected during remount to prevent data loss (#683)
 
 #### Updated
 
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -313,8 +313,8 @@ impl Group {
     }
 
     pub async fn exist(&self, keys: &[BobKey]) -> Result<Vec<bool>, Error> {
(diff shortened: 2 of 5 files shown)


bob-backend/src/pearl/group.rs
```diff
@@ -302,8 +312,9 @@ impl Group {
         result
     }
 
-    pub async fn exist(&self, keys: &[BobKey]) -> Vec<bool> {
+    pub async fn exist(&self, keys: &[BobKey]) -> Result<Vec<bool>, Error> {
         let mut exist = vec![false; keys.len()];
+        let _reinit_lock = self.reinit_lock.try_read().map_err(|_| Error::holder_temporary_unavailable())?;
```
ikopylov, line 317: Please, move this line to the very beginning of this function. In this way it will be unified with the other functions. Also `try_read` is non-blocking, so there is no point in moving any operation to a position before it

bob-backend/src/pearl/group.rs
```diff
@@ -333,11 +344,14 @@ impl Group {
     }
 
     pub async fn attach(&self, start_timestamp: u64) -> BackendResult<()> {
-        let holders = self.holders.read().await;
-        if holders
-            .iter()
-            .map(|x| x.start_timestamp())
-            .any(|timestamp| timestamp == start_timestamp)
+        if {
+            let holders = self.holders.read().await;
```
ikopylov, line 348: `attach` is a rare operation, so I think it is better to acquire `write` lock at the beginning of the function and held it until the end. `save_pearl` can be manually inlined here to avoid deadlock

ikopylov (approved):
LGTM

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Bob versions changelog
 
 
 #### Fixed
-
+- Data access operations will be locked during remount (#683)
```
ikopylov, line 13:
```suggestion
- Data access operations will be protected during remount to prevent data loss (#683)
```

PR #731, opened 2023-01-12
690 checsum mismatch error

### LINKED ISSUE #690: Checksum mismatch error for large record sizes
Reproduces on cluster with 4 nodes, 2 replicas. Record size: 512Mb. Writes with `bobp` with 32 threads.
Timeout occurs during the write (`operation_timeout: 60s`), so this may somehow affect the manifestation of the bug. Normally, timeout should not lead to checksum mismatch error.

```
2022-12-14 00:52:10         pearl::record:151 ERROR Error {
    kind: Validation {
        kind: RecordDataChecksum,
        cause: "wrong data checksum 392551754 vs 3261716721",
    },
}
```

Boneyan:
Fixed data corruption in case of put request timeout
Closes #690

### DIFF
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -166,9 +166,13 @@ impl BackendStorage for Pearl {
             .iter()
             .find(|dc| dc.can_process_operation(&op));
         if let Some(disk_controller) = dc_option {
-            disk_controller
-                .put(op, key, data)
-                .await
+            let handler = tokio::spawn(
+                disk_controller.clone().put(op, key, data.clone())
+            );
+            match handler.await {
+                Ok(res) => res,
+                Err(e) => Err(Error::failed(e.to_string()))
+            }
         } else {
             debug!(
                 "PUT[{}] Cannot find disk_controller, operation: {:?}",
@@ -234,7 +238,13 @@ impl BackendStorage for Pearl {
             .find(|dc| dc.can_process_operation(&op));
 
         if let Some(disk_controller) = dc_option {
-            disk_controller.delete(op, key, meta).await
+            let handler = tokio::spawn(
+                disk_controller.clone().delete(op, key, meta.clone())
+            );
+            match handler.await {
+                Ok(res) => res,
+                Err(e) => Err(Error::failed(e.to_string()))
+            }
         } else {
             Err(Error::dc_is_not_available())
         }
--- a/bob-backend/src/pearl/core.rs
+++ b/bob-backend/src/pearl/core.rs
@@ -142,6 +142,17 @@ impl MetricsProducer for Pearl {
     }
 }
 
+async fn spawn_call<T, R>(future: T) -> BackendResult<R>
+where 
+T: Future<Output=BackendResult<R>> + std::marker::Send + 'static,
+R: std::marker::Send + 'static,
+{
+    match tokio::spawn(future).await {
+        Ok(res) => res,
+        Err(e) => Err(Error::failed(e.to_string()))
+    }
+}
+
 #[async_trait]
 impl BackendStorage for Pearl {
     async fn run_backend(&self) -> AnyResult<()> {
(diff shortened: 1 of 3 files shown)


bob-backend/src/pearl/core.rs
```diff
@@ -166,9 +166,13 @@ impl BackendStorage for Pearl {
             .iter()
             .find(|dc| dc.can_process_operation(&op));
         if let Some(disk_controller) = dc_option {
-            disk_controller
-                .put(op, key, data)
-                .await
+            let handler = tokio::spawn(
+                disk_controller.clone().put(op, key, data.clone())
```
ikopylov, line 170: `put_alien` below should also be updated

bob-backend/src/pearl/core.rs
```diff
@@ -234,7 +238,13 @@ impl BackendStorage for Pearl {
             .find(|dc| dc.can_process_operation(&op));
 
         if let Some(disk_controller) = dc_option {
-            disk_controller.delete(op, key, meta).await
+            let handler = tokio::spawn(
+                disk_controller.clone().delete(op, key, meta.clone())
```
ikopylov, line 242: `delete_alien` below should also be updated

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -400,10 +400,10 @@ impl DiskController {
     }
 
     pub(crate) async fn put(
-        &self,
+        self: Arc<Self>,
         op: Operation,
         key: BobKey,
-        data: &BobData,
+        data: BobData,
```
ikopylov, line 406: You can still pass `&BobData`. And I think this is better in general

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -400,10 +400,10 @@ impl DiskController {
     }
 
     pub(crate) async fn put(
-        &self,
+        self: Arc<Self>,
```
ikopylov, line 403: Why this is needed? `&self` would be better

bob-backend/src/pearl/core.rs
ikopylov, line 170: Probably, you should use async block to avoid requirements of changing signature of functions in disk_contoller:
```
tokio::spawn(async move {
    disk_controller.clone().put(op, key, data.clone()).await
});

File: bob-backend/src/pearl/disk_controller.rs (hunk 4 above)
Comment (Boneyan) [reply bob-backend/src/pearl/disk_controller.rs:403]: This function is called inside spawn(), so it is not possible to pass simple reference to another thread

File: bob-backend/src/pearl/disk_controller.rs (hunk 3 above)
Comment (Boneyan) [reply bob-backend/src/pearl/disk_controller.rs:406]: Same situation as with self, it is not possible to pass simple reference to another thread (in case of timeout this reference will expire)

Comment (Boneyan) [reply bob-backend/src/pearl/disk_controller.rs:406]: Ok, it is possible in case of async block, so I will use it

File: bob-backend/src/pearl/disk_controller.rs (hunk 4 above)
Comment (Boneyan) [reply bob-backend/src/pearl/disk_controller.rs:403]: Ok, I will use async block to do so

File: bob-backend/src/pearl/core.rs [hunk 5]
```diff
@@ -165,10 +176,11 @@ impl BackendStorage for Pearl {
             .disk_controllers
             .iter()
             .find(|dc| dc.can_process_operation(&op));
-        if let Some(disk_controller) = dc_option {
-            disk_controller
-                .put(op, key, data)
-                .await
+        if let Some(disk_controller) = dc_option.cloned() {
+            let data = data.clone();
+            spawn_call(async move {
```
ikopylov, line 181: Reminder for the future (no need to do it right now):
to avoid DDOS by writing large records with short timeout, it would be good to limit the number of spawning requests with the help of Semaphore

ikopylov:
Was fixed on the pearl side with the PR https://github.com/qoollo/pearl/pull/250 
Right now current PR is not needed, but in future it can be reopened

PR #732, opened 2023-01-12
700 order holders by timestamp

### LINKED ISSUE #700: Order holders inside group by their start timestamp
`read_vdisk_directory` should return vector of `Holders` ordered by `start_timestamp`. This is required for common correctness

https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/group.rs#L427

Boneyan:
Holders returned by read_vdisk_directory() are now ordered by start_timestamp
Closes #700

### DIFF
--- a/bob-backend/src/pearl/group.rs
+++ b/bob-backend/src/pearl/group.rs
@@ -534,7 +534,9 @@ impl Group {
             {
                 if let Some(name) = PartitionName::try_from_string(&file_name) {
                     let pearl_holder = self.create_pearl_holder(name.timestamp, &name.hash);
-                    holders.push(pearl_holder);
+                    let pos = holders.binary_search_by(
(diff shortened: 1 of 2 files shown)


bob-backend/src/pearl/group.rs
```diff
@@ -534,7 +534,9 @@ impl Group {
             {
                 if let Some(name) = PartitionName::try_from_string(&file_name) {
                     let pearl_holder = self.create_pearl_holder(name.timestamp, &name.hash);
-                    holders.push(pearl_holder);
+                    let pos = holders.binary_search_by(
+                        |holder: &Holder| holder.start_timestamp().cmp(&name.timestamp));
```
ikopylov, line 538: For safety, use `holder.start_timestamp()` instead of name.timestamp.
Also, `sort_by` at the end would be simpler (just 1 line)

ikopylov (approved):
LGTM

PR #734, opened 2023-01-12
removed outdated parameters from node.yaml

### LINKED ISSUE #703: Remove `open_blobs_soft_limit` and `open_blobs_hard_limit` from node config in config-examples
We now have `index_memory_limit` parameter which is preferable to use

Ziktorin:
closes #703


ikopylov (approved):
LGTM

PR #740, opened 2023-01-13
702 get exist optimisation

### LINKED ISSUE #702: Implement an optimization for `get` and `exist` to skip holders by their timestamp
The current implementation reads data from all holders, which can be slow when there are many records for the desired key. We can implement an optimization that skips holders whose maximum record timestamp in that holder is less than the record timestamp already read. The proper implementation requires tracking max and min record timestamp within holder and tracked by issue #694. But we can implement it faster in simplified and unsafe way.

The problem is that now the holder's `end_timestamp` is calculated as `start_timestamp + self.settings.timestamp_period_as_secs()`. If suddenly the value of `timestamp_period` in the config changes, the calculation will be incorrect.

With proposing unsafe optimization, we can try to estimate the maximum period of creation of existed holders, multiply it by some safe coefficient, and also adjust it by the value of the new config parameter. 

First, we should introduce new config parameter `skip_holders_by_timestamp_step_when_reading`. By default it should be `None` (means that optimization is turned off), but it can be configured with the human readable time period (e.g. `1d`).

Procedure of calculating of safe timestamp step:
1. If `skip_holders_by_timestamp_step_when_reading` is `None` then return `None`
2. Find maximum period from existed holders:
  2.1. Copy `start_timestamp` from `holders` collection to a vector
  2.2. Order vector by ascending
  2.3. Add current timestamp to an end
  2.4. Find maximum difference between 2 timestamps in that vector
3. Calculate safe step by formula `max(settings.skip_holders_by_timestamp_step_when_reading, max(2 * max_period_from_existed_holders, 5 * settings.timestamp_period)) + 1h`

Then within `get` and `exist` functions we can implement smart skipping by compare timestamp of already read record and `end_timestamp` of current holder with addition of safe timestamp step (https://github.com/qoollo/bob/blob/master/bob-backend/src/pearl/group.rs#L264):
``` 
if max_timestamp == None || safe_timestamp_step == None || holder.end_timestamp() + safe_timestamp_step.unwrap() >= max_timestamp {
  // Process this holder
}
```

This issue should be done after https://github.com/qoollo/bob/pull/621 will be merged

Boneyan:
Added optional get & exist optimization
Closes #702


bob-backend/src/pearl/group.rs
```diff
@@ -79,14 +81,16 @@ impl Group {
...
         debug!("{}: start holders", self);
         Self::run_pearls(&mut holders, pp).await
     }
 
-    pub async fn run(&self, pp: impl Hooks) -> AnyResult<()> {
+    pub async fn run(&mut self, pp: impl Hooks) -> AnyResult<()> {
         let _reinit_lock = self.reinit_lock.write().await;
-        self.run_under_reinit_lock(pp).await
+        let res = self.run_under_reinit_lock(pp).await;
+        self.safe_timestamp_step = self.safe_timestamp_step().await;
```
ikopylov, line 92: It is better to do this in `run_under_reinit_lock` to cover both `run` and `remount`

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Bob versions changelog
 - Added integration tests for aliens (#642)
 - Added integration tests for bobt (#648)
 - Added authentification support for integration tests (#672)
+- Added optional get & exist optimization (#702)
```
ikopylov, line 17: Move this line to `[Unreleased]` section

config-examples/node.yaml
```diff
@@ -61,6 +61,8 @@ bloom_filter_memory_limit: 8 GiB
 authentication_type: Basic
 # [size] memory limit for all indexes. Unlimited if not specified
 index_memory_limit: 8 GiB
+# [time] optional, threshold for get & exist holder reading optimisation
+skip_holders_by_timestamp_step_when_reading: 1h
```
ikopylov, line 65: By default it should be None. Is it possible to write `None` here? If not, then it is better to remove this parameter from config-examples

config-examples/node.yaml
```diff
@@ -61,6 +61,8 @@ bloom_filter_memory_limit: 8 GiB
 authentication_type: Basic
 # [size] memory limit for all indexes. Unlimited if not specified
 index_memory_limit: 8 GiB
+# [time] optional, threshold for get & exist holder reading optimisation
```
ikopylov, line 64:
```suggestion
# [time] optional, threshold for unsafe get & exist holder reading optimization
```

bob-backend/src/pearl/group.rs
```diff
@@ -233,6 +239,34 @@ impl Group {
...
+            let mut start_timestamps = self.holders.read().await
+                .iter()
+                .map(|holder| holder.start_timestamp())
+                .collect::<Vec<u64>>();
+            start_timestamps.sort();
+            start_timestamps.push(get_current_timestamp());
+            let mut max_diff = None;
+            for i in 0..start_timestamps.len() - 1 {
+                let diff = start_timestamps[i + 1] - start_timestamps[i];
+                if let Some(max_diff_v) = max_diff {
```
ikopylov, line 253: Maybe use reference (`let Some(max_diff_v) = max_diff.as_mut()`). In this way you can later set value by `*max_diff_v = diff`

bob-backend/src/pearl/group.rs
```diff
@@ -233,6 +239,34 @@ impl Group {
...
+                    if max_diff_v < diff {
+                        max_diff = Some(diff);
+                    }
+                } else {
+                    max_diff = Some(diff);
+                }
+            }
+            if let Some(max_diff) = max_diff {
```
ikopylov, line 261: If `max_diff` is None, then init it to `self.settings.timestamp_period_as_secs()`:
```
let max_diff = max_diff.unwrap_or(self.settings.timestamp_period_as_secs());
```

bob-backend/src/pearl/group.rs
```diff
@@ -302,32 +338,41 @@ impl Group {
...
-                    ReadResult::NotFound => continue,
                 }
             }
             exist[ind] = result;
         }
         Ok(exist)
     }
 
+    #[inline]
+    fn should_check_holder(&self, holder: &Holder, max_timestamp: Option<u64>) -> bool {
```
ikopylov, line 371: Please, make this function static: don't pass `self` as parameter, pass `self.safe_timestamp_step` instead

CHANGELOG.md
Boneyan: It is already there

config-examples/node.yaml
Boneyan: It is not possible

CHANGELOG.md
Boneyan: Moved after to [Unreleased] after resolving conflict

ikopylov (approved):
LGTM

bob-backend/src/pearl/group.rs
```diff
@@ -306,35 +339,44 @@ impl Group {
...
                 }
             }
             exist[ind] = result;
         }
         Ok(exist)
     }
 
-    
+    #[inline]
+    fn should_check_holder(safe_timestamp_step: Option<u64>, holder: &Holder, max_timestamp: Option<u64>) -> bool {
```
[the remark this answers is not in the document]
ikopylov: There doesn't seem to be much difference. I changed it according to your suggestion

idruzhitskiy (approved):
LGTM

PR #746, opened 2023-01-14
708 attach timestamp to deletion record and use it to find newest records

### LINKED ISSUE #708: Attach the user's timestamp to the deletion records and use it to find the newest record in get and exist functions
We should look for the newest entries only by comparing the user's timestamps.
Here is the incorrect Pearl level timestamp used and they should be replaced with user level timestamps:
https://github.com/qoollo/bob/blob/54498e8797f17030c33d983001671607ffdb6642/bob-backend/src/pearl/group.rs#L277
https://github.com/qoollo/bob/blob/54498e8797f17030c33d983001671607ffdb6642/bob-backend/src/pearl/group.rs#L328
https://github.com/qoollo/bob/blob/54498e8797f17030c33d983001671607ffdb6642/bob-backend/src/pearl/group.rs#L335

ikopylov:
Closes #708


Cargo.toml
```diff
@@ -85,7 +85,7 @@ tokio = { version = "1.28", features = [] }
 
 # pearl
 [workspace.dependencies.pearl]
-version = "0.19.0"
+version = "0.20.0"
```
ikopylov, line 88: This is temporary change. https://github.com/qoollo/bob/pull/848 should be merge first

PR #765, opened 2023-01-23
added doubled exist test

Ziktorin:


integration-tests/tests.py
```diff
@@ -61,8 +86,12 @@ def get_run_args(mode, args, run_conf):
...
-    for key in run_args:
-        if run_args.get(key) != None:
-            args_str += f'{key} {run_args.get(key)} '
+    args_str = make_args(run_args)
     run_tests(item, args_str)
 
+#run doubled range exist
+run_args = get_run_args(item, parsed_args, test_run_config)
+run_args['-c'] = int(run_args['-c']) * 2 + 1
+run_args['-s'] = 5001
```
ikopylov, line 95: Update `run_args` inside `run_doubled_exist_test`, because `run_doubled_exist_test` currently depends on the specific value (if you change `-c` value, it fails)

integration-tests/tests.py
```diff
@@ -21,18 +21,43 @@ def run_tests(behaviour, args):
...
+def run_doubled_exist_test(args):
+    try:
+        print(f'Running bobp -b exist {args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b exist {args.rstrip()}')).decode('ascii')
+        print(str(p))
+        found_exist = re.search(r'\b[0-9]{1,}\sof\s[0-9]{1,}\b', str(p))
+        if not found_exist:
+            sys.exit(f"No exist output captured, check output")
+        exists = found_exist.group(0).split(' of ')
+        if int(exists[0]) * 2 + 1 != int(exists[1]):
```
ikopylov, line 41: Bad check. Pass `expected_exists_keys` and compare with its value: `if int(exists[0]) != expected_exists_keys:`

integration-tests/tests.py
```diff
@@ -21,18 +21,43 @@ def run_tests(behaviour, args):
...
+            sys.exit(f"{exists[0]} of {exists[1]} keys, expected {int(int(exists[1]) / 2)} of {exists[1]} instead, exist test failed, see output")
```
ikopylov, line 42: Here the message should also be updated

integration-tests/tests.py
```diff
@@ -21,18 +21,47 @@ def run_tests(behaviour, args):
...
         sys.exit(str(e.stderr))
     except Exception as e:
         sys.exit(str(e))
 
+def run_doubled_exist_test(run_args, expected_exist_keys):
+    try:
+        run_args['-c'] = expected_exist_keys * 2 + 1
+        run_args['-s'] = 5001
+        run_args['-t'] = 1
+        args = make_args(run_args)
```
ikopylov, line 37: `make_args` declared blow. Due to this the function probably is not available here

integration-tests/tests.py
```diff
@@ -21,14 +21,43 @@ def run_tests(behaviour, args):
...
+        print(f'Running bobp -b exist {args.rstrip()}')
+        p = subprocess.check_output(shlex.split(f'./bobp -b exist {args.rstrip()}')).decode('ascii')
+        print(str(p))
+        found_exist = re.search(r'\b[0-9]{1,}\sof\s[0-9]{1,}\b', str(p))
+        if not found_exist:
+            sys.exit(f"No exist output captured, check output")
+        exists = found_exist.group(0).split(' of ')
+        if int(exists[0]) != expected_exist_keys:
+            sys.exit(f"{exists[0]} of {exists[1]} keys, expected {expected_exist_keys} of {exists[1]} instead, exist test failed, see output")
```
ikopylov, line 53: Looks like python does not print the message, passed to `exit`. It is better to print it explicitly. E.g. you can declare function `print_then_exit` that prints message and the call `sys.exit` with the same message

ikopylov (approved):
LGTM

ikopylov (approved):
LGTM

PR #766, opened 2023-01-23
748 background fsync

### LINKED ISSUE #748: Significant `put` performance drop due to `fsync_all` call on blob close
We must find the best way to sync data. Perhaps we can run `fsync_all` in background. Or run `fsync_all` periodically.

Associated issue in Pearl: https://github.com/qoollo/pearl/issues/234

Boneyan:
Blob performes fsync if buffered bytes are larger than max_dirty_bytes_before_sync config param
Closes #748


ikopylov (approved):
Trigger integration tests

PR #775, opened 2023-02-19
[774] BobClient reduce clone overhead

### LINKED ISSUE #593: Optimize mapper performance
Use HashMap instead of iteration where possible (example: `get_operation` at https://github.com/qoollo/bob/blob/master/bob-common/src/mapper.rs#L184).
Avoid data cloning where possible (example: `DiskPath::from` at https://github.com/qoollo/bob/blob/master/bob-common/src/data.rs#L283)

### LINKED ISSUE #724: Place `Node` struct internals under `Arc`
Node cloned many times and it consists of 2 string fields. It would be good to have `Arc` to reduce copy overhead

https://github.com/qoollo/bob/blob/master/bob-common/src/node.rs#L19

### LINKED ISSUE #774: Reduce `BobClient` clone overhead
All fields except `client` can be shared between clones with the help of `Arc`. This significantly reduce clone overhead.
Additionally, `node` field can be replaced with `target_node_name` and `target_node_address` fields. This also reduce memory footprint

ikopylov:
Closes #774
Closes #724
Closes #593 

Covers required changes from https://github.com/qoollo/bob/pull/653


bob-common/src/configs/cluster.rs
```diff
@@ -428,62 +418,60 @@ impl Cluster {
 impl Validatable for Cluster {
     fn validate(&self) -> Result<(), String> {
         if self.nodes.is_empty() {
-            let msg = "bob requires at least one node to start".to_string();
-            error!("{}", msg);
-            return Err(msg);
+            error!("bob requires at least one node to start");
```
idruzhitskiy, line 421: Why duplicate logs here while err will be processed up the stack?

bob-common/src/name_types.rs
```diff
@@ -0,0 +1,235 @@
...
+    where
+        E: serde::de::Error, 
+    {
+        Ok(v.into())
+    }
+}
+
+// ============= NodeName =============
+
+impl NodeName {
```
idruzhitskiy, line 57: Maybe it should be a generic type with marker, like `Name<Node>`, `Name<Disk>`? So code for them can be unified

idruzhitskiy: And then we can make `NodeName = Name<Node>`, `DiskName = Name<Disk>`.

bob-common/src/configs/cluster.rs
ikopylov: Updated. Now errors are printed on the top level.

I checked that at this point logger is not initialized, thus this messages goes nowhere

bob-common/src/name_types.rs
ikopylov: Interesting idea. I've implemented it

bob-common/src/bob_client.rs
```diff
@@ -58,24 +62,27 @@ pub mod b_client {
...
+                local_node_name: local_node_name, 
+                operation_timeout: operation_timeout, 
+                auth_header: auth_header, 
+                metrics: metrics
             })
         }
 
         // Getters
-
         #[allow(dead_code)]
```
idruzhitskiy, line 79: Are those allows needed?

bob-common/src/configs/cluster.rs
```diff
@@ -345,9 +317,8 @@ impl Cluster {
     pub async fn try_get(filename: &str) -> Result<Self, String> {
         let config = YamlBobConfig::get::<Cluster>(filename).await?;
         if let Err(e) = config.validate() {
-            let msg = format!("config is not valid: {}", e);
-            error!("{}", msg);
-            Err(msg)
+            debug!("Cluster config is not valid: {}", e);
```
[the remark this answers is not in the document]
ikopylov: This is single `debug` log on the config loading path. I think it is better to keep it, just in case. It can be helpful in tests

bob-common/src/configs/cluster_tests.rs
```diff
@@ -585,7 +584,7 @@ pearl:
   settings:                     # describes how create and manage bob directories. required for 'pearl'
     root_dir_name: bob            # root dir for bob storage. required for 'pearl'
     alien_root_dir_name: alien    # root dir for alien storage in 'alien_disk'. required for 'pearl'
-    timestamp_period: 2w      # period when new pearl directory created. required for 'pearl'
+    timestamp_period: 500d      # period when new pearl directory created. required for 'pearl'
```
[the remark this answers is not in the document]
ikopylov: There was an artificial limit on the `timestamp_period` value equal to 1 week (https://github.com/qoollo/bob/blob/master/bob-common/src/configs/node.rs#L71). There may be situations when there is not much data and period can be set to a larger value. Thus I extended this limit to 1 year and updated the associated test

bob-common/src/bob_client.rs
ikopylov: No, they are not. I removed them all

bob-common/src/configs/cluster.rs
```diff
@@ -1,30 +1,27 @@
...
-        if self.name().is_empty()
+        if self.name().as_str().is_empty()
             || self.path().is_empty()
             || self.name() == "~"
             || self.path() == "~"
         {
-            let msg = "node disks must contain not empty fields \'name\' and \'path\'".to_string();
-            error!("NodeDisk validation failed: {}", msg);
-            Err(msg)
+            Err("NodeDisk validation failed: node disks must contain not empty fields \'name\' and \'path\'".to_string())
```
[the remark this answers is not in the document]
ikopylov: I removed them all

bob-common/src/core_types.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    pub fn nodes(&self) -> &[Node] {
+        &self.nodes
+    }
+}
+
+impl DiskPath {
+    /// Creates new `DiskPath` with disk's name and path.
+    #[must_use = "memory allocation"]
```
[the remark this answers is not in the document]
ikopylov: It seems that these attributes are not needed here.
https://rustwiki.org/en/reference/attributes/diagnostics.html#the-must_use-attribute

idruzhitskiy (approved):
LGTM

PR #787, opened 2023-04-10
782 panic abort

### LINKED ISSUE #782: Terminate Bob on thread panic
https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html

`panic = 'abort'` should be added to all profiles

idruzhitskiy:
Closes #782

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,15 +8,19 @@ members = [
 ]
 
 
-[profile.release-lto]
-inherits = "release"
-lto = true
-
 [profile.release]
 panic = "abort"
 
+[profile.release-lto]
+inherits = "release"
+panic = "abort"
+lto = true
+
 [profile.dev]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -12,9 +12,15 @@ members = [
 inherits = "release"
 lto = true
 
+[profile.release]
+panic = "abort"
```
ikopylov, line 16: You should add this line to `release-lto` and to `integration-test` profiles as well. The only profile, that should preserve unwind behavior is `test`

idruzhitskiy: `release-lto` inherits `release`, so no reason to duplicate it here. I'll add one to `integration-test`

idruzhitskiy: Done

PR #789, opened 2023-05-05
788 prometheus leak fix

### LINKED ISSUE #788: Memory leak in Prometheus exporter from `metrics` crate
According to the issue https://github.com/metrics-rs/metrics/issues/245 Prometheus exporter has a memory leak inside. When no one requests Prometheus counters, the values are accumulated internally without clearing. After a day of work under the load, the memory consumed by the collector increased by ~1Gb. 
We should run the background task, that retrieves metrics periodically. Or wait until the issue will be fixed inside the library

idruzhitskiy:
Closes #788

### DIFF
--- a/bob-common/src/metrics/mod.rs
+++ b/bob-common/src/metrics/mod.rs
@@ -124,6 +124,8 @@ pub const TOTAL_SPACE: &str = "hardware.total_space";
 
 const CLIENTS_METRICS_DIR: &str = "clients";
 
+const PROMETHEUS_RENDER_INTERVAL_SEC: u64 = 3600;
+
 /// Type to measure time of requests processing
 pub type Timer = Instant;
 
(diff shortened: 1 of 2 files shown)


bob-common/src/metrics/mod.rs
```diff
@@ -390,6 +392,17 @@ fn build_prometheus(node_config: &NodeConfig) -> PrometheusRecorder {
         }
     };
     tokio::spawn(future);
+
+    let handle = recorder.handle();
```
ikopylov, line 396: Please, add comment, why this logic is needed (also add a full url of associated issue: https://github.com/qoollo/bob/issues/788)

ikopylov (approved):
LGTM

ikopylov (approved):
LGTM

PR #809, opened 2023-05-24
[807] Do not stop initialization if disk is not available v2.0

ikopylov:


bob/src/hw_metrics_collector.rs
```diff
@@ -24,20 +24,30 @@ impl HWMetricsCollector {
...
-        System::new_all()
+        System::new_with_specifics(RefreshKind::new().with_disks_list().with_disks())
             .disks()
             .iter()
             .filter_map(|d| {
                 let path = d.mount_point();
+                let path_metadata = match path.metadata() {
+                    Ok(meta) => meta,
+                    Err(err) => {
+                        warn!("Can't get metadata from OS for disk '{}' returned by 'sysinfo': {:?}", path.display(), err);
```
idruzhitskiy, line 35: How often this message will be displayed? Shouldn't it be debug?

ikopylov: Once for each problematic disk. `collect_used_disks` called only once at startup

idruzhitskiy (approved):
LGTM

PR #813, opened 2023-05-30
797 memory leak on delete

### LINKED ISSUE #797: Memory leak during record deletion
Reproduces with single node cluster with single vdisk (configs from `config-examples`) with the following `bobp` command:
```
bobp -b test -c 500000 --user admin --password password -t 6 
```

Perhaps the bug can be connected with bug in ObserverWorker, that never stops. This bug was fixed in https://github.com/qoollo/pearl/pull/265, thus the test should be done again after the new Pearl will be released

idruzhitskiy:
Closes #797

### DIFF
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -75,7 +75,7 @@ version = "0.19.0"
 mimalloc = { version = "0.1", default-features = false, optional = true }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
-jemallocator ={ version = "0.5.0", optional = true }
+jemallocator = { version = "0.5.0", optional = true }
 
 [build-dependencies]
 tonic-build = "0.6"
--- a/bob-apps/Cargo.toml
+++ b/bob-apps/Cargo.toml
@@ -18,6 +18,7 @@ edition = "2021"
 default = ["mimalloc"]
 mimalloc = ["dep:mimalloc"]
 mimalloc-secure = ["dep:mimalloc", "mimalloc/secure"]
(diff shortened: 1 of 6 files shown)


bob-apps/Cargo.toml
```diff
@@ -18,6 +18,9 @@ edition = "2021"
 default = ["mimalloc"]
 mimalloc = ["dep:mimalloc"]
 mimalloc-secure = ["dep:mimalloc", "mimalloc/secure"]
+# Important: can't be used with panic = "abort"
+# Please use profile "jemalloc-profile"
+jemalloc-profile = ["jemallocator", "jemallocator/profiling"]
```
ikopylov, line 23: Please, use `dep` prefix for dependencies: 
```suggestion
jemalloc-profile = ["dep:jemallocator", "jemallocator/profiling"]
```

idruzhitskiy: It doesn't work with features, "jemallocator/profiling" can't be used with "dep:jemallocator"

idruzhitskiy: I've fixed mimalloc-secure as it was not compiling as well

bob-apps/bin/bobd.rs
```diff
@@ -1,4 +1,9 @@
+#[cfg(all(any(feature = "mimalloc", feature = "mimalloc-secure"), any(feature = "jemallocator", feature = "jemallocator-profile")))]
+compile_error!("features `mimalloc` and `jemalloc-profile` are mutually exclusive");
```
ikopylov, line 2:
```suggestion
compile_error!("features `mimalloc` and `jemallocator` are mutually exclusive");
```

PR #814, opened 2023-06-01
808 more delete logs

### LINKED ISSUE #808: Extend logs in bobd to detect the error, received by OldPartitionRemover
It would be good to add more logs to detect the problem from this issue: https://github.com/qoollo/bob-tools/issues/75

idruzhitskiy:
Closes #808

### DIFF
--- a/bob/src/api/mod.rs
+++ b/bob/src/api/mod.rs
@@ -367,8 +367,8 @@ async fn find_group<A: Authenticator>(
         .find(|dc| dc.vdisks().iter().any(|&vd| vd == vdisk_id))
         .ok_or_else(|| {
             let dcs = dcs.iter()
-                .map(|dc| format!("DC: {:?}, vdisks: {}",
-                                  dc.disk(),
+                .map(|dc| format!("DC: {}, vdisks: {}",
+                                  dc.disk().name(),
                                   dc.vdisks().iter().map(|v| format!("#{}", v)).collect::<Vec<_>>().join(", ")))
                 .collect::<Vec<_>>()
                 .join(", ");
--- a/bob/src/api/mod.rs
+++ b/bob/src/api/mod.rs
@@ -366,12 +366,22 @@ async fn find_group<A: Authenticator>(
         .iter()
         .find(|dc| dc.vdisks().iter().any(|&vd| vd == vdisk_id))
         .ok_or_else(|| {
-            let err = format!("Disk Controller with vdisk #{} not found", vdisk_id);
+            let dcs = dcs.iter()
+                .map(|dc| format!("DC: {:?}, vdisks: {}",
+                                  dc.disk(),
+                                  dc.vdisks().iter().map(|v| format!("#{}", v)).collect::<Vec<_>>().join(", ")))
+                .collect::<Vec<_>>()
+                .join(", ");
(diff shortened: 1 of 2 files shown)


bob/src/api/mod.rs
```diff
@@ -366,12 +366,22 @@ async fn find_group<A: Authenticator>(
         .iter()
         .find(|dc| dc.vdisks().iter().any(|&vd| vd == vdisk_id))
         .ok_or_else(|| {
-            let err = format!("Disk Controller with vdisk #{} not found", vdisk_id);
+            let dcs = dcs.iter()
+                .map(|dc| format!("DC: {:?}, vdisks: {}",
+                                  dc.disk(),
```
ikopylov, line 371: It is better to print only the name
```suggestion
                .map(|dc| format!("DC: {}, vdisks: {}",
                                  dc.disk().name(),
```

bob/src/api/mod.rs
```diff
@@ -366,12 +366,22 @@ async fn find_group<A: Authenticator>(
...
+            let err = format!("Disk Controller with vdisk #{} not found, available dcs: {}", vdisk_id, dcs);
             warn!("{}", err);
             StatusExt::new(StatusCode::NOT_FOUND, false, err)
         })?;
-    needed_dc.vdisk_group(vdisk_id).await.map_err(|_| {
-        let err = format!("Disk Controller with vdisk #{} not found", vdisk_id);
+    needed_dc.vdisk_group(vdisk_id).await.map_err(|e| {
+        let err = format!("VDiskGroup #{} is missing on disk controller {:?}, error: {:?}, available vdisks: {}", 
+                          vdisk_id,
+                          needed_dc.disk(),
```
ikopylov, line 382: Same here
```suggestion
        let err = format!("VDiskGroup #{} is missing on disk controller '{}', error: {:?}, available vdisks: {}", 
                          vdisk_id,
                          needed_dc.disk().name(),
```

bob/src/api/mod.rs
```diff
@@ -366,12 +366,22 @@ async fn find_group<A: Authenticator>(
...
         })?;
-    needed_dc.vdisk_group(vdisk_id).await.map_err(|_| {
-        let err = format!("Disk Controller with vdisk #{} not found", vdisk_id);
+    needed_dc.vdisk_group(vdisk_id).await.map_err(|e| {
+        let err = format!("VDiskGroup #{} is missing on disk controller '{}', error: {:?}, available vdisks: {}", 
+                          vdisk_id,
+                          needed_dc.disk().name(),
+                          e,
+                          needed_dc.vdisks().iter().map(|v| format!("#{}", v)).collect::<Vec<_>>().join(", "));
         warn!("{}", err);
```
ikopylov, line 385: Noticed, that error `e` also passed to the client. I think it is not good, because error can contain sensitive information. I suggest to remove `e` from this common message and add it only to `warn!`:

```suggestion
        let err = format!("VDiskGroup #{} is missing on disk controller '{}', available vdisks: {}", 
                          vdisk_id,
                          needed_dc.disk().name(),
                          needed_dc.vdisks().iter().map(|v| format!("#{}", v)).collect::<Vec<_>>().join(", "));
        warn!("{}. Error: {:?}", err, e);
```

ikopylov (approved):
LGTM

PR #827, opened 2023-08-02
[823] add detailed disk metrics

### LINKED ISSUE #823: Metrics: add detailed information about total, used and free space on every disk
Also add these data to the result of `/status/space`

archeoss:
Closes #823


bob/src/api/mod.rs
```diff
@@ -145,12 +145,18 @@ pub(crate) struct NodeConfiguration {
 }
 
 #[derive(Debug, Serialize)]
-pub(crate) struct SpaceInfo {
+struct Space {
     total_disk_space_bytes: u64,
     free_disk_space_bytes: u64,
     used_disk_space_bytes: u64,
     occupied_disk_space_bytes: u64,
```
ikopylov, line 152: If we decide to allow breaking changes, then this field should be renamed to `occupied_by_bob_disk_space_bytes`. Such a name will better reflect the meaning of the field

bob/src/api/mod.rs
```diff
@@ -313,31 +319,69 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
...
-        map.insert(dc.disk().name().to_string(), dc.disk_used().await);
+    let mut disk_space_by_disk = HashMap::new();
+    let mut total_occupied = 0;
+    for dc in dcs {
+        let occupied_space = dc.disk_used().await;
+        total_occupied += occupied_space;
+        let DiskSpaceMetrics {
+            total_space,
+            used_space,
+            free_space,
```
ikopylov, line 341: This decomposition is not needed, because below you create a new struct anyway. That new struct can be easily initialized from the fields of `DiskSpaceMetrics`

bob/src/api/mod.rs
```diff
@@ -313,31 +319,69 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
...
+        } = disk_metrics
+            .remove_entry(dc.disk().name())
```
ikopylov, line 343: This `remove_entry` is also not needed. It only slows down this function

bob/src/api/mod.rs
```diff
@@ -313,31 +319,69 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
...
+        .and_modify(|s| s.occupied_disk_space_bytes += adc_space)
+        .or_insert({
+            let DiskSpaceMetrics {
+                total_space,
+                used_space,
+                free_space,
+            } = disk_metrics
+                .remove_entry(adc.disk().name())
+                .map(|(_, space)| space)
+                .unwrap_or_default();
```
ikopylov, line 368: Same comments on this approach

bob/src/hw_metrics_collector.rs
```diff
@@ -1,21 +1,23 @@
...
 use std::path::{Path, PathBuf};
 use std::process::{self, Command};
-use std::fs::read_to_string;
 use sysinfo::{DiskExt, ProcessExt, ProcessRefreshKind, RefreshKind, System, SystemExt};
 
 const DESCRS_DIR: &str = "/proc/self/fd/";
 const CPU_STAT_FILE: &str = "/proc/stat";
 const DISK_STAT_FILE: &str = "/proc/diskstats";
 
+#[derive(Debug, Serialize, Default, Clone)]
```
ikopylov, line 20: `Serialize` is not needed here

bob/src/hw_metrics_collector.rs
```diff
@@ -91,10 +93,10 @@ impl HWMetricsCollector {
             match cpu_s_c.iowait() {
                 Ok(iowait) => {
                     gauge!(CPU_IOWAIT, iowait);
-                },
+                }
```
ikopylov, line 96: Most of the time formatting changes leads to a merge conflicts in the future. Right now you can keep them, but in the future requests, please, do not commit formatting changes in the code, that is not directly affected by the changes, required by issue itself

bob/src/hw_metrics_collector.rs
```diff
@@ -124,11 +129,14 @@ impl HWMetricsCollector {
...
 
-    fn update_space_metrics_from_disks(disks: &HashMap<PathBuf, DiskName>) -> DiskSpaceMetrics {
+    fn update_space_metrics_from_disks(
+        disks: &HashMap<PathBuf, DiskName>,
+    ) -> HashMap<DiskName, DiskSpaceMetrics> {
         let disks_metrics = Self::space(disks);
-        gauge!(TOTAL_SPACE, bytes_to_mb(disks_metrics.total_space) as f64);
-        gauge!(USED_SPACE, bytes_to_mb(disks_metrics.used_space) as f64);
-        gauge!(FREE_SPACE, bytes_to_mb(disks_metrics.free_space) as f64);
+        let summed_space: DiskSpaceMetrics = disks_metrics.values().sum();
```
ikopylov, line 136: Unfortunately, this sum is incorrect, because multiple bob disks can be mapped to a single physical disk

bob/src/api/mod.rs
```diff
@@ -313,31 +319,69 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
 async fn get_space_info<A: Authenticator>(
     bob: Extension<BobServer<A>>,
 ) -> Result<Json<SpaceInfo>, StatusExt> {
+    let mut disk_metrics = bob.grinder().hw_counter().update_space_metrics();
     let DiskSpaceMetrics {
         total_space,
         used_space,
         free_space,
-    } = bob.grinder().hw_counter().update_space_metrics();
+    } = disk_metrics.values().sum();
```
ikopylov, line 327: Same here: this sum is incorrect, because multiple bob disks can be mapped to a single physical disk

bob/src/api/mod.rs
```diff
@@ -145,12 +145,18 @@ pub(crate) struct NodeConfiguration {
...
     used_disk_space_bytes: u64,
     occupied_disk_space_bytes: u64,
-    occupied_disk_space_by_disk: HashMap<String, u64>,
+}
+
+#[derive(Debug, Serialize)]
+pub(crate) struct SpaceInfo {
+    #[serde(flatten)]
+    total_space: Space,
+    disk_space_by_disk: HashMap<DiskName, Space>,
```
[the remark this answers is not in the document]
ikopylov: There is another problem here: you try to combine logical and physical structure in this map. Probably, it is better to split them. `occupied_disk_space_by_disk` is created to represent space info within logical structure. I suggest to keep it without changes (`occupied_disk_space_by_disk: HashMap<DiskName, u64>`) and add additional `disk_space_by_disk` for physical structure (its key will be a mount point of the disk, not its name)

```Rust
pub(crate) struct SpaceInfo {
    #[serde(flatten)]
    total_space: Space,
    occupied_disk_space_by_disk: HashMap<DiskName, u64>,  // key is a Bob disk name
    disk_space_by_disk: HashMap<String, Space>  // Key is mount point
}
```

bob/src/hw_metrics_collector.rs
ikopylov: I suggest to return a map from `space` function, which key is a mount point (not DiskName) and DiskName can be a field of `DiskSpaceMetrics` structure

archeoss: Mount point as a key is even worse alternative to the `DiskName` because there are usually multiple mount points for the same Bob disk (but this may be changed in #828). The best way to ensure the uniqueness of physical disks is to map them to the `f_fsid` of `statvfs` system call. In that way we don't depend on any outer function to provide correct map of mount points and disk names and ensure the uniqueness of the result

bob/src/api/mod.rs
archeoss: see comment above

bob/src/hw_metrics_collector.rs
archeoss: Although I'm not sure that f_fsid is unique across multiple physical disks...

bob/src/api/mod.rs
```diff
@@ -145,12 +145,25 @@ pub(crate) struct NodeConfiguration {
...
+
+    /// Total occupied disk space by Bob
     occupied_disk_space_bytes: u64,
-    occupied_disk_space_by_disk: HashMap<String, u64>,
+
+    /// Key - Bob disk name
+    occupied_disk_space_by_disk: HashMap<DiskName, u64>,
+    
+    /// Key - Mount point or Bob disk name if no mount point was found
+    disk_space_by_disk: HashMap<String, Space>,
```
ikopylov, line 166: I just thought that it would be better to use disk name as the key (on Linux it is `sda1`, `sdb1`, on Windows it is `C:`, `D:`), because this is the common way of identification of disks in OS

bob/src/api/mod.rs
```diff
@@ -313,31 +326,42 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
...
+    let disk_space_by_disk = disk_metrics.into_values().map(|disk| (
+        disk_path.get(&disk.disk_name.to_string()).unwrap_or(&disk.disk_name.to_string()).clone(), 
+        Space {
+            total_disk_space_bytes: disk.total_space,
+            free_disk_space_bytes: disk.free_space,
+            used_disk_space_bytes: disk.used_space 
+        }
+    )).collect();
+
     let adc_space = adc.disk_used().await;
```
ikopylov, line 351: Move the code that adds entry for aliens above to the place where `occupied_disk_space_by_disk` is initialized (line 339)

bob/src/hw_metrics_collector.rs
```diff
@@ -157,30 +167,35 @@ impl HWMetricsCollector {
...
 
-    fn space(disks: &HashMap<PathBuf, DiskName>) -> DiskSpaceMetrics {
-        let mut total = 0;
-        let mut used = 0;
-        let mut free = 0;
-
-        for mount_point in disks.keys() {
+    /// Maps [`DiskSpaceMetrics`] to the corresponding mount point.
+    ///
+    /// `f_fsid` field of [`statvfs`] is used to ensure that disks are unique.
```
ikopylov, line 172: According to the code of the function, you return mount point as the key. Please, fix the comment

bob/src/hw_metrics_collector.rs
```diff
@@ -63,7 +67,10 @@ impl HWMetricsCollector {
         tokio::spawn(Self::task(self.interval_time, self.disks.clone()));
     }
 
-    pub(crate) fn update_space_metrics(&self) -> DiskSpaceMetrics {
+    /// Returns the updated space metrics of this [`HWMetricsCollector`].
+    ///
+    /// Key -- `f_fsid` of underlying [`statvfs`] function, represents file system ID
```
ikopylov, line 72: Same here. Comment is incorrect

bob/src/api/mod.rs
```diff
@@ -313,31 +326,42 @@ async fn status<A: Authenticator>(bob: Extension<BobServer<A>>) -> Json<Node> {
...
-    }
+    
+    let mut occupied_disk_space_by_disk: HashMap<DiskName, u64> = futures::future::join_all(
+        dcs.iter().map(|dc| async { ( dc.disk().name().clone(), dc.disk_used().await ) })
+    ).await.into_iter().collect();
+
+    let disk_path: HashMap<_, _> = dcs.iter().map(|dc| (dc.disk().name().to_string(), dc.disk().path().to_string())).collect();
+
+    let disk_space_by_disk = disk_metrics.into_values().map(|disk| (
+        disk_path.get(&disk.disk_name.to_string()).unwrap_or(&disk.disk_name.to_string()).clone(),
```
ikopylov, line 343: This transition is not needed. `disk_metrics` already have mount path as the key

bob/src/api/mod.rs
ikopylov: I think, it would be better to do later in another request. Right now, mount point would be enough here

bob/src/hw_metrics_collector.rs
```diff
@@ -16,7 +18,9 @@ const DESCRS_DIR: &str = "/proc/self/fd/";
 const CPU_STAT_FILE: &str = "/proc/stat";
 const DISK_STAT_FILE: &str = "/proc/diskstats";
 
+#[derive(Debug, Clone)]
 pub(crate) struct DiskSpaceMetrics {
+    pub(crate) disk_name: DiskName,
```
ikopylov, line 23: It's not a good idea to put the disk_name filed in this struct, since the mapping is ambiguous (multiple bob disks can be placed on single physical disk). Also, this field will not be used anywhere after the corrections will be made accrding to my previous comments. Thus I would suggest to remove this field

bob/src/hw_metrics_collector.rs
```diff
@@ -157,30 +167,35 @@ impl HWMetricsCollector {
...
-                let bavail = stat.f_bavail as u64;
-                let bfree = stat.f_bfree as u64;
-                total += bsize * blocks;
-                free += bsize * bavail;
-                used += (blocks - bfree) * bsize;
+                let bsize = stat.f_bsize;
+                let blocks = stat.f_blocks;
+                let bavail = stat.f_bavail;
+                let bfree = stat.f_bfree;
+                fs_ids.insert(stat.f_fsid).then(||
```
ikopylov, line 184: Please, use regular `if` instead of `then`. `if` gives better readability

bob/src/hw_metrics_collector.rs
```diff
@@ -157,30 +167,35 @@ impl HWMetricsCollector {
...
-                let blocks = stat.f_blocks as u64;
-                let bavail = stat.f_bavail as u64;
-                let bfree = stat.f_bfree as u64;
-                total += bsize * blocks;
-                free += bsize * bavail;
-                used += (blocks - bfree) * bsize;
+                let bsize = stat.f_bsize;
+                let blocks = stat.f_blocks;
+                let bavail = stat.f_bavail;
+                let bfree = stat.f_bfree;
```
ikopylov, line 183: Please, return back explicit type declaration, because it is required for proper calculations. On some targets the type of all this values can be 'u32' instead of 'u64', so it is better to explicitly convert it to `u64` here to avoid overflow on multiplication operation later

archeoss: *nerd mode on*
*Theoretically*, if that's the case, it shouldn't even build at this point. Multiplication of two `u32` numbers produces another `u32` number, but  `DiskSpaceMetrics` has `u64` field, and rust prohibits **any** implicit type conversions, thus the compiler should produce the error
In order to convert the type we must use either `as <type>` syntax or `.into()` method
*nerd mode off*
But maybe I'm missing something

In any case, the type annotations are placed back, they wouldn't hurt

bob/src/api/mod.rs
```diff
@@ -145,12 +145,25 @@ pub(crate) struct NodeConfiguration {
...
 #[derive(Debug, Serialize)]
-pub(crate) struct SpaceInfo {
+struct Space {
     total_disk_space_bytes: u64,
     free_disk_space_bytes: u64,
     used_disk_space_bytes: u64,
+}
+
+#[derive(Debug, Serialize)]
+pub(crate) struct SpaceInfo {
```
idruzhitskiy, line 155: Please update openapi in `config-examples`

bob/src/hw_metrics_collector.rs
```diff
@@ -63,7 +66,11 @@ impl HWMetricsCollector {
         tokio::spawn(Self::task(self.interval_time, self.disks.clone()));
     }
 
-    pub(crate) fn update_space_metrics(&self) -> DiskSpaceMetrics {
+    /// Returns the updated space metrics of this [`HWMetricsCollector`].
```
idruzhitskiy, line 75: Most of the time when we need result of this method we also need sum of spaces. I suggest using new structure with existing map and total already calculated.

archeoss: Something like this?
```rust
#[derive(Debug, Clone)]
pub(crate) struct DiskSpaceMetrics {
    pub(crate) total_space: u64,
    pub(crate) used_space: u64,
    pub(crate) free_space: u64,
    pub(crate) per_disk: HashMap<PathBuf, SpaceMetrics>
}

#[derive(Debug, Clone)]
pub(crate) struct SpaceMetrics {
    pub(crate) total_space: u64,
    pub(crate) used_space: u64,
    pub(crate) free_space: u64,
}
    ...
pub(crate) fn update_space_metrics(&self) -> DiskSpaceMetrics {
    ...
}
```

bob/src/api/mod.rs
archeoss: I think it would be good to integrate [utoipa](https://github.com/juhaku/utoipa) for openapi auto-gen at some point

bob/src/hw_metrics_collector.rs
idruzhitskiy: I guess just `total: SpaceMetrics` will do, since we already have this structure

idruzhitskiy (commented):
LGTM

PR #828, opened 2023-08-09
[825] Incorrect disk space calculation inside docker container 

### LINKED ISSUE #825: Incorrect disk space calculation inside docker container
Looks like we should use `frsize` instead of `bsize` here: https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L169. But this should be tested.

Statvfs struct declaration: https://man7.org/linux/man-pages/man3/statvfs.3.html

Here is the issue for MacOs: https://github.com/docker/for-mac/issues/2136

archeoss:
Closes #825 

Note 1: usage of frsize instead of bsize seems to do nothing (at least on Linux). Needs to be tested on some Mac machine, where it's first occured. Also this change about frsize needs to be done after #823 will be merged, since that function was rewritten there


ikopylov:
Please, do not forget to fix space calculation from results of `statvfs` syscall. It should use `frsize` instead of `bsize`

bob/src/hw_metrics_collector.rs
```diff
@@ -36,29 +36,68 @@ impl HWMetricsCollector {
...
+                .mount_point()
+                .metadata()
+                .expect("Can't get metadata from OS");
+            res.entry(metadata.dev()).or_insert(disk);
+            if dev_path
+                .iter()
+                .filter(|(dev, _)| *dev == metadata.dev())
+                .any(|(_, path)| disk.mount_point().file_name() == path.file_name())
+            {
+                res.insert(metadata.dev(), disk);
```
ikopylov, line 94: Please, add comment that this `insert` replaces previous value in `HashMap`

bob/src/hw_metrics_collector.rs
```diff
@@ -36,29 +36,68 @@ impl HWMetricsCollector {
...
+        for disk in sys_disks {
+            let metadata = disk
+                .mount_point()
+                .metadata()
+                .expect("Can't get metadata from OS");
+            res.entry(metadata.dev()).or_insert(disk);
+            if dev_path
+                .iter()
+                .filter(|(dev, _)| *dev == metadata.dev())
+                .any(|(_, path)| disk.mount_point().file_name() == path.file_name())
```
ikopylov, line 92: This is not correct. We should check that `disk.mount_point()` is a prefix of `path`. Example: `mount_point = "/opt/disk1/", path = "/opt/disk1/bob"`. Use `starts_with` function

bob/src/hw_metrics_collector.rs
ikopylov, line 94: Please, add a comment that this `insert` replaces the previous value

PR #831, opened 2023-08-31
830 lsof zombie fix

### LINKED ISSUE #830: Bob produces `lsof` zombie child processes
`lsof` is used in hardware metrics to calculate the number of opened file descriptors (https://github.com/qoollo/bob/blob/master/bob/src/hw_metrics_collector.rs#L492). For unknown reasons, these processes sometimes become zombies. The problem was noticed when working in docker

idruzhitskiy:
Closes #830

### DIFF
--- a/bob/src/hw_metrics_collector.rs
+++ b/bob/src/hw_metrics_collector.rs
@@ -8,7 +8,7 @@ use bob_common::metrics::{
 use libc::statvfs;
 use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
-use std::process::{self, Command};
+use tokio::process::Command;
 use std::fs::read_to_string;
 use sysinfo::{DiskExt, ProcessExt, ProcessRefreshKind, RefreshKind, System, SystemExt};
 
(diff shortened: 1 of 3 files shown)


bob/src/hw_metrics_collector.rs
```diff
@@ -473,50 +474,63 @@ impl DescrCounter {
...
+                                                return Some(count - 5); // exclude stdin, stdout, stderr, lsof pipe and wc pipe
+                                            }
+                                            Err(e) => {
+                                                debug!("failed to parse lsof result: {}", e);
+                                            }
+                                        }
+                                    },
                                     Err(e) => {
-                                        debug!("failed to parse lsof result: {}", e);
+                                        debug!("can't use lsof, wc error (fs /proc will be used): {}", e);
```
ikopylov, line 518: Why are the error messages swapped?

ikopylov:
Have you checked that descriptors counting works properly after these changes?

bob/src/hw_metrics_collector.rs
idruzhitskiy: There is another level of match added

ikopylov: Good. Git diff a little bit misleading here. In the whole file I see, that everything is correct.

The level of nesting became very large, but I did not find an easy way to refactor it, since we process each error independently

ikopylov (approved):
LGTM

PR #850, opened 2023-10-20
849 grpc error save

### LINKED ISSUE #849: Save gRPC error info in case of unparsable error

idruzhitskiy:
Closes #849

### DIFF
--- a/bob-common/src/error.rs
+++ b/bob-common/src/error.rs
@@ -156,7 +156,7 @@ impl From<Status> for Error {
         let name = words.next();
         let length = status.message().len();
         match name {
-            None => Self::failed(format!("Can't parse status from {}", status.message())),
+            None => Self::failed(format!("Can't parse status from {:?}, {}", status.code(), status.message())),
             Some(name) => match name {
                 "KeyNotFound" => parse_next(words, Self::key_not_found)
                     .unwrap_or_else(|| Self::failed(format!("Failed to parse key from {}", status.message()))),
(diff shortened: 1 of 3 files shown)


bob-common/src/error.rs
```diff
@@ -156,23 +156,24 @@ impl From<Status> for Error {
         let name = words.next();
         let length = status.message().len();
         match name {
-            None => None,
+            None => Self::failed(format!("Can't parse status from {}", status.message())),
```
ikopylov, line 159: It would be good to add `status.code()` into the message too

bob-common/src/error.rs
```diff
@@ -156,23 +156,24 @@ impl From<Status> for Error {
...
+                "VDiskNotFound" => parse_next(words, Self::vdisk_not_found)
+                    .unwrap_or_else(|| Self::failed(format!("Failed to parse vdisk_id from {}", status.message()))),
+                "Storage" => Self::storage(rest_words(words, length)),
+                "VDiskIsNotReady" => Self::vdisk_is_not_ready(),
+                "Failed" => Self::failed(rest_words(words, length)),
+                "Internal" => Self::internal(),
+                "PearlChangeState" => Self::pearl_change_state(rest_words(words, length)),
+                "Unauthorized" => Self::unauthorized(),
+                "HolderTemporaryUnavailable" => Self::holder_temporary_unavailable(),
+                _ => Self::failed(format!("Can't parse status {:?} from {}", name, status.message())),
```
ikopylov, line 174: And here too

bob-common/src/error.rs
ikopylov: Please, use `Debug` formatter for status code, because in `Display` it returns text description, that we don't need. It is better to see just enum value name

ikopylov (approved):
LGTM

ikopylov (approved):
Trigger Integration Tests

ikopylov (approved):
Trigger integration tests

PR #852, opened 2023-10-23
try fix alien search

idruzhitskiy:


bob-backend/src/pearl/disk_controller.rs
```diff
@@ -290,35 +290,34 @@ impl DiskController {
...
 
-    async fn get_or_create_pearl(&self, operation: &Operation) -> BackendResult<Group> {
+    async fn get_or_create_first_pearl(&self, operation: &Operation) -> BackendResult<Group> {
         trace!("try get alien pearl, operation {:?}", operation);
         // block for read lock: it should be dropped before write lock is acquired (otherwise - deadlock)
         {
             let read_lock_groups = self.groups.read().await;
             let pearl = Self::find_in_groups(read_lock_groups.iter(), operation);
-            if let Some(g) = pearl {
+            if let Some(g) = pearl.first().cloned() {
```
ikopylov, line 313: This is not good. Please, use old logic that returned single Group for this particular case

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -290,35 +290,34 @@ impl DiskController {
...
 
-    async fn find_group(&self, operation: &Operation) -> BackendResult<Group> {
-        Self::find_in_groups(self.groups.read().await.iter(), operation).ok_or_else(|| {
-            Error::failed(format!("cannot find actual alien folder. {:?}", operation))
-        })
+    async fn find_groups(&self, operation: &Operation) -> BackendResult<Vec<Group>> {
+        Ok(Self::find_in_groups(self.groups.read().await.iter(), operation))
     }
 
     fn find_in_groups<'pearl, 'op>(
```
ikopylov, line 297: Split this function into 2: `find_all_groups` - returns vector, `find_single_group` - returns single result (in debug additionally checks, that there are no more matches in the list)

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -290,35 +290,34 @@ impl DiskController {
...
-            if let Some(g) = pearl {
+            if let Some(g) = pearl.first().cloned() {
                 return Ok(g);
             }
         }
 
         let mut write_lock_groups = self.groups.write().await;
         let pearl = Self::find_in_groups(write_lock_groups.iter(), operation);
-        if let Some(g) = pearl {
+        if let Some(g) = pearl.first().cloned() {
```
ikopylov, line 320: Same performance problem here

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -290,35 +290,34 @@ impl DiskController {
...
+    ) -> Vec<Group> {
         pearls
-            .find(|group| group.can_process_operation(operation))
+            .filter(|group| group.can_process_operation(operation))
             .cloned()
+            .collect()
     }
 
-    async fn get_or_create_pearl(&self, operation: &Operation) -> BackendResult<Group> {
+    async fn get_or_create_first_pearl(&self, operation: &Operation) -> BackendResult<Group> {
```
ikopylov, line 307: Naming is not good. What does it mean to `create a first pearl`? Looks like it is better to revert to original name

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -465,15 +464,11 @@ impl DiskController {
 
     pub(crate) async fn get_alien(&self, op: Operation, key: BobKey) -> Result<BobData, Error> {
         if *self.state.read().await == GroupsState::Ready {
-            let vdisk_group = self.find_group(&op).await;
-            if let Ok(group) = vdisk_group {
-                group.get(key).await
+            let mut results = self.get_alien_results_unordered(&op, |g| async move { g.get(key).await }).await.unwrap_or(vec![]);
+            results.sort_by_key(|d| d.meta().timestamp());
```
ikopylov, line 468: Pointless vector allocation and sorting. Function `get_alien_results` can return the latest result

ikopylov: Actually, `get_alien_results_unordered` is too generic. Just read the list of groups and run `for` loop over that list. Separate function make the code harder to read

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -712,4 +698,37 @@ impl DiskController {
...
+            RF: Future<Output = Result<R, Error>>,
+            F: Fn(Group) -> RF 
+    {
+        match self.find_groups(op).await {
+            Ok(groups) => {
+                if groups.is_empty() {
+                    return None;
+                }
+                let mut result = vec![];
+                let mut futures = FuturesUnordered::new();
```
ikopylov, line 713: Have you checked, that `FuturesUnordered` gives any improvement. I see 2 problems here:
1. All aliens are on the same physical disk, so any disk operation becomes a bottleneck, and parallel execution does not bring benefit
2. Most of the time there will be 0 or 1 group in the vector returned from `self.find_groups(op)`

bob-backend/src/pearl/disk_controller.rs
idruzhitskiy: Debug assert will require to update calls to `find_single_group`, since it requires Iterator, so I'll leave it out

bob-backend/src/pearl/disk_controller.rs
idruzhitskiy: This function is reused in exist and delete where we need all results

idruzhitskiy: Both look like `fold` though, so I'll add it as argument to avoid allocations in other calls

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -290,13 +290,15 @@ impl DiskController {
         Ok(groups)
     }
 
-    async fn find_group(&self, operation: &Operation) -> BackendResult<Group> {
-        Self::find_in_groups(self.groups.read().await.iter(), operation).ok_or_else(|| {
-            Error::failed(format!("cannot find actual alien folder. {:?}", operation))
-        })
+    async fn find_all_groups(&self, operation: &Operation) -> Vec<Group> {
```
ikopylov, line 293: Most of the time the result will contain 0 or 1 record. I suggest to use `SmallVec<>` here to improve performance slightly

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -712,4 +712,28 @@ impl DiskController {
...
+            return None;
+        }
+        let mut result = init;
+        for g in groups {
+            let r = f(g).await;
+            match r {
+                Ok(r) => result = fold(result, r),
+                Err(e) => {
+                    trace!("error getting alien results for op {:?}: {:?}", op, e);
+                    self.process_error(e).await;
```
ikopylov, line 733: Behavior changed here. Previously only `put` and `delete` called `process_error`. Now it is called for read operations too. Looks like it should be ok here, but in backport to v2.0 later the original behavior should be preserved as much as possible

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -465,15 +467,20 @@ impl DiskController {
 
     pub(crate) async fn get_alien(&self, op: Operation, key: BobKey) -> Result<BobData, Error> {
         if *self.state.read().await == GroupsState::Ready {
-            let vdisk_group = self.find_group(&op).await;
-            if let Ok(group) = vdisk_group {
-                group.get(key).await
+            let result = self.get_alien_results(&op, |g| async move { g.get(key).await },
```
ikopylov, line 470: It seems like if you implement loop over groups right here (remove `get_alien_results` function) then the code will be much simpler.
```Rust
let mut result = None;
for g in self.find_all_groups(op).await {
    match g.get(key).await {
          Ok(data) if data.meta().timestamp() > result.map_or(0, |d| d.meta().timestamp()) => { 
              result = data; 
          },
          Ok(_) => { },
          Err(e) if e.is_key_not_found() => { },
          Err(e) => { 
              debug!("error getting data from alien for op {:?}: {:?}", op, e);
          }
     }
}
```

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -510,16 +517,13 @@ impl DiskController {
...
-            if let Ok(group) = vdisk_group {
-                group.exist(keys).await
-            } else {
-                trace!(
-                    "EXIST[alien] No alien group has been created for vdisk #{}",
-                    operation.vdisk_id()
-                );
-                Ok(vec![false; keys.len()])
-            }
+            Ok(self.get_alien_results(&operation, |g| async move { g.exist(keys).await },
```
ikopylov, line 520: Same here

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -561,30 +565,26 @@ impl DiskController {
...
         if *self.state.read().await == GroupsState::Ready {
-            let vdisk_group =
-                if !force_delete {
-                    match self.find_group(&op).await {
-                        Ok(group) => group,
-                        Err(_) => return Ok(0)
-                    }
-                } else {
-                    // we should create group only when force_delete == true
+            if let Some(result) = self.get_alien_results(&op,
```
ikopylov, line 568: And here

bob-backend/src/pearl/disk_controller.rs
ikopylov: IMHO, a regular loop in place has much better readability than calling a function that performs folding.

In any case, I looked at the code of `delete_alien`: the error handling logic should be different there. `delete_alien` should return error if any deletion also ended with error (it should process all groups, store the latest error and return it)

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -561,30 +565,26 @@ impl DiskController {
...
-                if !force_delete {
-                    match self.find_group(&op).await {
-                        Ok(group) => group,
-                        Err(_) => return Ok(0)
-                    }
-                } else {
-                    // we should create group only when force_delete == true
+            if let Some(result) = self.get_alien_results(&op, 
+                |g| async move { g.delete(key, meta, StartTimestampConfig::new(false), force_delete).await },
+                0, |s, n| s + n).await {
```
ikopylov, line 570: Error handling logic should be different here. `delete_alien` should return error if any deletion ended with error (it should process all groups, store the latest error and return it)

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -465,15 +467,22 @@ impl DiskController {
...
+            let mut result: Option<BobData> = None;
+            for g in self.find_all_groups(&op).await {
+                match g.get(key).await {
+                    Ok(data) if data.meta().timestamp() > result.as_ref().map_or(0, |d| d.meta().timestamp()) => {
+                        result = Some(data);
+                    },
+                    Ok(_) => { },
+                    Err(e) if e.is_key_not_found() => { },
+                    Err(e) => {
+                        debug!("errorgetting data from alien for op {:?}: {:?}", op, e);
```
ikopylov, line 479:
```suggestion
                        debug!("error getting data from alien for op {:?}: {:?}", op, e);
```

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -506,20 +515,22 @@ impl DiskController {
...
-                Ok(vec![false; keys.len()])
+            let mut result = vec![false; keys.len()];
+            for g in self.find_all_groups(&op).await {
+                match g.exist(keys).await {
+                    Ok(r) => {
+                        for i in 0..r.len() {
+                            result[i] |= r[i];
+                        }
+                    },
+                    Err(e) => trace!("error getting exist results for op {:?}: {:?}", op, e),
```
ikopylov, line 530:
```suggestion
                    Err(e) => debug!("error getting exist results for op {:?}: {:?}", op, e),
```

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -561,30 +572,35 @@ impl DiskController {
...
+                        err = Some(Error::vdisk_not_found(op.vdisk_id()))
                     }
-                };
-
-            match vdisk_group.delete(key, meta, StartTimestampConfig::new(false), force_delete).await
-            {
-                Err(e) => Err(self.process_error(e).await),
-                Ok(x) => Ok(x),
+                }
+            }
```
ikopylov, line 593: You can safely add `else` here because `if` statement has `groups.is_empty()` condition, so the loop below does nothing in this case (the collection is empty). If you do this, then the code above can be simplifying by returning errors immediately (no need to store them in `err` variable)

bob-backend/src/pearl/disk_controller.rs
```diff
@@ -561,30 +572,35 @@ impl DiskController {
...
+            for g in groups {
+                match g.delete(key, meta, StartTimestampConfig::new(false), force_delete).await {
+                    Ok(r) => result += r,
+                    Err(e) => err = Some(self.process_error(e).await),
```
ikopylov, line 597: It would be good to write all errors to log on debug level

ikopylov (approved):
LGTM

PR #866, opened 2024-01-31
568 nodes list generation by pattern

### LINKED ISSUE #568: `ccg`: add support for nodes list generation by pattern
With a large number of servers, it becomes difficult to fill even the `nodes` section.
I suggest to add command to ccg, that allows to generate `nodes` section for `cluster.yaml` or add new nodes to existed `cluster.yaml` by patterns. Patten should support range syntax. 

Example: `192.168.0.[10-18]/opt/disk[1-12]`
With this pattern `nodes` section should have 8 nodes (`192.168.0.10`, `192.168.0.11`, ..., `192.168.0.18`) each of which should have 12 disks (`/opt/disk1`, `/opt/disk2`, ..., `/opt/disk12`).

Node name should be generated from node address by replacing non digit or letters symbols to `_`. Example: `192_168_0_10`.
Disk name should be generated with auto increment on every node. Example: `disk1`, `disk2`, ...

Another valid pattern examples: `node-[1-8]/opt/l[1-2]/d[1-4]`, `192.168.[0-4].[10-12]/disk1`

ficusinapot:


bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,267 @@
+use anyhow::{anyhow, Result as AnyResult};
+use bob_common::configs::cluster::Node;
+use bob_common::core_types::{DiskName, DiskPath};
+use itertools::Itertools;
+
+fn parse_address_pattern(pattern: &String) -> AnyResult<((String, u16), String)> {
+    debug!("Pattern to parse: {}", pattern);
+    let re: regex::Regex = regex::Regex::new(r"^([\w.]+):(\d+)(/[\w/]+)$").unwrap();
```
idruzhitskiy: I think it's better to use named groups here, like `addr`, `port`, `path`. Then you can use `captures["addr"]` to get group value, and will not depend on any future modifications to group count or order

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,267 @@
...
+            .parse()
+            .map_err(|_| anyhow!("Failed to parse port {}", port))?;
+        Ok(((ip, port), path))
+    } else {
+        Err(anyhow!("Failed to parse the pattern {}", pattern))
+    }
+}
+
+fn substitute_node(node_pattern: &str, ip: &str, port: u16, id: usize) -> String {
+    node_pattern
```
idruzhitskiy: Please use `Regex::replace_all` to replace multiple matches

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,267 @@
...
+        start..=end
+    });
+
+    re.split(pattern)
+        .zip_longest(ranges)
+        .map(|x| {
+            if let itertools::EitherOrBoth::Both(part, range) = x {
+                range.map(|i| part.to_string() + &i.to_string()).collect()
+            } else {
+                vec![x.left().unwrap().to_string()]
```
idruzhitskiy: IMO it's better to use `expect` with some clarification why we don't expect error here, like `split > ranges`

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,267 @@
...
+) -> AnyResult<Vec<Node>> {
+    let mut err = Ok(());
+    let mut node_counter = old_nodes.len();
+    let new_nodes: Vec<Node> = generate_range_samples(&pattern)
+        .map_while(|key| {
+            let result = parse_address_pattern(&key);
+            match result {
+                Ok(((ip, port), path)) => Some(((ip, port), path)),
+                Err(e) => {
+                    err = Err(e);
```
idruzhitskiy: Using mutable var within functional code seems wrong :) Also later you return error if error occured here, so I suggest to rewrite it as `for` with `?`

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,267 @@
...
+                    None
+                }
+            }
+        })
+        .group_by(|key| key.0.clone())
+        .into_iter()
+        .filter_map(|(ip_port, addresses)| {
```
idruzhitskiy: Better split this into two operations:
1. Create node
2. Merge with old nodes: either add node or add disks to existing one

idruzhitskiy (changes requested):

bob-common/src/configs/cluster.rs
```diff
@@ -97,6 +105,20 @@ impl Node {
         &self.disks
     }
 
+    /// Merges new disk paths into a node's disk list, without duplicates.
+    pub fn merge_disks(&mut self, new_disks_paths: impl Iterator<Item = String>) {
```
idruzhitskiy: The logic with naming disk by counters seems more related to pattern generation, than to actual node. IMO better to move this to `config_cluster_generator...`, simply recreating node with new disks.

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,250 @@
...
+        })
+        .multi_cartesian_product()
+        .map(|x| x.concat())
+        .into_iter()
+}
+
+fn pattern_extend_disks(node: &mut Node, disk_paths: impl Iterator<Item = String>) {
+    let old_disks = node.disks();
+    let new_disks: Vec<DiskPath> = disk_paths
+        .filter(|disk_path| !old_disks.iter().any(|d| d.path() == *disk_path))
```
idruzhitskiy: It's better to create set here (`HashSet`) with old disk paths and check if set does not contain path

bob-apps/bin/config_cluster_generator/pattern.rs
```diff
@@ -0,0 +1,250 @@
...
+                        ip_port.1,
+                        old_nodes.len() + 1,
+                    ),
+                    address,
+                    vec![],
+                ));
+                p_node = old_nodes.last_mut();
+            }
+            let p_node = p_node.expect("is some because it's either the pushed one or found");
+            pattern_extend_disks(p_node, addresses.map(|(_, path)| path.to_owned()));
```
idruzhitskiy: I think this cycle should be split in two: 
1. Create missing nodes
2. Extend disks for all nodes

idruzhitskiy (approved):

PR #867, opened 2024-02-08
730 crossplatform metrics

### LINKED ISSUE #730: Make hardware metrics crossplatform
Right now some metrics depends on `unix`. We can extract platform dependent metrics into separate module and make 2 versions of it: for unix and for windows. If some metrics can't be acquired easily on windows, then we can simply return `unavailable` error and skip them.

idruzhitskiy:
Closes #730


ikopylov:
Here is the issue with incorrect disk space calculation on MacOs: https://github.com/qoollo/bob/issues/825

`frsize` field should be used instead of `bsize` in the calculation logic. You can temporary change the field and verify the calculation on macOs

PR #877, opened 2024-11-27
[876] Reduce startup logs

### LINKED ISSUE #876: Reduce startup logs in normal flow
When starting with sync IO when it's set in config, bob should not print `warn` messages about AIO being disabled

idruzhitskiy:
Closes #876


bob-backend/src/pearl/holder.rs
```diff
@@ -405,8 +405,8 @@ impl Holder {
                 }
             }
         } else {
-            warn!("bob will start with standard sync fs io api");
-            warn!("cause: disabled in config");
+            trace!("bob will start with standard sync fs io api");
+            trace!("cause: disabled in config");
```
ikopylov: Please, combine 2 separate calls into one

